PR #31, opened 2023-06-10
cleaned up to Zepp for demo and upped the max token count in ai.txs

larryatfixie:


larryatfixie:
Cleaned up the zepp demo and upped the maxIndex count in ai.tsx

PR #33, opened 2023-06-12
Use Typescript 5.1 JSX ElementType

petersalas:
This uses the [new Typescript 5.1 feature](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types) to add proper type-checking for AI.JSX elements.

Additionally it fixes a couple places where were (correctly!) identified as incorrectly typed.


petersalas:
@NickHeiner how do you feel about requiring TS 5.1 for our types? On one hand it's a barrier to adoption, OTOH it's one that will go down (presumably quickly) over time. Is the TS community generally very quick to adopt new TS versions?

packages/examples/src/debate.tsx
```diff
@@ -27,32 +27,29 @@ async function ConversationRemapper(
...
-      }
+    if (assistant(node)) {
+      return <AssistantMessage>{node}</AssistantMessage>;
+    }
 
-      return node;
-    })
-    .filter((n) => n !== null);
+    return node;
+  });
```
NickHeiner, line 52: Is this just a formatting change?

NickHeiner:
I was hoping this would also clean up some of the types in NextJS, but I recognize that's a hot mess.

NickHeiner (approved):

packages/examples/src/debate.tsx
petersalas: Not quite, the types now allowed TS-lint to correctly identify that the `filter` was not necessary. (Which apparently triggered prettier to reformat the rest of the code.)

petersalas:
> I was hoping this would also clean up some of the types in NextJS, but I recognize that's a hot mess.

I think we can likely use the same strategy to provide different separate JSX types for the integration.

PR #35, opened 2023-06-12
Add LogImplementation.logException and better error types for OpenAI

petersalas:
This adds:
- A `LogImplementation.logException` method, which by default invokes `LogImplementation.log`. In the default implementation an exception is logged with level `"error"` the first time it is seen, and level `"trace"` subsequently (as it propagates up the element tree).
- More consistent error handling for OpenAI HTTP exceptions. We now parse the error message out of the response body and put it in the `Error` message.


NickHeiner:
We can ignore the Vercel warnings.

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -29,7 +29,7 @@ function AutomaticCompletionModel({ children, ...props }: ModelPropsWithChildren
   }
 
   throw new Error(
-    'No completion model was specified. Specify a CompletionProvider or set the OPENAI_API_KEY environment variable.'
+    'No completion model was specified. Set the OPENAI_API_KEY environment variable to use OpenAI or use an explicit CompletionProvider.'
   );
```
NickHeiner, line 33: It doesn't have to be in this PR, but eventually I'd like all our errors to have an `AI.JSX(1234)` error code, for easy googling / referencing from docs / etc.

packages/ai-jsx/src/core/log.ts
```diff
@@ -1,26 +1,61 @@
...
-  metadataOrMessage: object | string,
-  message?: string
-) => void;
+export abstract class LogImplementation {
+  protected readonly loggedExceptions = new WeakMap<object, boolean>();
+
+  /**
+   * @param level The logging level.
+   * @param element The element from which the log originated.
+   * @param renderId A unique identifier associated with the rendering request.
```
NickHeiner, line 15: This comment should say whether `renderId` is scoped to the entire top-level render or just the individual component render.

packages/ai-jsx/src/core/log.ts
```diff
@@ -1,26 +1,61 @@
...
+   * @param metadataOrMessage An object to be included in the log, or a message to log.
+   * @param message The message to log, if `metadataOrMessage` is an object.
```
NickHeiner, line 17: I think there's a way with TS overloads we can avoid the "object or a string" and "undefined or a string" thing.

packages/ai-jsx/src/core/log.ts
```diff
@@ -1,26 +1,61 @@
...
+    metadataOrMessage: object | string,
+    message?: string
+  ): void;
+
+  /**
+   * Logs exceptions thrown during an element's render. By default invokes `log` with level `"error"`
+   * for the element that threw the exception and level `"trace"` for elements through which the exception
+   * propagated.
+   *
+   * @param element The element from which the exception originated or through which the exception was propagated.
```
NickHeiner, line 32: If an `ErrorBoundary` catches an error, will it appear as an `element` here?

packages/ai-jsx/src/core/log.ts
```diff
@@ -1,26 +1,61 @@
...
+   * @param renderId A unique identifier associated with the rendering request for this element.
+   * @param exception The thrown exception.
+   */
+  logException(element: Element<object>, renderId: string, exception: any) {
```
NickHeiner, line 36:
```suggestion
  logException(element: Element<object>, renderId: string, exception: unknown) {
```

packages/ai-jsx/src/core/log.ts
```diff
@@ -1,26 +1,61 @@
...
+      alreadyLoggedException ? 'trace' : 'error',
+      element,
+      renderId,
+      { exception },
+      `Rendering element ${elementTag} failed with exception: ${exception}`
+    );
+  }
+}
+
+export class EmptyLogImplementation extends LogImplementation {
```
NickHeiner, line 56:
```suggestion
export class SilentLogImplementation extends LogImplementation {
```

```suggestion
export class NoOpLogImplementation extends LogImplementation {
```

minor suggestion / do whatever seems best to you

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -126,6 +128,49 @@ function logitBiasOfTokens(tokens: Record<string, number>) {
...
+      // The response wasn't JSON, ignore it.
+    }
+
+    const parsedMessage = errorResponse?.error?.message;
+    if (typeof parsedMessage === 'string' && parsedMessage.trim().length > 0) {
+      responseSuffix = `: ${parsedMessage}`;
+    }
+
+    super(
+      `OpenAI ${method} request failed with status code ${response.status}${responseSuffix}`,
```
NickHeiner, line 152: Whenever this bubbles up to being read by the human, I'd like to link directly to the OpenAI dev docs.

NickHeiner:
When I run this against `simple-chat` with an example that overflows the token count, I get:

```
OpenAIError: OpenAI createChatCompletion request failed with status code 400: This model's maximum context length is 4097 tokens. However, your messages resulted in 12519 tokens. Please reduce the length of the messages.
    at checkOpenAIResponse (file:///Users/nth/code/ai-jsx/packages/ai-jsx/dist/esm/lib/openai.js:88:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async OpenAIChatModel (file:///Users/nth/code/ai-jsx/packages/ai-jsx/dist/esm/lib/openai.js:163:5) {
  statusCode: 400,
  responseBody: '{\n' +
    '  "error": {\n' +
    `    "message": "This model's maximum context length is 4097 tokens. However, your messages resulted in 12519 tokens. Please reduce the length of the messages.",\n` +
    '    "type": "invalid_request_error",\n' +
    '    "param": "messages",\n' +
    '    "code": "context_length_exceeded"\n' +
    '  }\n' +
    '}\n',
  responseHeaders: {
    date: 'Tue, 13 Jun 2023 02:09:18 GMT',
    'content-type': 'application/json',
    'content-length': '282',
    connection: 'close',
    'access-control-allow-origin': '*',
    'openai-organization': 'fixie-ai',
    'openai-processing-ms': '5198',
    'openai-version': '2020-10-01',
    'strict-transport-security': 'max-age=15724800; includeSubDomains',
    'x-ratelimit-limit-requests': '3500',
    'x-ratelimit-limit-tokens': '90000',
    'x-ratelimit-remaining-requests': '3499',
    'x-ratelimit-remaining-tokens': '85903',
    'x-ratelimit-reset-requests': '17ms',
    'x-ratelimit-reset-tokens': '2.73s',
    'x-request-id': '6bca147a29a141d30e86aa8eec30b143',
    'cf-cache-status': 'DYNAMIC',
    server: 'cloudflare',
    'cf-ray': '7d66ca49faec8f17-IAD',
    'alt-svc': 'h3=":443"; ma=86400'
  },
  errorResponse: {
    error: {
      message: "This model's maximum context length is 4097 tokens. However, your messages resulted in 12519 tokens. Please reduce the length of the messages.",
      type: 'invalid_request_error',
      param: 'messages',
      code: 'context_length_exceeded'
    }
  }
}
```

This is a big improvement over `main`'s current behavior, which is:

```
file:///Users/nth/code/ai-jsx/packages/ai-jsx/dist/esm/lib/openai.js:147
        throw new Error(`OpenAI request failed with status code ${chatResponse.status}`);
              ^

Error: OpenAI request failed with status code 400
    at OpenAIChatModel (file:///Users/nth/code/ai-jsx/packages/ai-jsx/dist/esm/lib/openai.js:147:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v18.16.0
```

That said, I would like to present a more polished dev experience. Something that looks more like:

```
----------------------------------
| AI.JSX(1234): Model Call Error  |
----------------------------------
This model's maximum context length is 4097 tokens. However, your messages resulted in 12519 tokens. Please reduce the length of the messages.

For more detail, see ai-jsx.log.
Need help? https://ai-jsx-docs.vercel.app/
```

In that last part, we have a common suffix.

This can be a separate PR.

packages/ai-jsx/src/core/errors.ts
```diff
@@ -0,0 +1,10 @@
+export class HttpError extends Error {
```
NickHeiner, line 1: To your point about normalization / keeping core thin – what if we didn't have any of these fields, and just included the Axios response object? 

The way I see it, we have two use-cases:
1. Programmatic log observers wants as much data as possible
2. Humans reading logs in the command line or in the logs wants focused results.

This approach hits a middle ground that doesn't fully satisfy either use-case. 

I think we want less normalization within AI.JSX – so just attach the Axios object. But at the very edge, where we format for the dev, we have an opinionated formatter. (Maybe there's an `OpenAIError` class that contains that opinionation, via a `toString` or `formatForLogs` method or something.)

For the log file, maybe what we want is to log twice:

```
log.trace(error, 'error details');
log.error(summarize(error), 'error summary');
```

I don't love that solution tbh. Another solution (which presumes pino) is to include a `pino-pretty` command that filters which log fields to show, but I don't love that either.

~~~

Having said all that, what we have here is basically fine, and it's probably the case that in the majority of the time, all you need are the fields you've picked out here, and it is genuinely useful to have them normalized across whatever HTTP libraries the various model providers use.

NickHeiner (approved):
This isn't the be-all-end-all of error logging but it's an improvement, so 🚢 🇮🇹 . 

I'd say spend at most an hour addressing / considering my comments, but no need to spiral out on them.

packages/ai-jsx/src/core/log.ts
petersalas: I think there is... but as I started to do this I realized that the user audience for this interface is actually on the implementor side, in which case I think it's easier to present the one canonical interface to implement, rather than requiring the dev to synthesize the overloads into a single implementation.

packages/ai-jsx/src/lib/openai.tsx
petersalas: Good idea!

packages/ai-jsx/src/core/errors.ts
petersalas: TBH because the Axios object is such a dumpster fire when logged I went out of my way to _avoid_ including it in the (logged) exception.

I agree that I'm sort of guessing at a middle ground which is a dicey proposition. I think if I were to make a change here I'd just merge `HttpError` and `OpenAIError`, but I'm going to roll with this for the time being while internally considering `HttpError` a tenuous abstraction that should be revisited if/when we add more service-backed components :)

packages/ai-jsx/src/core/errors.ts
```diff
@@ -0,0 +1,10 @@
...
+  constructor(
+    message: string,
+    readonly statusCode: number,
+    readonly responseBody?: string,
+    readonly responseHeaders?: Record<string, string>
```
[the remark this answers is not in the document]
petersalas: Yeah I was originally going to include them but the footgun I was trying to avoid was unconditionally logging api keys at the error level.

I think the escape hatch is that components can (do, in the case of OpenAI components) log these details at more verbose levels.

PR #36, opened 2023-06-13
Set up a create-react-app demo

NickHeiner:


packages/ai-jsx/src/core/log.ts
```diff
@@ -38,6 +38,7 @@ const defaultPinoLogger = _.once(() =>
  * @returns A `LogImplementation`.
  */
 export function pinoLogger(pinoLogger?: pino.Logger): LogImplementation {
+  debugger;
```
petersalas, line 41: Leftover from debugging?

packages/create-react-app-demo/.gitignore
```diff
@@ -0,0 +1,23 @@
...
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
petersalas, line 23: Should we add ai-jsx.log since we log to that path by default?

packages/create-react-app-demo/src/ai.tsx
```diff
@@ -0,0 +1,175 @@
...
+} from './recipe/page.tsx';
+
+export function useAI(children: LLMx.Node, dependencies: unknown[], when: boolean = true) {
+  const isInProgressRef = useRef(false);
+  const mostRecentlyRenderedChildren = useRef(children);
+  // If `children` changes, but a previous call is still in progress, will we properly start a new one?
+  const [result, setResult] = React.useState('');
+  const [isDone, setIsDone] = React.useState(false);
+
+  useEffect(() => {
```
petersalas, line 23: IIUC this is client-side -- because of the API key problem I'm not sure fully client-side will ever be a first class scenario, right?

packages/create-react-app-demo/src/ai.tsx
```diff
@@ -0,0 +1,175 @@
...
+    if (isInProgressRef.current || !when) {
+      return;
+    }
+    setResult('');
+    setIsDone(false);
+    LLMx.createRenderContext({
+      logger: console.log,
```
petersalas, line 30: Might want `pinoLogger(pino())` instead here

PR #38, opened 2023-06-13
Prompts and constrained output (JSON/YAML)

farzadab:
This PR adds some prompts as well as JSON/YAML validated ChatCompletion components.

To make sure the PR merges in time, schema validation functionality is left for follow up PRs.


NickHeiner:
For next time, it would be better to split the prompt templates and the constrained output into two separate PRs.

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,141 @@
+/**
+ * This module defines affordances for constraining the output of the model
+ * into specific formats, such as JSON, YAML, or Markdown.
+ */
+import * as LLMx from '../index.js';
+import { ChatCompletion, SystemMessage, AssistantMessage, UserMessage } from './completion';
+import { memo } from './memoize';
+import yaml from 'js-yaml';
+
+function isJsonString(str: string): boolean | string {
```
NickHeiner, line 10: This is a bit dangerous because someone could see it and write:

```
if (isJsonString(myStr)) { 
```

But this method will return a truthy value in both cases. 

Instead, I'd make the return type something like `{ isJsonString: true; error: null } | { isJsonString: false; error: string}`.

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,141 @@
...
+ * @param validator function that returns true if the output is valid, or a string with an error message if not
+ *     // TODO: what should the signature of this function be?
+ * @param retries if output fails validation, how many times should we retry?
+ * @param verbose returns intermediate output if true
+ * @returns a string that validates as the given type (TODO: should I return the parsed object instead?)
+ */
+async function* ObjectFormatChatCompletion(
+  {
+    retries = 2,
+    verbose = false,
```
NickHeiner, line 43: Don't take a verbosity param. Just use the log levels. `logger.trace`, `debug`, `warn`, etc.

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,141 @@
...
+  if (verbose) yield childrenWithCompletion;
+
+  let output = await render(childrenWithCompletion);
+  let valiationResults = validator(output);
+  if (valiationResults === true) {
+    return output;
+  } else {
+    if (verbose) yield `Intermediate output did not validate: '''${output}'''`;
+
+    for (let i = 0; i < retries; i++) {
```
NickHeiner, line 75:
```suggestion
    for (let retryIndex = 0; retryIndex < retries; retryIndex++) {
```

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,141 @@
...
+      output = await render(completionRetry);
+      valiationResults = validator(output);
+      if (valiationResults === true) {
+        return output;
+      }
+
+      if (verbose) yield `Intermediate output did not validate: '''${output}'''`;
+    }
+  }
+  // TODO: throw error or fail silently with a message?
```
NickHeiner, line 104: Throw. That will generate a message.

packages/ai-jsx/src/core/prompts.tsx
```diff
@@ -0,0 +1,93 @@
+/**
+ * Basic prompts for LLMs.
+ * This file attemps to have suggestions to guide the thinking of inexperienced users
+ * to use good practices when using LLMs.
+ *
+ * Prompting is still a developing field. There are many different ways to prompt an LLM.
+ * This file attempts to provide some good defaults as a starting point.
+ */
```
NickHeiner, line 8: This is good context, but you should generally assume people won't see it. The parts people will see are the intellisense comments on the exported API members. So if there's critical context, it should live there.

packages/ai-jsx/src/core/prompts.tsx
```diff
@@ -0,0 +1,93 @@
...
+import * as LLMx from '../index.js';
+import { SystemMessage } from './completion';
+
+/**
+ * A (system) prompt collection with some useful defaults to get you started.
+ * By default all prompts are disabled to make it easier for combining them.
```
NickHeiner, line 14:
```suggestion
 * By default all prompts are disabled to make it easier to combine them.
```

packages/ai-jsx/src/core/prompts.tsx
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * We intend to improve this collection over time and welcome contributions.
+ * Potentially, the prompts can be adjusted based on the model being used.
```
NickHeiner, line 17:
```suggestion
```

I tend not to want to write things like this in API docs. I think devs just want to know what the current state is.

packages/ai-jsx/src/core/prompts.tsx
```diff
@@ -0,0 +1,93 @@
...
+ *    the model to output better responses.
+ *
+ *    Example values:
+ *      * an expert in astrophysics
+ *      * Albert Einstein
+ *      * a 5 year old
+ *      * a tactful politician
+ *
+ *    @see https://arxiv.org/pdf/2305.14930.pdf
+ *    @see https://arxiv.org/pdf/2111.02080.pdf
```
NickHeiner, line 35: great!! I think we should vigorously link out to things like this.

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </SystemMessage>
+    </ChatCompletion>
+  );
+
+  if (verbose) yield childrenWithCompletion;
+
+  let output = await render(childrenWithCompletion);
+  let valiationResults = validator(output);
+  if (valiationResults === true) {
+    return output;
```
NickHeiner, line 71: @petersalas rather than returning a string, what would you think about returning the parsed value?

If one wrote:

```tsx
const character = await render(<JsonOutput><GenerateFantasyCharacter /><JsonOutput>)
```

It seems nice to have `character` be parsed for me already, and not need to wrap it in `JSON.parse`.

Will it mess up the render machinery to be passing arbitrary values through, rather than assuming the output of every render is a string or another well-known value?

packages/ai-jsx/src/core/constrained-output.tsx
farzadab: Oh I was hoping there would be something like this.

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,141 @@
...
+    return e.message;
+  }
+  return true;
+}
+
+/**
+ * ChatCompletion that constrains the output to be a valid object format (e.g. JSON/YAML).
+ *
+ *
+ * @param type_name name of the type, to be used in the prompt
```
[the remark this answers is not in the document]
farzadab: Good to know. I'm still a bit sad that we cannot export these to docs.

packages/ai-jsx/src/core/constrained-output.tsx
farzadab: Actually, I agree for some lines, but for some, the idea was that if I `render` something internally, you can't see it anywhere even with `showInspector`.
This might be solved with better observability, so maybe I shouldn't worry about it.

packages/ai-jsx/src/core/constrained-output.tsx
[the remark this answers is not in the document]
NickHeiner: I think we can! We just need to find a tool to do so.

packages/ai-jsx/src/core/constrained-output.tsx
petersalas: Currently everything (eventually) renders to strings, but partial rendering is the (not super ergonomic) exception.

`JsonOutput` could render to say `<Json value={value} />` that internally does a `JSON.stringify` if forced to render.

And then the caller could do `await render(<JsonOutput />, { stop: (e) => e.tag === Json })` which will return an array containing strings and/or `Json` elements (from which the result could be plucked from props).

At the moment I think `JSON.parse` is nicer than ☝️, but this is an absolutely a scenario on my mind. Right now the bias is towards strings so that everything can be trivially passed back _into_ the LLM but I can imagine us supporting other atomic types that can be coerced into strings (or not if the model supports e.g. images!).

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,141 @@
...
+  }: {
+    validator: (output: string) => boolean | string;
+    type_name: string;
+    retries?: number;
+    verbose?: boolean;
+    children: LLMx.Node[];
+  },
+  { render }: LLMx.RenderContext
+) {
+  const childrenWithCompletion = memo(
```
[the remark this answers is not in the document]
farzadab: I did that and I assumed it would work, but I noticed some weirdness at some points (I can't be sure, but I feel like the output would sometimes change; this happened very fast so I couldn't be sure). I didn't understand it, but I thought maybe that's why we have `memo` in the first place?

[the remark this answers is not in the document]
petersalas: Was it possible you had this instead?

```jsx
const children = <ChatCompletion>...</ChatCompletion>;
if (verbose) yield children
let output = await render(children)
```

That would cause independent renders because the thing being yielded is not via the render result but the tree itself.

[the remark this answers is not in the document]
farzadab: I feel like I was using the rendered version only, but I'm most likely wrong. It's good to know that I don't need `memo`. Removing it.

packages/ai-jsx/src/core/constrained-output.tsx
NickHeiner: Yeah, `showInspector` has a gap right now where it doesn't get those internal `render`s, as you point out. We just need to find a better way to do that.

farzadab:
Just to update: I'm kinda blocked on getting schema working. The issue is that for a schema to work, the model should also be aware of it. `zod`'s schema however, isn't human readable so not good for describing it to LLM (even if Perplexity doesn't agree). I also don't see a way of "creating" a `zod` schema from a human-readable string.

I have two paths:
- let user handle describing the schema themselves
- use some other library like [`ajv`](https://ajv.js.org/)

The only reason for my insistence on `zod` until now is because Nick told me this is what I should use.

image
image

farzadab:
PTAL
I decided not to go with schema for this PR.
I added some docs and examples too.

packages/ai-jsx/src/core/constrained-output.tsx
```diff
@@ -0,0 +1,131 @@
...
+interface ValidationResult {
+  success: boolean;
+  error: string;
+}
+
+// TODO: schema
+/**
+ * A ChatCompletion component that constrains the output to be a valid JSON string.
+ *
+ * @returns a string that validates as JSON or throws an error after `retries` attempts
```
NickHeiner, line 18:
```suggestion
 * @returns a string that is valid JSON or throws an error after `retries` attempts
```

packages/docs/docs/guides/prompting.md
```diff
@@ -41,3 +41,50 @@ Output:
...
+
+We have included a small set of prompts that we found useful in (`packages/ai-jsx/src/core/prompts.tsx`).
+You can use them either as shortcuts, or as a starting point if you are new to prompting:
+
+```tsx
+function App() {
+  // Asking the model to respond as if they were a React developer
+  return (
+    <ChatCompletion>
+      <Prompt persona="a React developer" />
```
NickHeiner, line 55: If the idea is that the `Prompt` is a drop-in `SystemMessage`, then I'd name it something different – maybe `PersonaSystemMessage`?

When I first had the idea of naming something `Prompt`, I thought it would look more like:

```tsx
<SystemMessage>
  <Prompt hhh />
  Answer using this context: <DocsQA />
</SystemMessage>
```

packages/docs/docs/guides/prompting.md
```diff
@@ -41,3 +41,50 @@ Output:
...
+      <UserMessage>What is JSX?</UserMessage>
+    </ChatCompletion>
+  );
+}
+```
+
+## Getting the AI to say the right thing: constrained output
+
+Sometimes you just want the model to respond in a certain format (e.g. JSON or YAML), but doing so reliably can be hard.
```
NickHeiner, line 64:
```suggestion
Sometimes you want the model to respond in a certain format (e.g. JSON or YAML), but doing so reliably can be hard.
```

packages/docs/docs/guides/prompting.md
```diff
@@ -41,3 +41,50 @@ Output:
...
+          Create a random object describing an imaginary person that has a "name", "gender", and "age".
+        </UserMessage>
+      </YamlChatCompletion>
+    </>
+  );
+}
+```
+
+Under the hood, this model will use a combination of prompting, validating the output, and asking them the model to retry
+if the validation fails (refer to `packages/ai-jsx/src/core/constrained-output.tsx`).
```
NickHeiner, line 90: I know we have a bunch of links like this, but going forward we should actually like to the file directly (e.g. https://github.com/fixie-ai/ai-jsx/blob/main/packages/ai-jsx/src/index.ts)

NickHeiner (approved):

packages/docs/docs/guides/prompting.md
farzadab: You can still do the above too, right?
From what I understood, everything except for the parent Message type gets resolved to the children, so you can just as easily write:
```tsx
<UserMessage>
  <Prompt hhh />
</UserMessage>
```
And it would become a `UserMessage`. It feels to me like I'm just creating a default: if they don't specify, it's going to be a `SystemMessage`.

packages/docs/docs/guides/prompting.md
farzadab: Oh okay, will do. I was just following what was there.

packages/docs/docs/guides/prompting.md
NickHeiner: Yep, I agree that would work – it just feels weird to see it `Prompt` written as a top-level child of `ChatCompletion`. But it's not that big a deal. No need to change it until we see if we get more feedback to this extent.

PR #39, opened 2023-06-13
Add Pendo tracking to AI.JSX docs.

mdwelsh:


packages/docs/static/js/pendo.js
```diff
@@ -0,0 +1,52 @@
...
+      // is_paying:    // Recommended if using Pendo Feedback
+      // monthly_value:// Recommended if using Pendo Feedback
+      // planLevel:    // Optional
+      // planPrice:    // Optional
+      // creationDate: // Optional
+
+      // You can add any additional account level key-values here,
+      // as long as it's not one of the above reserved names.
+    },
+  });
```
NickHeiner, line 51: I think we can just remove all this, since we don't have any means of authenticating users on the docs site.

PR #40, opened 2023-06-13
Hello World CodeSandbox

mdwelsh:
This is a simple AI.JSX app as a CodeSandbox template. I'll create some more examples but thought having a really bare-bones one as a starting point would make sense.


NickHeiner:
Oh damn, are we able to deploy to CodeSandbox directly from the monorepo?

packages/sandboxes/hello-world/.codesandbox/tasks.json
NickHeiner, line 1: You could try naming this file `.jsonc` to fix the syntax highlighting.

packages/sandboxes/hello-world/.codesandbox/template.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "AI.JSX - Hello World",
+  "description": "A simple hello world app for AI.JSX",
+  "iconUrl": "https://docs.ai-jsx.com/img/ai-jsx.svg",
```
NickHeiner, line 4: Woooo get that branding y'all

packages/sandboxes/hello-world/README.md
```diff
@@ -0,0 +1,27 @@
+# AI.JSX Hello World
+
+This is a simple [AI.JSX](https://ai-jsx.com) project that demonstrates how to use the `ai-jsx`
+package to create a simple AI-powered application.
```
NickHeiner, line 4:
```suggestion
package to create an AI-powered application.
```

We probably don't need to say "simple" twice in the same sentence.

packages/sandboxes/hello-world/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+This is meant to be as simple as possible, and as such the app here does not have any actual
+UI -- it just writes its output (in the form of verbose logs) to the console.
+
+## Building and running the app
+
+The easiest way to get started is to create a sandbox on CodeSandbox using
+[this template](https://codesandbox.io/p/sandbox/ai-jsx-hello-world-8683jx).
+
+To run the app, you will need to configure your `OPENAI_API_KEY` environment variable in the
```
NickHeiner, line 14: Ordered instructions like this are much more legible when structured as an ordered list:

```
To run the app, configure your  `OPENAI_API_KEY` environment variable in the CodeSandbox UI:
1. To do this, navigate to the menu icon in the top left, choose **Project Settings**,
then **Env Variables**.
1. Add a new variable with the name `OPENAI_API_KEY` and the value of your
OpenAI API key, which you can obtain from the
[OpenAI API dashboard](https://platform.openai.com/account/api-keys).
1. Save the variable and then click the prompt to restart the Sandbox VM instance, which allows
the new environment variable to take effect.
```

packages/sandboxes/hello-world/README.md
```diff
@@ -0,0 +1,27 @@
...
+CodeSandbox UI. To do this, navigate to the menu icon in the top left, choose **Project Settings**,
+then **Env Variables**. Add a new variable with the name `OPENAI_API_KEY` and the value of your
+OpenAI API key, which you can obtain from the
+[OpenAI API dashboard](https://platform.openai.com/account/api-keys).
+Save the variable and then click the prompt to restart the Sandbox VM instance, which allows
+the new environment variable to take effect.
+
+CodeSandbox should build the app and run it, and in the `start:logs` tab you should
```
NickHeiner, line 22:
```suggestion
CodeSandbox will build the app and run it, and in the `start:logs` tab you should
```

Let's be confident. :smile:

packages/sandboxes/hello-world/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <SystemMessage>
+        You are an agent that only asks rhetorical questions.
+      </SystemMessage>
+      <UserMessage>How can I learn about Ancient Egypt?</UserMessage>
+    </ChatCompletion>
+  );
+}
+const pinoStdoutLogger = pino({
+  name: "ai-jsx",
+  level: process.env.loglevel ?? "trace",
```
NickHeiner, line 22:
```suggestion
  level: process.env.loglevel ?? "debug",
```

`trace` is pretty noisy – we might want to stick to `debug`. But I don't feel strongly.

packages/sandboxes/hello-world/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "ai-jsx-hello-world",
```
NickHeiner, line 2: Set `"private": true` if you want a safeguard against publishing this to npm.

packages/sandboxes/hello-world/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "@tsconfig/node18/tsconfig.json",
```
NickHeiner, line 2: @petersalas I think we can include a `tsconfig` in our `ai-jsx` package and then people can do `"extends": "ai-jsx/tsconfig"` and get the requisite JSX settings automatically. Maybe we can do that once the dust has settled a bit more.

packages/sandboxes/hello-world/.codesandbox/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM node:18-bullseye
```
NickHeiner, line 1: Our root `package.json` defines all monorepo packages as matching this glob:

```json
  "workspaces": [
    "packages/*"
  ]
```

The files you've added here won't match that, because they're two levels deep.

To resolve this, I think you can do `"packages/**"`, or `"packages/*", "packages/sandboxes/*"`.

NickHeiner: Then you'll want to apply this diff to make `yarn test` work:

```diff
diff --git a/packages/sandboxes/hello-world/package.json b/packages/sandboxes/hello-world/package.json
index c1759b0..0da9253 100644
--- a/packages/sandboxes/hello-world/package.json
+++ b/packages/sandboxes/hello-world/package.json
@@ -5,7 +5,8 @@
   "license": "MIT",
   "scripts": {
     "typecheck": "tsc",
-    "test": "npm run typecheck && tsx index.tsx"
+    "test": "npm run typecheck",
+    "start": "npm run test && tsx index.tsx"
   },
   "dependencies": {
     "ai-jsx": "^0.2.0-1",
@@ -18,4 +19,3 @@
     "typescript": "^5.1.3"
   }
 }
-
diff --git a/packages/sandboxes/hello-world/tsconfig.json b/packages/sandboxes/hello-world/tsconfig.json
index 4d02c7c..456f427 100644
--- a/packages/sandboxes/hello-world/tsconfig.json
+++ b/packages/sandboxes/hello-world/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "extends": "@tsconfig/node18/tsconfig.json",
-  "include": ["src/*.tsx"],
+  "include": ["index.tsx"],
   "compilerOptions": {
     "jsx": "react",
     "jsxFactory": "LLMx.createElement",
@@ -11,4 +11,3 @@
     "noEmit": true
   }
 }
-
```

NickHeiner:
Let's update the docs site to include this. Docusaurus should support us embedding it directly in the page.

NickHeiner (approved):

PR #43, opened 2023-06-14
Adding Dalle image generation

farzadab:
Created an `ImageGen` component which defaults to `DalleImageGen`. Currently there is no other provider to choose from.


packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -0,0 +1,66 @@
...
+  { children, ...props }: ImageGenPropsWithChildren & Record<string, unknown>,
+  { getContext }: RenderContext
+) {
+  const [ImageGenComponent, defaultProps] = getContext(imageGenContext);
+  return (
+    <ImageGenComponent {...defaultProps} {...props}>
+      {children}
+    </ImageGenComponent>
+  );
+}
```
farzadab, line 66: For this file I pretty much followed what was in `completion.js`, so I pretty much don't know what is going on :))

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -291,3 +295,51 @@ export async function* OpenAIChatModel(
...
+
+  logger.debug({ imageRequest }, 'Calling createImage');
+
+  const response = await openai.createImage(imageRequest);
+
+  if (response.status < 200 || response.status >= 300) {
+    throw new Error(`createImage failed with status ${response.status}`);
+  } else {
+    logger.debug({ statusCode: response.status, headers: response.headers }, 'createImage succeeded');
+  }
```
farzadab, line 341: I originally wanted to use `checkOpenAIResponse`, but since this is not streaming, that might not be a good idea. I was hesitant about adding "createImage" to `OpenAIMethod`, but if I do I can throw an `OpenAIError` instead.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -291,3 +295,51 @@ export async function* OpenAIChatModel(
...
   return currentMessage.content;
 }
+
+/**
+ * Generates an image from a prompt using the DALL-E model.
+ *
+ * @returns the URL of the generated image.
+ *          If num_samples is greater than 1, URLs are separated by newlines.
+ */
+export async function DalleImageGen(
```
farzadab, line 305: I wanted to create a new file, but I decided against it for now. I'll let you guys decide on how the code should be organized.

mdwelsh (approved):
This is great, will let Peter/Nick decide on whether it's the right thing or not ;-)

Can I suggest that we start writing detailed docstrings in for each function / component that we're developing so future users of this library can understand how to use it?

packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -0,0 +1,66 @@
...
+        { ...previousDefaults, ...newDefaults },
+      ]}
+    >
+      {children}
+    </imageGenContext.Provider>
+  );
+}
+
+/**
+ * Perform a model call to do a [image generation](https://platform.openai.com/docs/guides/images/introduction).
```
NickHeiner, line 54: It would be helpful for this comment to specify the return type. Byte buffer? Byte stream? URL to the image? etc.

packages/ai-jsx/src/lib/openai.tsx
NickHeiner: I'm fine to default to staying in larger files until a good organizational structure presents itself.

packages/ai-jsx/src/lib/openai.tsx
NickHeiner: I think `OpenAIError` is what we want here.

Ultimately, we want to provide a layer on top of OpenAI to make the errors nicer – for instance, explaining that 429 is a rate limit, etc. If the OpenAI APIs are consistent across DallE / GPT, etc, then it would be nice for our checking to be consistent too.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -291,3 +295,52 @@ export async function* OpenAIChatModel(
...
+  const response = await openai.createImage(imageRequest);
+
+  if (response.status < 200 || response.status >= 300) {
+    throw new Error(`createImage failed with status ${response.status}`);
+  } else {
+    logger.debug({ statusCode: response.status, headers: response.headers }, 'createImage succeeded');
+  }
+
+  // return all image URLs as a newline-separated string
+  return _.map(response.data.data, (x) => x.url).join('\n');
```
NickHeiner, line 345:
```suggestion
  return _.pluck(response.data.data, 'url').join('\n');
```

packages/ai-jsx/src/lib/openai.tsx
NickHeiner: I think `OpenAIError` is what we want here.

Ultimately, we want to provide a layer on top of OpenAI to make the errors nicer – for instance, explaining that 429 is a rate limit, etc. If the OpenAI APIs are consistent across DallE / GPT, etc, then it would be nice for our checking to be consistent too.

packages/ai-jsx/package.json
```diff
@@ -29,6 +29,10 @@
       "import": "./dist/esm/core/completion.js",
       "require": "./dist/cjs/core/completion.js"
     },
+    "./core/image-gen": {
+      "import": "./dist/esm/core/image-gen.js",
+      "require": "./dist/cjs/core/image-gen.js"
+    },
```
NickHeiner, line 35: I think it would be good to add an example usage of this in the `examples` directory. And can we add an entry to the docs about this? Maybe on https://docs.ai-jsx.com/guides/prompting? Or https://docs.ai-jsx.com/getting-started?

NickHeiner (approved):
Looks great! I'm excited to try this out in the UI later.

packages/examples/package.json
```diff
@@ -29,15 +29,16 @@
     "demo:inline-completion": "yarn build && node dist/inline-completion.js",
     "demo:simple-chat": "yarn build && node dist/simple-chat.js",
     "demo:simple-completion": "yarn build && node dist/simple-completion.js",
-    "demo:getting-started": "yarn build && node dist/getting-started/index.js",
+    "demo:getting-started": "yarn build && cp src/getting-started/data.json dist/getting-started/ && node dist/getting-started/index.js",
```
farzadab, line 32: @NickHeiner I get an error when running `getting-started` since `data.json` is not present. This is a hacky solution. Let me know if there's a better way to fix it.

farzadab:
I'm merging this. Let me know if there are issues and I'll fix them in a follow-up.

packages/examples/package.json
NickHeiner: I think this is the best way to do it. Thanks for catching this.

packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Node, Component, RenderContext } from '../index.js';
+import { DalleImageGen } from '../lib/openai.js';
+
+export interface ImageGenProps {
+  /** defines how many image samples should be produced */
+  numSamples?: number;
+  /** defines the image resolution */
+  size?: '256x256' | '512x512' | '1024x1024';
+  /** if true, will clip prompts that are too long */
+  clipLongPrompt?: boolean;
```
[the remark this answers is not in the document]
NickHeiner: Hmm, we don't do this anywhere else, and I'm concerned this is a silent failure. 

I'm more ok with this if the default is `false`. :smile:

packages/docs/docs/getting-started.md
```diff
@@ -183,3 +183,34 @@ $ grep -i 'starting modelcall' packages/ai-jsx/llmx.log | yarn workspace ai-jsx
...
+## Beyond Text: Image Generation
+
+You are not restricted by text models. Let's create an image for the story as well:
+
+```tsx title="index.tsx"
+function StoryWithImage() {
+  const story = memo(<WriteStory />);
+  return (
+    <>
+      Banner URL: <ImageGen clipLongPrompt>Generate an image for this story: {story}</ImageGen>
```
NickHeiner, line 196: Ah, I see why you wanted `clipLongPrompt` – so we could pass a `story` of arbitrary length and not have to worry about overflowing the context window.

I think putting a prop on `ImageGen` in particular is not the right approach. I think what would be better is to have a `<Summarize>{story}</Summarize>` component (perhaps similar to the demo @juberti added). We could have various summarizers with different strategies; in this case, we'd have a `TruncateSummarize` that simply truncates.

That way, we get this important use case (summarization / prompt length management) in a modular way that can be applied anywhere, not just on particular components.

NickHeiner: And in the meantime, if we just want to naively truncate the string (which I think is fine for now), but we also don't want to make a whole `Summarize` component with docs etc, I'd just write `{story.slice(0, maxLength)}`.

NickHeiner, line 196: I would use the `// highlight-next-line` comment here to emphasize that this is the most important line in the code snippet.

packages/examples/src/image-generation.tsx
```diff
@@ -0,0 +1,23 @@
+import * as LLMx from 'ai-jsx';
+import { ChatCompletion, UserMessage } from 'ai-jsx/core/completion';
+import { Prompt } from 'ai-jsx/batteries/prompts';
+import { ImageGen } from 'ai-jsx/core/image-gen';
+import { memo } from 'ai-jsx/core/memoize';
+
+function RecipeWithImage() {
+  const recipeTitle = memo(
+    <ChatCompletion temperature={1}>
+      <Prompt persona="a Michelin Star Head Chef" />
```
NickHeiner, line 10: Nice use of the persona!

packages/docs/docs/getting-started.md
juberti: agreed, clipping may or may not give you the desired result, better to make this explicit

farzadab: Yeah okay I'll remove it. The issue is that a separate summarizer would not know about the context length of the image model.

farzadab: > `{story.slice(0, maxLength)}`
Makes sense, but how do you know `maxLength`?

farzadab: Good to know, thanks.

farzadab: Okay, what if we give the summarizer to the component itself?
```tsx
<ImageGen summarizer={<TruncateSummarizer />}>....<ImageGen>
```

This way the model can give summarizer any info that it needs. Also, it can decide not to summarize short responses.

farzadab: Where is Justin's example btw?

NickHeiner: I would say all this falls under the category of "new features", so let's defer it until after launch. For now, I'd say we should just remove this the `clipLongPrompt` prop and do a `slice` with `maxLength` set to something shorter than every context window, like `1000`. It'll be fine for the purposes of this demo.

I'd rather not pass `summarizer` as a prop to `ImageGen` for the same reasons stated above – structuring the prompt correctly is orthogonal to the `ImageGen` component. Other places will want this same functionality.

> The issue is that a separate summarizer would not know about the context length of the image model.

This is a good question. The summarizer would get it off `context`. We can talk more about how to do that after launch. :smile:

packages/docs/docs/getting-started.md
```diff
@@ -183,3 +183,34 @@ $ grep -i 'starting modelcall' packages/ai-jsx/llmx.log | yarn workspace ai-jsx
   }
 ],
 ```
+
+## Beyond Text: Image Generation
+
+You are not restricted by text models. Let's create an image for the story as well:
```
NickHeiner, line 189: This would read better as:

> You are not restricted to text models.

farzadab: Applied to new PR.

PR #44, opened 2023-06-14
Modify corpus interfaces to make it easier to replace them with remote implementations.

mdepinet:
In particular, this loosens our ties to langchain and makes the path to swapping out "LocalCorpus" with a "FixieCorpus" (while allowing complete customization) relatively straightforward.


NickHeiner:
To fix the CI error, run `yarn` locally and check in the changes to `yarn.lock` that result.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,498 @@
...
+  pino(
+    { name: 'ai-jsx-docs', level: 'trace' },
+    // N.B. pino.destination is not available in the browser
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    pino.destination?.({
+      dest: './ai-jsx-docs.log',
+      sync: true, // Synchronous logging
+    })
+  )
+)();
```
NickHeiner, line 24: @petersalas, I have [these docs](https://ai-jsx-docs-m4s36woh3-fixie-ai.vercel.app/guides/observability#logging-from-within-components) for how to log within a component, but what are we thinking for this context, where we're not immediately within a component? Maybe this is where we test our "can everything be a component" question? It seems a little clumsy for components to have to pass the logger into every helper they call.

NickHeiner, line 24: @petersalas, I have [these docs](https://ai-jsx-docs-m4s36woh3-fixie-ai.vercel.app/guides/observability#logging-from-within-components) for how to log within a component, but what are we thinking for this context, where we're not immediately within a component? Maybe this is where we test our "can everything be a component" question? It seems a little clumsy for components to have to pass the logger into every helper they call.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,498 @@
...
+    readonly encoding: string = 'utf-8',
+    /** The content type for this document. Defaults to plain text. */
+    readonly mimeType: string = 'text/plain'
+  ) {
+    this.pageContent = pageContent;
+    this.name = name;
+    this.encoding = encoding;
+    this.mimeType = mimeType;
+  }
+}
```
NickHeiner, line 48: I think this should be an interface, since it doesn't have any methods/instance state other than the set of fields itself.

https://www.typescriptlang.org/play?#code/C4TwDgpgBAIlC8UBKECGATA9gOwDYgB4BvAKCigDMAuKAZ2ACcBLbAcwG4SBfAPhJIDGOelHQ04iUuWpQA5KlroKs7vwD0aqABUAFk1p0dmAK650UCAwaYGUAEYQBqY7WhNgsgwzRY8IEugAdBRQANSI8orK7EA

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner, line 1: Can you link to https://docs.ai-jsx.com/guides/brand-new#accessing-knowledge-docs-qa from this file?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,498 @@
...
+ *      Responses may include new partitions to be loaded. Partitions are
+ *          non-overlapping subsets of a corpus which may be loaded in parallel.
+ *          A response's new partitions will be ignored if previously included
+ *          in another response.
+ *      When a response includes a page of documents, that page may indicate
+ *          that another page is available in the same partition. Pages are
+ *          always loaded serially in order. The partition is completed when
+ *          a response has a page with no next page token.
+ *
+ *  Loading will always begin with a  request with the default (unnamed)
```
NickHeiner, line 109:
```suggestion
 *  Loading will always begin with a request with the default (unnamed)
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,498 @@
...
+ *          no more children, at which point the response has no
+ *          next page token and the partition is complete.
+ *
+ *          Note: Including multiple parent_ids in each partition would also
+ *              work and would be an effective way to limit parallelism if
+ *              desired.
  */
-// TODO: maybe rename to DocumentProvider
-export type Loader = (...args: any[]) => Promise<Document[]>;
+export namespace CorpusLoading {
```
NickHeiner, line 153: Hmm, why do we need to use a namespace?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,498 @@
...
+       * be empty for the initial request, indicating the default partition.
+       * For subsequent requests, it will typically be the name of a partition
+       * returned by a previous request, though it could be empty if the
+       * default partition contains multiple pages.
+       */
+      this.partition = partition;
+      /**
+       * A token for paginating results within a corpus partition.
+       * If present, this will be echoed from a previous response.
+       */
```
NickHeiner, line 167: I'm loving all these comments!

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,498 @@
...
+    loadingState: LoadingState;
+    completedPartitions: number;
+    activePartitions: number;
+    numDocuments: number;
+    numChunks: number;
+  }
+}
+
+/* Resolving disagreement between formatter and linter. */
+/* eslint-disable brace-style */
```
NickHeiner, line 374: I'm surprised this is happening. If it is, we just need to go to `.eslintrc.cjs` and disable the `brace-style` rule globally.

NickHeiner:
I'll finish reviewing this later but I like the direction and idea. I was thinking our docs stuff was a little under-powered so I'm glad you took this on. Thanks!

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: :+1:

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: FWIW, I don't intend to merge the logging into main, at least as currently written. I just haven't removed it yet because it's useful for testing and I expect there to be other comments. :)

mdepinet: (resolving dupe)

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: Done.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: I thought it was a useful way to group related types, but this is something I was hoping to get feedback on.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: That's an option, but I think the brace-style rule is good in general. There's just disagreement with prettier for long lines with generics. I'm happy with whatever approach you prefer!

mdepinet:
> To fix the CI error, run `yarn` locally and check in the changes to `yarn.lock` that result.

Hmm... Running yarn locally didn't result in any changes. Could I have something misconfigured in my environment?

packages/ai-jsx/src/batteries/docs.tsx
petersalas: Certainly crawling doesn't make sense to be wrapped as component, in that the machinery that drives crawling is ~orthogonal to rendering/query-time machinery -- so I don't even think there's a sensible `Logger` to pass. (Though in general if we want logging to be associated with a render, I think passing the `Logger` explicitly is sensible?)

In general it's not obvious to me how exactly crawling code can/should fit into AI.JSX (vs. say a Fixie SDK).

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
```diff
@@ -1,47 +1,54 @@
...
-import { DocsQA, defaultChunkMany, DefaultInMemoryVectorStore } from 'ai-jsx/batteries/docs';
-import { OpenAIEmbeddings } from 'langchain/embeddings/openai';
+import { JsonObject } from 'type-fest';
 import { Article } from './load-articles.mjs';
-import _ from 'lodash';
-import { showInspector } from 'ai-jsx/core/inspector';
 
 const dirname = path.dirname(fileURLToPath(import.meta.url));
 
+interface RedditMetadata extends JsonObject {
```
NickHeiner, line 14: I don't think we want `extends JsonObject` here unless you want to let `RedditMetadata` have arbitrary other properties.

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
```diff
@@ -1,47 +1,54 @@
...
+      name: dataFile,
       metadata: rest,
-    };
+    } as Document<RedditMetadata>;
   })
 );
 
-const chunkedDocs = await defaultChunkMany(docs);
-const vectorStore = await DefaultInMemoryVectorStore.fromDocuments(chunkedDocs, new OpenAIEmbeddings());
+const corpus: Corpus = new LocalCorpus(CorpusLoading.staticLoader(docs, 10), defaultChunker);
```
NickHeiner, line 34:
```suggestion
const corpus = new LocalCorpus(CorpusLoading.staticLoader(docs, 10), defaultChunker);
```

Is the annotation necessary? Typically we want to let type inference do most of the work and only add type inference as necessary.

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
```diff
@@ -1,47 +1,54 @@
...
+await corpus.startCrawl();
```
NickHeiner, line 35: This is interesting. Should `DocsQA` call `startCrawl` itself? What are the scenarios / considerations about who should call `startCrawl`?

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: I get the angle about "this isn't related to a render", but I think we want the ability to write to the common log stream, even outside of things related to a render, like what we see here.

The point about "maybe this belongs outside of AI.JSX and in a Fixie SDK" is a good one – maybe eventually we'll split this type of logic out into a separate module?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+}
 
 /**
- * Load Documents from a source.
+ * Objects related to loading a corpus. This is facilitated by a series of
+ * requests and corresponding responses.
+ *
+ * In addition to returning documents, each response may expand the corpus
+ * space in one or both of two dimensions:
+ *      Responses may include new partitions to be loaded. Partitions are
```
NickHeiner, line 105: Can we define `partition` before using it? I think I get from the context what it means, but I'm not totally sure. I think it would be helpful to define up front the relationship between documents, pages, partitions, etc.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: I think we probably don't need `namespace`, and can just rely on the file itself as a way to group types. We have a bunch of types in `src/index.ts`, and they're not super tightly grouped, but it's been serving us fine.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: (This applies to a bunch of other `class`es in this file as well.)

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
-  document: Document<Metadata>;
-  score: number;
-}
+  export class Response<DocType extends RawDocument | Document> {
+    constructor(readonly page?: Page<DocType>, readonly partitions?: Partition[]) {
+      /** A page of documents from the requested partition. */
+      this.page = page;
+      /**
+       * Additional partitions that should be loaded in subsequent requests.
+       * These will be ignored if they were already included in a previous
```
NickHeiner, line 183: What defines partition identity? The `name`?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+       */
+      this.nextPageToken = nextPageToken;
+    }
+  }
 
-  search<Filter = unknown>(query: string, opts: { filter?: Filter; limit: number }): Promise<VectorSearchResult[]>;
-}
+  export class Partition {
+    constructor(readonly name: string, readonly firstPageToken?: string) {
+      /** The unique name of this partition. */
```
NickHeiner, line 204:
```suggestion
      /** The unique name of this partition. Two partitions are considered equivalent if they have the same name. */
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+  class StaticLoader<DocumentMetadata extends Jsonifiable = Jsonifiable> {
+    constructor(readonly documents: Document<DocumentMetadata>[], readonly pageSize?: number) {}
+
+    load(request: Request): Promise<Response<Document<DocumentMetadata>>> {
+      if (this.pageSize) {
+        const page: number = Number(request.pageToken ?? 0);
+        const start = page * this.pageSize;
+        const end = start + this.pageSize;
+        if (end > this.documents.length) {
+          return Promise.resolve(new Response(new Page(this.documents.slice(start))));
```
NickHeiner, line 242: You can mark `load` as `async`, and then the function will automatically return a promise, so you won't need the manual `Promise.resolve`s everywhere.

NickHeiner, line 242:
```suggestion
          return Promise.resolve(new Response(new Page([...documents])));
```

If you're just trying to clone the array, this is a cleaner way to do it IMO.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: Our general approach has been to disable all ESLint style-related rules and let prettier own it completely. But it's not a big deal either way.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
 
+/** A function that splits a document into multiple chunks, for example to ensure it fits in appropriate context windows. */
+export type Chunker<
+  DocumentMetadata extends Jsonifiable = Jsonifiable,
+  ChunkMetadata extends Jsonifiable = Jsonifiable
+> = (document: Document<DocumentMetadata>) => Promise<Chunk<ChunkMetadata>[]>;
+
 /**
- * If you need to chunk your documents, start with this chunker. Move to more powerful chunkers only as needed.
+ * A simple token size based text chunker. This is a good starting point for text documents.
```
NickHeiner, line 267: Heh, I like your phrasing. It's more approachable than what I wrote.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
 
-  return splitter.splitDocuments([
-    // TS doesn't realize that we've ensured that the metadata is defined.
-    // @ts-expect-error
-    doc,
-  ]);
+  return chunks;
 };
 
+/** A chunk is a piece of a document that's appropriately sized for an LLM's context window and for semantic search. */
```
NickHeiner, line 289: For `context window`, you can link to https://docs.ai-jsx.com/guides/brand-new#context-window.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+  /** The name of the document from which this chunk was extracted. */
+  documentName?: string;
+
+  /** Optional additional metadata associated with this chunk. */
+  metadata?: ChunkMetadata;
+}
+
 /**
- * If you need to chunk your documents, start with this chunker. Move to more powerful chunkers only as needed.
+ * An embedding is a function that maps a string to a vector encoding its semantic meaning.
```
NickHeiner, line 302: For embedding, let's link to https://docs.ai-jsx.com/guides/brand-new#semantic-similarity-embeddings.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+export const defaultEmbedding = new LangChainEmbeddingWrapper(new OpenAIEmbeddings());
 
 /**
- * A cast wrapper to convert an AI.JSX document to a LangChain document.
+ * An embedded chunk is a piece of a document that is ready to be added into a vector space.
  */
-export function toLangChainDoc(doc: Document): LangChainDocument {
-  return doc as LangChainDocument;
+export interface EmbeddedChunk<ChunkMetadata extends Jsonifiable = Jsonifiable> {
+  /** The identifier for this chunk. For insertions, this will be generated if not populated.*/
```
NickHeiner, line 328: What are insertions?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+> implements Corpus<ChunkMetadata>
+{
+  /* eslint-enable */
+  private readonly vectors: EmbeddedChunk<ChunkMetadata>[] = [];
+  private readonly documents: Document<DocumentMetadata>[] = [];
+  private readonly completedPartitions = new Set<string>();
+  private readonly activePartitionsToToken = new Map<string | null, string | null>();
+  private loadingState = Corpus.LoadingState.NOT_STARTED;
+  constructor(
+    readonly loader: CorpusLoading.Loader<DocumentMetadata>,
```
NickHeiner, line 392: There's syntax sugar you can use when the `constructor` is just setting instance vars:

https://www.typescriptlang.org/play?#code/MYGwhgzhAEBi0G8BQ1rAPYDsIBcBOArsDungBQqrQAOeAlgG5g4Cm0eLYAJliAJ7QARmDwAuIenQhOmSgEpEAXyTKkGbDmgAzaAF5omFgHc4ZfARZyA3EA

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+    pinoLogger.info({ documents: docs.length }, 'Starting chunking for documents');
+    const chunks: Chunk<ChunkMetadata>[] = [];
+    await Promise.all(docs.map((doc) => this.chunker(doc).then((docChunks) => chunks.push(...docChunks))));
+    pinoLogger.info({ chunks: chunks.length }, 'Chunking completed. Beginning embed.');
+    const vectors: number[][] = await this.embedding.embedBatch(chunks.map((chunk) => chunk.content));
+    pinoLogger.info({ vectors: vectors.length }, 'Embedding completed.');
+    return chunks.map((chunk, i) => ({ ...chunk, vector: vectors[i] } as EmbeddedChunk<ChunkMetadata>));
+  }
+
+  getStats(): Promise<Corpus.Stats> {
```
NickHeiner, line 468: Why does this need to return a promise?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+export class LocalCorpus<
+  DocumentMetadata extends Jsonifiable = Jsonifiable,
+  ChunkMetadata extends Jsonifiable = Jsonifiable
+> implements Corpus<ChunkMetadata>
+{
+  /* eslint-enable */
+  private readonly vectors: EmbeddedChunk<ChunkMetadata>[] = [];
+  private readonly documents: Document<DocumentMetadata>[] = [];
+  private readonly completedPartitions = new Set<string>();
+  private readonly activePartitionsToToken = new Map<string | null, string | null>();
```
NickHeiner, line 389: If the key of this map is meant to be the partition name, you can make that more explicit with:

```suggestion
  private readonly activePartitionsToToken = new Map<Partition['name'] | null, string | null>();
```

(Same logic applies to the token.)

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+    readonly loader: CorpusLoading.Loader<DocumentMetadata>,
+    readonly chunker: Chunker<DocumentMetadata, ChunkMetadata>,
+    readonly embedding: Embedding = defaultEmbedding
+  ) {
+    this.loader = loader;
+    this.chunker = chunker;
+    this.embedding = embedding;
+  }
+
+  async startCrawl(): Promise<Corpus.Stats> {
```
NickHeiner, line 401: This function name is a bit confusing to me. I would expect `await startCrawl()` to mean the crawl had begun. But actually, if we `await` the result of `startCrawl()`, it means the crawl has completed.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+    }
+    return this.getStats();
+  }
+
+  private async doCrawl(): Promise<void> {
+    pinoLogger.info('Starting crawl');
+    this.activePartitionsToToken.set(null, null);
+    while (this.activePartitionsToToken.size > 0) {
+      const partition = this.activePartitionsToToken.keys().next().value;
+      const token = this.activePartitionsToToken.get(partition);
```
NickHeiner, line 421: If you want the key and the value, you can use `map.entries()`:

```
const [partition, token] =  this.activePartitionsToToken.entries().next().value
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+      pinoLogger.info(
+        {
+          partitions: response.partitions?.length ?? 0,
+          documents: response.page?.documents.length ?? 0,
+          nextPage: response.page?.nextPageToken != undefined,
+        },
+        'Received response'
+      );
+      for (const newPartition of response.partitions ?? []) {
+        if (!this.completedPartitions.has(newPartition.name) && !this.activePartitionsToToken.has(newPartition.name)) {
```
NickHeiner, line 434:
```suggestion
        if (!(this.completedPartitions.has(newPartition.name) || this.activePartitionsToToken.has(newPartition.name))) {
```

minor

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+        pinoLogger.info('Completed vectorization');
+        this.vectors.push(...vectors.flat());
+        if (response.page.nextPageToken) {
+          this.activePartitionsToToken.set(partition, response.page.nextPageToken);
+        } else {
+          this.completedPartitions.add(partition);
+        }
+      }
+      if (!response.page && !response.partitions) {
+        // Corner case - invalid response.
```
NickHeiner, line 451: I'm slightly concerned about a seemingly-silent failure here. Is there any legit case where this would happen? If not, it seems better to throw.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+        this.completedPartitions.add(partition);
+      }
+    }
+    pinoLogger.info('Crawl complete');
+  }
+
+  private async vectorize(docs: Document<DocumentMetadata>[]): Promise<EmbeddedChunk<ChunkMetadata>[]> {
+    pinoLogger.info({ documents: docs.length }, 'Starting chunking for documents');
+    const chunks: Chunk<ChunkMetadata>[] = [];
+    await Promise.all(docs.map((doc) => this.chunker(doc).then((docChunks) => chunks.push(...docChunks))));
```
NickHeiner, line 461: I would write this as:
```suggestion
    await Promise.all(docs.map(async (doc) => {
      const docChunks = await this.chunker(doc);
      chunks.push(...docChunks)
    }));
```

Not a big deal either way though.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+  chunk: EmbeddedChunk<ChunkMetadata>;
+  score: number;
+}
+
+export interface Corpus<ChunkMetadata extends Jsonifiable = Jsonifiable> {
+  startCrawl: () => Promise<Corpus.Stats>;
+
+  getStats: () => Promise<Corpus.Stats>;
+
+  search: (
```
NickHeiner, line 355: It would be great to get `/** */` comments on all these.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -1,154 +1,503 @@
...
+    query: string,
+    params?: { limit?: number; score_threshold?: number }
+  ) => Promise<ScoredChunk<ChunkMetadata>[]>;
+}
+
+export namespace Corpus {
```
NickHeiner, line 361: I think it might cause confusion in either the tooling or the humans to have a `namespace` and an `interface` with the same name.

NickHeiner (approved):
This is a great expansion of our docs capability.

I think the main thing I'd ask to make this a bit more robust is some entries in `examples` that exercise the various expanded cases we've added here (various partitions, etc). I feel like the reddit case is only the simplest happiest path.

NickHeiner:
Use your discretion on all my comments; nothing is blocking.

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
mdepinet: I added this to satisfy the constraint that RedditMetadata is a Jsonifiable.  Is there a better way to do this?

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
mdepinet: No, just ensuring I'm not making references to anything that exists on the concrete type but not the interface.  Removed.

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
mdepinet: It could. I decided against that because loading a corpus feels like something that should be handled separately from actually building your prompts/UI (i.e. it needs to be done ahead of time in general). Having DocsQA call startCrawl if the corpus isn't loaded seems like it'd steer users toward loading corpora just in time instead.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: lol That's what I wrote initially, but the linter complained.  I switched back and suppressed the linter for this function.

mdepinet: It's not a clone, it's a sublist.  Am I missing something?

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: Clarified in comments. This is more to support the Fixie implementation than the existing one.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: Clarified this in comments.  In general awaiting the result of startCrawl won't mean that the crawl has completed. That just happens to be the case for LocalCorpus.

NickHeiner:
> > To fix the CI error, run `yarn` locally and check in the changes to `yarn.lock` that result.
> 
> Hmm... Running yarn locally didn't result in any changes. Could I have something misconfigured in my environment?

Hmm, this surprises me. Let's see how things change after you pull in the latest `main`, since it changes some yarn things.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: Ok, but what about the former case? Is that a malformed response?

My only concern here is if the system is silently failing for a malformed response. If all the cases are valid responses, then what we have is fine.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: oh whoops no, I was misreading `start` as `0`.

NickHeiner: I would just disable that lint rule. But it's fine either way.

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
NickHeiner: Yep! 

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwFIGcIDsDyAjAVgUwGMYBfOAMyghDgHIYBPMfAWnPxRloG4AoX4Ohj4o5AIaF8cAEpx8AD2HoAJijioMOAsUS84FCBABccbIYA2+Mel4l+youbFQphDJzhQT0vrwD0fnAA6lIA7tbwMAAWwGowEHJQVFAANKYArvDA8MoQHHDoEDBp2ERi6ShSAAYKSqrqaFh4RDBVcCBW6HFRIq7WplLO2NlQzgxwANb4DCgAdLxQs2IoyuS+jMxwALIVMABC+BpN2jAAPAAqAHxwALxw53KK+CpqR1otcAD893Am6PgANxE6yYUmkACZbttdgc3s1iKcEHoDMZTBZOrZLvZHM5XO54FBwd5wb4AvcYmoRMkUGllnBQvhaMo4FEIMx5oSlis1vwydIAMz04DmcxwNwdOAQTJweIDGWguBVf5AqBtazM6IVOXpdAVMTYSwlMoVKTZChiYXdVw9QgTeYbMGCu47TiwxrvBFI-TkQwmMwQSzWFLI8joEwACgAlLdrgg7CQsbwgA

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: I'm somewhat skeptical of treating an empty response and a non-empty response whose fields are empty as being different in general, but you've convinced me.  Updated.

PR #48, opened 2023-06-15
Switch to yarn PnP

petersalas:
This switches the `yarn` linker to PnP. The impetus is to avoid this problem: https://github.com/yarnpkg/yarn/issues/8540#issuecomment-756093588

(CRA sets restrictions on what aliases can be used within webpack, so I wasn't able to get the webpack workaround to work.)

There are a few side-effects of this change:
- We need to be more explicit about our dependencies in a few places where we were importing transitive dependencies
- VSCode needs to be configured to use `yarn`'s version of things as per https://yarnpkg.com/getting-started/editor-sdks
- `eslint-plugin-import` needs to use https://yarnpkg.com/features/pnp#fallback-mode as per https://psidium.github.io/lerna/monorepo/yarn/nodejs/pnp/zero-install/2021/08/23/migrating-a-monorepo-from-lerna-to-yarn-3.html
- `tsx` doesn't work with PnP on Node 18 until https://github.com/nodejs/node/pull/43772 is backported, so this switches the example package away from `tsx`
- I needed to inject some fake `peerDependencies` into `postcss-loader` via `packageExtensions`


.yarn/sdks/eslint/bin/eslint.js
NickHeiner, line 1: It feels a little funny to me that we're supposed to check this in, but that seems to be the guidance, so 🤷

packages/create-react-app-demo/package.json
```diff
@@ -36,6 +36,7 @@
     "jest-resolve": "^27.4.2",
     "jest-watch-typeahead": "^1.0.0",
     "jotai": "^2.2.0",
+    "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
petersalas: Yup exactly. I think these changes are Good

.yarn/sdks/eslint/bin/eslint.js
petersalas: Yeah...and their "zero-install" would mean checking in the dependencies themselves which I'm not quite ready to commit to.

NickHeiner (approved):

PR #51, opened 2023-06-15
Rework React integration

petersalas:
This refactors the React integration and applies it to the Recipe example in `create-react-app`.

- Use ES6 proxies instead of a global `WeakMap` to link `React` elements to `AI` elements
- The integration is used by doing `import * as AI from "ai-jsx/react"` and setting the JSX factory to be `AI.createElement`
- There are two "boundary" components: `AI.jsx` and `AI.React` which transition to and from AI.jsx respectively
- If an `AI.React` component is forced to be rendered in AI.jsx, it renders to a serialized version of the React tree. If it's rendered within React it is equivalent to `React.Fragment`.
- If an `AI.jsx` component is rendered in React, it manages the AI render context via an effect. If it's rendered with AI.jsx, it's equivalent to an `AI.Fragment`

This takes a stab at splitting up the Recipe example into two layers:
 - An LLM invocation that produces a recipe
 - An LLM invocation that creates UI that represents that recipe (implemented by `UICompletion`)

I'm not particularly wedded to this layering and expect it to change -- I mainly wanted to move the UI serialization/deserialization to be within `ai-jsx/react` in a quasi-generic way.

I renamed the Next.js `AI` component to `AIComponent` to avoid confusion with the new `AI` import but otherwise didn't change it. We need to do some Babel surgery to let us override the JSX factory, and create a different `AI.jsx` RSC.


petersalas:
`packages/create-react-app-demo/src/recipe/page.tsx` is the "usage" example to take a look at.

packages/ai-jsx/.eslintrc.cjs
```diff
@@ -11,6 +11,8 @@ module.exports = {
 
   env: {
     node: true,
+    browser: true,
```
NickHeiner, line 14: We might want to be a little careful with this. This setting will enable browser globals, which means we could write code that depends on browser APIs, which then fails when the code runs in Node.

A more conservative approach would be to enable `env:browser` in just files/directories that actually use it.

But this isn't worth spending more than 60 seconds thinking further about or addressing.

packages/ai-jsx/src/index.ts
```diff
@@ -15,8 +15,9 @@ export interface Element<P> {
   [attachedContext]?: RenderContext;
 }
 
-interface IndirectNode {
-  _magic: never;
+const indirectNodeSymbol = Symbol('AI.indirectNode');
+export interface IndirectNode {
```
NickHeiner, line 19: Can we get block comments on this (and, ultimately, everything else exported from this file)?

packages/ai-jsx/src/index.ts
```diff
@@ -180,17 +181,22 @@ export function createContext<T>(defaultValue: T): Context<T> {
...
 }
 
-const indirectWeakMap = new WeakMap<IndirectNode, Node>();
 export function isIndirectNode(value: unknown): value is IndirectNode {
-  return indirectWeakMap.has(value as IndirectNode);
+  if (value !== null && typeof value === 'object') {
+    return indirectNodeSymbol in value;
+  }
+  return false;
 }
```
NickHeiner, line 189: I would tend to write this as:

```
return (value !== null && typeof value === 'object' && indirectNodeSymbol in value;
```

packages/ai-jsx/src/index.ts
```diff
@@ -180,17 +181,22 @@ export function createContext<T>(defaultValue: T): Context<T> {
...
 
 export function getIndirectNode(value: IndirectNode): Node {
```
NickHeiner, line 191: This name is a little funny, given that it says `getIndirectNode` but returns `Node`.

NickHeiner, line 191:
```suggestion
export function getNodeOfIndirectNode(value: IndirectNode): Node {
```

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  function toComponent(serializedComponent: SerializedComponent | string) {
+    if (!serializedComponent || typeof serializedComponent === 'string') {
+      return serializedComponent;
+    }
+
+    const Component = validComponents[serializedComponent.name];
+    if (!Component) {
+      logger.warn({ serializedComponent }, `Component not found for ${serializedComponent.name}`);
```
NickHeiner, line 77: This message won't be actionable on its own to the user. Can we include more detail about what the user can do to resolve this?

```suggestion
      logger.warn({ serializedComponent, validComponentNames: Object.keys(validComponents) }, `Component not found for ${serializedComponent.name}`);
```

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -0,0 +1,97 @@
...
+      return null;
+    }
+
+    if (!('children' in serializedComponent)) {
+      throw new Error(`Unrecognized JSON: ${JSON.stringify(serializedComponent)}`);
```
NickHeiner, line 82: This is unrecognized JSON produced by the model, right? Let's say that explicitly.

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -0,0 +1,97 @@
...
+      if ('children' in node.props) {
+        collectComponents(node.props.children, childrenAreReact);
+      }
+    }
+  }
+
+  collectComponents(example, true);
+
+  const modelResult = await render(
+    <ChatCompletion>
```
NickHeiner, line 42: Let's use @farzadab's json output constraint component here.

packages/ai-jsx/src/react/index.ts
```diff
@@ -0,0 +1,84 @@
...
+import * as LLMx from '../index.js';
+import { Serialize } from './serialize.js';
+export * from '../index.js';
+
+export declare namespace JSX {
+  type ElementType = ReactModule.JSX.ElementType | LLMx.JSX.ElementType | typeof React | typeof jsx;
+  type Element = ReactModule.JSX.Element & LLMx.Node;
+  type IntrinsicElements = ReactModule.JSX.IntrinsicElements;
+  type ElementChildrenAttribute = ReactModule.JSX.ElementChildrenAttribute & LLMx.JSX.ElementChildrenAttribute;
+}
```
NickHeiner, line 11: When people import this file, I wonder if it'll cause TS to now be more permissive than they perhaps want. Like, in vanilla React, you can't do:

```tsx
<ReactComponent />
<AIComponent />
```

Once you import our types here, you can. But it also lets you put AI Components where they don't belong, right?

Even if that's true, this is still a net win and something we should move forward with. But we might want to call out the caveat.

packages/ai-jsx/src/react/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  ReactModule.useEffect(() => {
+    let shouldStop = false;
+    async function stream() {
+      setResult([]);
+      setIsDone(false);
+
+      // TODO: add a way for a render context to be aborted
+      const renderResult = LLMx.createRenderContext().render(children, {
+        stop: (e) => e.tag == React,
+        map: (frame) => frame,
```
NickHeiner, line 39: Why do we need a map if it's an identity function?

packages/ai-jsx/src/react/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      });
+      for await (const frame of renderResult) {
+        if (shouldStop) {
+          return;
+        }
+
+        setResult(frame.map((e) => (LLMx.isElement(e) ? e.props.children : e)) as ReactModule.ReactNode);
```
NickHeiner, line 46: So we have an invariant that every element will either be a React node, or its direct children will be an array of React nodes? Or is there something else going on?

packages/ai-jsx/src/react/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    return () => {
+      shouldStop = true;
+    };
+  }, [children]);
+
+  return { result, isDone };
+}
+
+export function jsx({ children }: { children: LLMx.Node }, context?: any | LLMx.ComponentContext) {
```
NickHeiner, line 67: We need block comments on all of these. :smile:

packages/ai-jsx/src/react/serialize.tsx
```diff
@@ -0,0 +1,30 @@
...
+export function Serialize({ children }: { children: React.ReactNode }): AI.Renderable {
+  if (Array.isArray(children)) {
+    return children.map((child) => <Serialize>{child}</Serialize>);
+  }
+
+  const child = children;
+  if (React.isValidElement(child) && child.type !== AI.jsx) {
+    // Serialize the React element and any children.
+    const typeName = reactComponentName(child.type);
+    // TODO: support prop serialization
```
NickHeiner, line 21: Yeah, and we don't support prop application either right now, right? That's fine; we should just call it out as a known caveat.

packages/create-react-app-demo/src/recipe/page.tsx
```diff
@@ -78,33 +79,31 @@ export default function RecipeWrapper() {
...
-              {'}'}
-              For example:
-              {'{'}
-              "name": "Recipe", "children": [{'{'}"name": "RecipeTitle", "children": ["My Recipe"]{'}'}
-              "my description" ]{'}'}. Respond with only the JSON. Do not include with an explanatory suffix or prefix.
-            </SystemMessage>
-            <UserMessage>Give me a recipe for {query}. Respond with only the JSON.</UserMessage>
-          </ChatCompletion>
-        </AI>
+        <AI.jsx>
```
NickHeiner, line 82: damn is this the best thing I've ever seen?

packages/create-react-app-demo/src/recipe/page.tsx
```diff
@@ -78,33 +79,31 @@ export default function RecipeWrapper() {
...
+          <UICompletion
+            example={
```
NickHeiner, line 84: Splitting out the example as a separate prop, rather than the user mixing it in with everything else, is an interesting approach. I like it.

packages/docs/docs/guides/ai-ui.md
NickHeiner, line 1: We're going to need to rework this doc in a pretty serious way but I can do that later.

NickHeiner:
I notice this drops the loading spinner from the demo – was it hard to preserve it?

image

NickHeiner:
This doesn't need to be a blocker, but I'm a bit confused by the logging behavior.

Here are the logs I saw:

[localhost-1686887943261.log](https://github.com/fixie-ai/ai-jsx/files/11765863/localhost-1686887943261.log)

It's a bunch of:

```
browser.js:228 Object 'Got delta message'
```

This confuses me, because I thought the default logger was a no-op, and this was the relevant render, and it's not defining a logger:

```
      // TODO: add a way for a render context to be aborted
      const renderResult = LLMx.createRenderContext().render(children, {
        stop: (e) => e.tag == React,
        map: (frame) => frame,
      });
```

Separately, the last log line was:

```
browser.js:228 Objectelement: "<UICompletion>"renderId: "da6abce2-4430-4988-aa92-edfa151a48c7"serializedComponent: children: Array(5)0: {name: 'h1', children: Array(1)}children: ['Ingredients:']name: "h1"[[Prototype]]: Object1: {name: 'ul', children: Array(8)}2: {name: 'h2', children: Array(1)}3: {name: 'ol', children: Array(10)}4: {name: 'p', children: Array(1)}length: 5[[Prototype]]: Array(0)name: "div"[[Prototype]]: Object[[Prototype]]: Object 'Component not found for div'
LOG @ browser.js:228
```

Looks like the model gave invalid output – we should propagate this as an error we can display in the UI. Otherwise, the UI just looks like it's still in our non-existent "loading state":

image

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -0,0 +1,97 @@
...
+      return serializedComponent;
+    }
+
+    const Component = validComponents[serializedComponent.name];
+    if (!Component) {
+      logger.warn({ serializedComponent }, `Component not found for ${serializedComponent.name}`);
+      return null;
+    }
+
+    if (!('children' in serializedComponent)) {
```
NickHeiner, line 81: The more robust thing to do here would be to define a Zod schema and fully validate it. It would also be nice if we could serialize the Zod schema directly into the prompt.

I think @farzadab was experimenting with the some part of this and it didn't work well, so maybe I'm making a bad suggestion. :smile:

NickHeiner (approved):

NickHeiner:
When the model fails to produce valid JSON, we get:

image

The model will continue to be squirrely and we won't fix that as part of this PR, but it would be nice to throw a more explicit error.

packages/create-react-app-demo/src/recipe/page.tsx
```diff
@@ -78,33 +79,31 @@ export default function RecipeWrapper() {
       <InputPrompt label="What would you like a recipe for?" value={query} setValue={setQuery} />
 
       <ResultContainer title={`AI comes up with a recipe for ${query}`}>
```
NickHeiner, line 81: I added this:

```tsx
      <ResultContainer title={`AI comes up with a recipe for ${query}`}>
        <AI.jsx>
          <ChatCompletion>
            <SystemMessage>You are a master, verbose story-teller</SystemMessage>
            <UserMessage>Tell me a story about someone who eats {query}</UserMessage>
          </ChatCompletion>
        </AI.jsx>
      </ResultContainer>
```

And it worked great!

packages/ai-jsx/src/react/completion.tsx
farzadab: It works, but I just didn't find the abstraction too clean. I can probably make it better, but if you're the one designing it, I think the schema can work well.

packages/ai-jsx/src/react/index.ts
petersalas: A while back I decided that yielding the stopped elements by default was a footgun that shouldn't be the default. So the default `map` is _not_ identity, but rather "filter out elements matched by `stop` and `.join("")` the rest". Will rework a bit to make it somewhat more idiomatic. (The stopping thing is what makes things awkward which should ultimately move down into core.)

packages/ai-jsx/src/react/index.ts
petersalas: Right, because we have
```
stop: (e) => e.tag == React,
```
the only elements will be `AI.React` elements whose `children` prop is treated as a `ReactNode`.

packages/ai-jsx/src/react/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (typeof context?.render === 'function') {
+    // We're in AI.JSX already.
```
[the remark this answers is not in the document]
petersalas: Yeah I don't think there's a scenario where it's necessary, but it's a trivial way to relax "transition from React to AI.jsx" into "makes AI.jsx nodes valid" so I went for it. (And I did something similar with `AI.React` to allow _either_ `example={<AI.React><Recipe /></AI.React>}` or `example={<Recipe />}`.)

packages/ai-jsx/src/index.ts
petersalas: Added some to (most of) the new stuff

packages/ai-jsx/src/react/serialize.tsx
petersalas: I didn't end up adding anything just because I think this is something we almost certainly want to fix Soon

petersalas:
re: spinner, I feel like the right way to do this is to add a suspense-like fallback prop on `AI.jsx`? unless there's a way to get it to work directly with `Suspense` but that seems finicky atm.

Deferring this for now

petersalas:
oops, I meant to auto-merge not merge now

PR #52, opened 2023-06-15
Clean up some docs around the examples.

mdwelsh:
This PR just clarifies how to run the examples. It also updates the ai-jsx dependencies within the tree to be relative, and also fixes cases where the yarn scripts were referring to llmx.log instead of ai-jsx.log.


packages/ai-jsx/readme.md
```diff
@@ -1,8 +1,6 @@
-<!-- This is the README that shows up on the npm package page: https://www.npmjs.com/package/ai-jsx -->
-
 # AI.JSX
 
-AI.JSX is a framework for building AI-native apps with React.
+AI.JSX is a framework for building AI-native apps with JSX.
```
NickHeiner, line 3: lol, whoops.

packages/create-react-app-demo/package.json
```diff
@@ -12,7 +12,7 @@
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
     "@testing-library/user-event": "^13.5.0",
-    "ai-jsx": "0.2.0-4",
+    "ai-jsx": "file:../ai-jsx",
```
NickHeiner, line 15: Ah, is this the better way to do the monorepo than needing to make sure the versions line up?

packages/examples/README.md
```diff
@@ -0,0 +1,25 @@
...
+```
+$ yarn workspace examples run demo:debate
+```
+
+to run the `debate` example.
+
+You will need to get an API key for OpenAI from the [OpenAI API dashboard](https://platform.openai.com/account/api-keys) and set the `OPENAI_API_KEY` environment variable, in order for the
+examples to invoke the OpenAI language models.
+
+These apps are currently run directly with Node and do not have a web UI of their own.
```
NickHeiner, line 24:
```suggestion
These apps run directly with Node and do not have a web UI of their own.
```

NickHeiner (approved):
Looks great. Thanks for the cleanup.

packages/create-react-app-demo/package.json
mdwelsh: Yes, but it does not always work for some reason - reverting this change for now.

PR #54, opened 2023-06-16
Update logo images.

mdwelsh:
These are on Figma and we can tweak them if needed.


mdwelsh:
Gonna YOLO this, can revert if you hate it

PR #55, opened 2023-06-16
Dependency bugfix and prompt updates for examples

farzadab:
During the demo I found a few issues and updated some prompts for the examples.


NickHeiner (approved):
Thanks!

PR #56, opened 2023-06-16
Docs Search

farzadab:
We have local search!

Note: if the docs site grows we might need a hosted solution, but this seems to work for now.

Just go to the preview: https://ai-jsx-docs-git-farzad-docs-search-fixie-ai.vercel.app/

You should see the search (⌘+K for shortcut):
image

### DIFF
--- a/packages/docs/docusaurus.config.js
+++ b/packages/docs/docusaurus.config.js
@@ -111,14 +111,14 @@ const config = {
 
   themes: [
     [
-      require.resolve("@easyops-cn/docusaurus-search-local"),
+      require.resolve('@easyops-cn/docusaurus-search-local'),
       /** @type {import("@easyops-cn/docusaurus-search-local").PluginOptions} */
       ({
         // ... Your options.
         // `hashed` is recommended as long-term-cache of index file is possible.
         hashed: true,
         indexBlog: false,
-        // docsDir: ".",
+        docsRouteBasePath: '/',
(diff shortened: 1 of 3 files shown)


NickHeiner:
image

Kinda funky that there are duplicate results but this is still 1000x better than nothing. :smile:

NickHeiner (approved):

NickHeiner:
If local works for now, I'm strongly in favor of moving forward with that, and deferring an Algolia discussion for later.

farzadab:
I looked through the [options](https://github.com/easyops-cn/docusaurus-search-local#theme-options) but I don't see how I can change that. I enabled highlighting the search word in that page though.

I'll merge once the checks pass.

NickHeiner:
Yeah the highlighting seems good.

Do you know why we get the duplicate results? Is it a bug in Docusaurus?

farzadab:
Not sure. There are two distinct matches to the search word "image"; one in the title and one in the text. I feel like it's treating them differently. In this case I agree that it should be aggregating them into the same record anyway.

You can see it matching to the same page in other samples too. If you search for "memo", I actually feel like it's doing the right thing because the matches are far enough apart. I don't know really.

image
image

NickHeiner:
Makes sense, and I agree it's doing the right thing when there's a lot of space between the hits. So I feel less bad about it happening.

PR #57, opened 2023-06-16
Add Google Analytics to docs site.

mdwelsh:


packages/docs/docusaurus.config.js
```diff
@@ -110,13 +114,6 @@ const config = {
         darkTheme: darkCodeTheme,
       },
     }),
-
-  scripts: [
-    {
-      src: '/js/pendo.js',
```
NickHeiner, line 116: Let's keep Pendo on the site. If we ever do go back to them, we'll have full retroactive analytics, even if we hadn't been paying at the time.

NickHeiner (approved):

PR #58, opened 2023-06-16
Expose OpenAI's latest FunctionCall to ChatCompletion

hessamb:
To test:
```
yarn workspace examples run demo:chat-function-call
```


packages/ai-jsx/src/core/completion.tsx
```diff
@@ -14,6 +16,18 @@ export type ModelPropsWithChildren = ModelProps & {
 
 export type ModelComponent<T extends ModelPropsWithChildren> = Component<T>;
 
+export type FunctionDefinition = {
```
hessamb, line 19: One thing I don't like about this PR is that I have function definitions as props to ChatCompletion as opposed to being children.

I chose to set them as properties because it was hard to parse a nested jsx structure in the children into these types. However, I think it would look much better if they were children because 1) It looks like a more elegant JSX syntax, and 2) A subtree can generate both "SystemMessage" and "FunctionDefinition" for the consumption of this parent.

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -14,6 +15,18 @@ export type ModelPropsWithChildren = ModelProps & {
...
+  name: string;
+  description?: string;
+  parameters: Record<string, FunctionParameter>;
+}
+
+export interface FunctionParameter {
+  description?: string;
+  type?: string;
+  required: boolean;
+}
```
NickHeiner, line 28: Hmm, the OpenAI module doesn't export these types?

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -142,6 +155,49 @@ export function AssistantMessage({ children }: { children: Node }) {
   return children;
 }
 
+/**
+ * Provide a Function Call to the chat model.
+ *
+ * The function call tells the model a previous function that was invoked by the model. See https://platform.openai.com/docs/guides/gpt/chat-completions-api for more detail.
```
NickHeiner, line 161:
```suggestion
 * The function call tells the model that a function was previously invoked by the model. See https://platform.openai.com/docs/guides/gpt/chat-completions-api for more detail.
```

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -142,6 +155,49 @@ export function AssistantMessage({ children }: { children: Node }) {
...
+ * In case of model returning a function call, ChatCompletion would return a <FunctionCall> component as well.
```
NickHeiner, line 162:
```suggestion
 * When the model returns a function call, <ChatCompletion> returns a <FunctionCall> component.
```

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -142,6 +155,49 @@ export function AssistantMessage({ children }: { children: Node }) {
...
+ *
+ * @example
+ *    <ChatCompletion>
+ *      <UserMessage>What is 258 * 322?</UserMessage>
+ *      <FunctionCall name="evaluate_math" args={expression: "258 * 322"} />
+ *      <FunctionResponse name="evaluate_math">83076</FunctionResponse>
+ *
+ *    ==> "That would be 83,076."
+ */
+export function FunctionCall({ name, args }: { name: string; args: Record<string, string | number | boolean | null> }) {
```
NickHeiner, line 174: Functions can only be called with `string | number | boolean`? The model can't call a function by passing an object? e.g. `f({a: 123})`

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -142,6 +155,49 @@ export function AssistantMessage({ children }: { children: Node }) {
...
+  return `Call function ${name} with ${JSON.stringify(args)}`;
+}
+
+/**
+ * Provide a Function Response to the chat model.
+ *
+ * The function response tells the output of a previous <FunctionCall /> request. See https://platform.openai.com/docs/guides/gpt/chat-completions-api for more detail.
```
NickHeiner, line 181:
```suggestion
 * The function response provides the output of a previous <FunctionCall /> request. See https://platform.openai.com/docs/guides/gpt/chat-completions-api for more detail.
```

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -215,11 +219,20 @@ export async function* OpenAICompletionModel(
...
   { render, getContext, logger }: LLMx.ComponentContext
 ) {
   const messageElements = await render(props.children, {
-    stop: (e) => e.tag == SystemMessage || e.tag == UserMessage || e.tag == AssistantMessage,
+    stop: (e) =>
+      e.tag == SystemMessage ||
+      e.tag == UserMessage ||
+      e.tag == AssistantMessage ||
+      e.tag == FunctionCall ||
+      e.tag == FunctionResponse,
```
NickHeiner, line 235: At this point, maybe we should switch to `[SystemMessage, UserMessage /* ... etc */].includes(e.tag)`, but no big deal.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -249,12 +277,32 @@ export async function* OpenAIChatModel(
...
+      parameters: {
+        type: 'object',
+        required: Object.keys(functionDefinition.parameters).filter(
+          (name) => functionDefinition.parameters[name].required
+        ),
+        properties: Object.keys(functionDefinition.parameters).reduce((map: Record<string, any>, paramName) => {
+          map[paramName] = {
+            type: functionDefinition.parameters[paramName].type,
+          };
+          return map;
```
NickHeiner, line 293:
```suggestion
          return { 
            ...map,
            [paramName]: {
              type: functionDefinition.parameters[paramName].type,
            }, 
            };
```

packages/examples/src/chat-function-call.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <UserMessage>{query}</UserMessage>
+        <FunctionCall name="evaluate_expression" args={{ expression: '2523231 * 2382382' }} />
+        <FunctionResponse name="evaluate_expression">6011300116242</FunctionResponse>
+      </ChatCompletion>
+    </ChatProvider>
+  );
+}
+
+showInspector(<ModelProducesFunctionCall query="What is 2523231 * 2382382?" />);
+showInspector(<ModelProducesFinalResponse query="What is 2523231 * 2382382?" />);
```
NickHeiner, line 65: `showInspector` takes over the whole terminal, so we don't want to run two at once. I'd just leave one of these commented out, then the user can decide which to enable. Or you can make each one its own demo.

packages/examples/src/chat-function-call.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <SystemMessage>You are a tool that may use functions to answer a user question.</SystemMessage>
+        <UserMessage>{query}</UserMessage>
+        <FunctionCall name="evaluate_expression" args={{ expression: '2523231 * 2382382' }} />
+        <FunctionResponse name="evaluate_expression">6011300116242</FunctionResponse>
+      </ChatCompletion>
+    </ChatProvider>
+  );
+}
+
+showInspector(<ModelProducesFunctionCall query="What is 2523231 * 2382382?" />);
```
NickHeiner, line 64: @petersalas from this example, we see the output is:

```tsx
<FunctionCall name="evaluate_expression" args={{"expression":"2523231 * 2382382"}} />
```

I was thinking the best way to use this from `UseTools` would be to `partialRender` through to a `FunctionCall`s, then inspect the `props` to figure out which funcs to invoke. Does that match what you were thinking? If so, I'm comfortable merging this PR as we have a clear path forward.

NickHeiner: I guess this is what `ChatCompletion` is doing already, and it seems good, so 👍

NickHeiner (approved):
Normally I'd run this myself and inspect the model input/output to make sure it matches my expectations, but for expediency I'll skip that step and trust you to have done it. :smile:

packages/examples/src/chat-function-call.tsx
hessamb: If I comment one out I get a warning for the unused function that fails the test

packages/ai-jsx/src/core/completion.tsx
NickHeiner: Can you show me a snippet of what the alternative would look like?

packages/examples/src/chat-function-call.tsx
NickHeiner: To work around that, you can either suppress that warning with an `// eslint-disable-next-line no-undef` comment, or you can make a separate file for each permutation.

Or, you could allow the user to pick on the command line, then use `process.argv` to decide which demo variant to run.

packages/ai-jsx/src/core/completion.tsx
hessamb: Something along these lines:
```
<ChatCompletion>
  <Function name="foo">
    This function does the blah operation
    <Parameter name="param1" type="string">
      This parameter controls the blah.
    </Parameter>
    <Parameter name="param2" type="integer">
      Another argument of the function.
    </Parameter>
  </Function>
</ChatCompletion>
```

packages/ai-jsx/src/core/completion.tsx
hessamb: OpenAI exposes a type (`ChatCompletionFunctions`) that is a bit incomplete (doesn't explicitly mark all settable fields for parameters; instead uses a Record<string, any>), and is optimized for API data transfer (there are some "type": "object" that need to be set on that Record<string, any> that is meaningless in terms of Function interface definition). My goal here is to decouple the AI.JSX function definition from the underlying API implementation (which may be OpenAI or otherwise).

packages/ai-jsx/src/core/completion.tsx
hessamb: Yes the model expects a flat list of _named_ parameters, each of which with a type like that. I didn't see any mention of the full list of supported types in the openai docs. I hypothesize whatever we pass as "type" (which is a string field) will be passed to the model, and the model may or may not do the right thing.

But since the model calls function by named parameters; maybe the javascript equivalent of that will become that gets all arguments as an object `func({param: value})`.

packages/ai-jsx/src/lib/openai.tsx
hessamb: That gives me some type mismatch error between Component<any> and some complex inferred type. I leave it for now :)

packages/examples/src/chat-function-call.tsx
hessamb: Thanks! I did the `//eslint` trick since this file is just for demonstration not a real demo.

hessamb:
Yes I tried a handful of the demos and they all seem to be healthy.

packages/ai-jsx/src/lib/openai.tsx
NickHeiner: Great. It would be a waste to spend more than 10 seconds trying to figure it out. :smile:

PR #60, opened 2023-06-16
Basic CSR React chat example 

juberti:
This is a basic chat example, cut down from the existing CYOA example, that is intended to be a good starting point for various conversational apps. Support for docsqa will be added in a similar example in an upcoming PR.


packages/create-react-app-demo/src/basic-chat/ai.tsx
```diff
@@ -0,0 +1,75 @@
+// @ts-nocheck
```
NickHeiner, line 1: This is no longer necessary after #51 landed.

NickHeiner:
Damn, gpt-3.5-turbo feels shockingly fast as I play around with this.

NickHeiner:
Looks great! I think someday we'll have richer abstractions for devs to use in a few places here.

NickHeiner (approved):

juberti:
agreed, I want to build a history component that should make things a bit simpler

PR #61, opened 2023-06-16
Removing clipLongPrompt from ImageGen

farzadab:
Removing the `clipLongPrompt` option from `<ImageGen>` component based on comments from @NickHeiner and @juberti.
The idea is to officially add `Summarizer` types and also use context to learn about how long the text should be in later PRs instead of the `clipLongPrompt` option.

### DIFF
--- a/packages/docs/docs/getting-started.md
+++ b/packages/docs/docs/getting-started.md
@@ -209,7 +209,7 @@ function WriteStoryWithImage() {
 }
 ```
 
-We use a second LLM pass to create a description for the image. We then feed this description to an `<ImageGen>` compoent.
+We use a second LLM pass to create a description for the image. We then feed this description to an `<ImageGen>` component.
 `<ImageGen>` by default uses [Dalle](https://platform.openai.com/docs/guides/images/introduction).
 
 Note that we also used a new component, `memo`. If we simply use `<WriteStory />` twice (even if you assign it to a variable), you will have two separate LLM calls and possibly two different stories as a result.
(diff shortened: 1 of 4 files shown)


packages/docs/docs/getting-started.md
```diff
@@ -186,23 +186,31 @@ $ grep -i 'starting modelcall' packages/ai-jsx/llmx.log | yarn workspace ai-jsx
...
-      Banner URL: <ImageGen clipLongPrompt>Generate an image for this story: {story}</ImageGen>
+      Banner URL:{' '}
+      <ImageGen>
+        <ChatCompletion>
+          <UserMessage>
+            You are an artist. You are creating a sketch for a new book. Concisely describe a sketch for scene from the
+            following story: {story}
+          </UserMessage>
+        </ChatCompletion>
+      </ImageGen>
```
NickHeiner, line 204: Yeah this works!

packages/docs/docs/getting-started.md
```diff
@@ -186,23 +186,31 @@ $ grep -i 'starting modelcall' packages/ai-jsx/llmx.log | yarn workspace ai-jsx
...
       {'\n\n'}
       {story}
     </>
   );
 }
 ```
 
-The `<ImageGen>` by default uses [Dalle](https://platform.openai.com/docs/guides/images/introduction).
-Since the story will likely be longer than the model's prompt length, we allow the model to clip it using the `clipLongPrompt` parameter.
+We use a second LLM pass to create a description for the image. We then feed this description to an `<ImageGen>` compoent.
```
NickHeiner, line 212:
```suggestion
We use a second LLM pass to create a description for the image. We then feed this description to an `<ImageGen>` component.
```

NickHeiner (approved):

juberti (approved):

PR #62, opened 2023-06-16
Add basic API docs generation.

mdwelsh:
This PR only adds the docusaurus-plugin-typedoc plugin and a basic configuration to add API docs to our docs site. We still need to, erm, actually *write* some of these docs, and the config might need to be tweaked to pull in more of the modules. But, it's a start.

### DIFF
--- a/packages/ai-jsx/package.json
+++ b/packages/ai-jsx/package.json
@@ -89,6 +89,7 @@
     "README.md"
   ],
   "devDependencies": {
+    "@tsconfig/node16": "^1.0.4",
     "@tsconfig/node18": "^2.0.1",
     "@types/eslint": "^8",
     "@types/js-yaml": "^4.0.5",
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@ dist
 .next
 .docusaurus
 *build*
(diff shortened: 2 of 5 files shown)


packages/ai-jsx/package.json
```diff
@@ -89,6 +89,7 @@
     "README.md"
   ],
   "devDependencies": {
+    "@tsconfig/node16": "^1.0.4",
```
NickHeiner, line 92: Is this necessary?

NickHeiner:
This is a good start. Most of the things Typedoc found are internal (e.g. `isIndirectNode`), and I don't see a ton of the external things (e.g. `ChatCompletion`). I think we just need to configure Typedoc to find the things we want / maybe suppress some of the things we don't.

Also, Docusaurus gives us a left-hand nav and a top nav. Right now, the top nav only says "docs".

image

Maybe with this, it's time to make "Guides" and "API" the two top nav items? That would allow us to reduce nesting in the LHS as well.

NickHeiner:
Also, moving "API" to the top nav would solve the problem that "About AI.JSX" is no longer the top entry in the list.

image

PR #63, opened 2023-06-17
Rework NextJS integration

petersalas:
This layers the NextJS integration (`"ai-jsx/next"`) on top of the React integration (`"ai-jsx/react"`).

The main difference between the two is that `AI.jsx` is a React Server Component that streams results down.

Something (presumably about our monorepo?) prevents Typescript from being able to resolve imports from `"ai-jsx"`, so unfortunately the NextJS files are still all marked `// @ts-nocheck`.


petersalas:
@NickHeiner this works well enough for the one-shot demos but anything more than that would presumably require passing state via the URL or old-school `<input type="hidden" />` or something? Is there some NextJS pattern for that or is that like way out of scope of RSC?

petersalas:
It looks like https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions is maybe what they're building for these kinds of things

NickHeiner:
Yeah, I think that's at least in the same neighborhood.

I think fundamentally, the model is that state from one render to the next is done via form actions. So think of it like you'd think of a traditional server-driven HTML app – a form submits a POST, and the POST data is used to generate the next page. (I think with Next the idea is that it doesn't need to do a full round-trip / refresh the entire page, but it's still the same model.)

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+import * as ReactModule from 'react';
+import * as LLMx from '../react/core.js';
+export * from '../react/core.js';
+import { markAsJsxBoundary } from '../react/jsx-boundary.js';
+
+function unwrapReact(partiallyRendered: LLMx.PartiallyRendered): ReactModule.ReactNode {
+  if (LLMx.isElement(partiallyRendered)) {
+    // This should be an AI.React element.
+    if (partiallyRendered.tag !== LLMx.React) {
+      throw new Error('unwrapReact only expects to see AI.React elements or strings.');
```
NickHeiner, line 11: Can we add one or two more sentences to help devs know how to fix this? And to make it clear whether it's a dev or internal problem? 

Imagine the person seeing this error is brand new to our framework and also their child is asking them for attention at this moment. :smile:

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+function computeSuffix(prefix: ReactModule.ReactNode[], frame: ReactModule.ReactNode[]) {
+  const computedSuffix = [] as ReactModule.ReactNode[];
+  for (let i = 0; i < frame.length; ++i) {
+    if (i >= prefix.length) {
+      computedSuffix.push(frame[i]);
+    } else if (computedSuffix.length > 0) {
+      return null;
+    } else if (prefix[i] === frame[i]) {
```
NickHeiner, line 27: You love an `else` after a `return`. :smile:

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+    return partiallyRendered.props.children;
+  }
+
+  return partiallyRendered;
+}
+
+function computeSuffix(prefix: ReactModule.ReactNode[], frame: ReactModule.ReactNode[]) {
+  const computedSuffix = [] as ReactModule.ReactNode[];
+  for (let i = 0; i < frame.length; ++i) {
+    if (i >= prefix.length) {
```
NickHeiner, line 23: minor: it might be cleaner to concat the entire remainder of the string once `i` exceeds the prefix length: 

```
if (i >= prefix.length) {
    computedSuffix.push(...frame.slice(i))
}
```

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+      computedSuffix.push(frame[i]);
+    } else if (computedSuffix.length > 0) {
+      return null;
+    } else if (prefix[i] === frame[i]) {
+      continue;
+    } else {
+      const fromPrefix = prefix[i];
+      const fromFrame = frame[i];
+
+      if (typeof fromPrefix === 'string' && typeof fromFrame === 'string' && fromFrame.startsWith(fromPrefix)) {
```
NickHeiner, line 33: What's the alternative to them being strings? It's other intermediate values that haven't finished rendering out to strings yet?

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+        return null;
+      }
+    }
+  }
+
+  return computedSuffix;
+}
+
+/**
+ * A JSX component that allows AI.jsx elements to be used in an RSC component tree.
```
NickHeiner, line 45:
```suggestion
 * A JSX component that allows AI.jsx elements to be used in a [NextJS RSC component tree](https://nextjs.org/docs/getting-started/react-essentials#server-components).
```

NickHeiner: "RSC" and "NextJS RSC" may refer to subtly different things, so as long as we're specifically testing against NextJS, I think we should be clear about that in the language.

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  const renderResult = LLMx.createRenderContext().render(children, {
+    stop: (e) => e.tag === LLMx.React,
+    map: (frame) => frame.map(unwrapReact),
+  });
+  const asyncIterator = renderResult[Symbol.asyncIterator]();
+
+  /**
+   * (Ab)uses <Suspense> to stream an AI.JSX response, taking care to only stream deltas if
```
NickHeiner, line 60: > (Ab)uses

💀 💀 💀 

image

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+      children: ReactModule.createElement(() => finalResult as any),
+    });
+
+    return replace(replaceableStream);
+  }
+
+  return ReactModule.createElement(Stream as any, { replace: (node: ReactModule.ReactNode) => node });
+}
+
+markAsJsxBoundary(jsx);
```
NickHeiner, line 114: For fun, can we just apply this as a decorator? (No need to spend more than 2 minutes trying to figure that out.)

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+    stop: (e) => e.tag === LLMx.React,
+    map: (frame) => frame.map(unwrapReact),
+  });
+  const asyncIterator = renderResult[Symbol.asyncIterator]();
+
+  /**
+   * (Ab)uses <Suspense> to stream an AI.JSX response, taking care to only stream deltas if
+   * each frame only appends to the previous one.
+   */
+  async function Stream({
```
NickHeiner, line 63: So we don't need to synchronously return all `Suspense` elements up-front? That was the presumption behind my shitty hack. :smile:

packages/ai-jsx/src/next/index.ts
```diff
@@ -0,0 +1,114 @@
...
+        nextResult.value,
+        ReactModule.createElement(Stream as any, { prefix: nextResult.value, replace: newReplace }),
+      ],
+      children: ReactModule.createElement(() => finalResult as any),
+    });
+
+    return replace(replaceableStream);
+  }
+
+  return ReactModule.createElement(Stream as any, { replace: (node: ReactModule.ReactNode) => node });
```
NickHeiner, line 111: Can you tell me more about what `replace` does? It seems like most of the time, it's actually not a replacement, but an identity function. And the last time it returns an empty value.

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -1,19 +1,19 @@
...
 
 export async function UICompletion(
-  { example, children }: { example: React.ReactNode; children: AI.Node },
-  { render, logger }: AI.ComponentContext
+  { example, children }: { example: React.ReactNode; children: LLMx.Node },
+  { render, logger }: LLMx.ComponentContext
 ) {
   const reactComponents = new Set<React.JSXElementConstructor<any> | string>();
-  function collectComponents(node: React.ReactNode | AI.Node, inReact: boolean) {
+  function collectComponents(node: React.ReactNode | LLMx.Node, inReact: boolean) {
```
NickHeiner, line 16: I thought this was the opposite direction of where we wanted to be moving?

packages/ai-jsx/src/react/jsx-boundary.ts
```diff
@@ -0,0 +1,15 @@
+const isJsxBoundarySymbol = Symbol('AI.jsx boundary');
+
+/**
+ * Indicates that the tag is an AI.jsx tag. (Different flavors of React have different implementations.)
```
NickHeiner, line 4:
```suggestion
 * Indicates that the tag is an AI.jsx tag. (Different flavors of React (e.g. React vs NextJS) have different implementations.)
```

packages/docs/docs/guides/ai-ui.md
```diff
@@ -55,32 +55,29 @@ For this demo, we've set up a hacked version of NextJS to support server-side re
...
        <ChatCompletion temperature={1}>
          <UserMessage>Give me ten facts about {query}</UserMessage>
        </ChatCompletion>
-     </AI>
+     </AI.jsx>
    </ResultContainer>
    ```
-1. If you want to embed React components as a (potentially transitive) child of `<AI>`, you need to update the hacky list in our custom react wrapper (`packages/nextjs-demo/src/app/react.ts`).
+
+1. TODO: Support embedded React components
```
NickHeiner, line 80: I'm fine with this as a TODO for launch if it makes the rest of the path golden.

```suggestion
## Future Work
Support nesting React components within AI.JSX components.
```

packages/nextjs-demo/babel.config.js
```diff
@@ -1,3 +1,13 @@
...
+  plugins: [
+    [
+      '@babel/plugin-transform-react-jsx',
+      {
+        throwIfNamespace: false,
+        runtime: 'automatic',
+        importSource: 'ai-jsx/next',
+      },
+    ],
+  ],
```
NickHeiner, line 12: jfc was this all it took. I tried to do something like this so 💀 🔫  that I couldn't get it working. 😄

packages/nextjs-demo/src/app/nl-gh-search/page.tsx
```diff
@@ -18,17 +19,17 @@ export default function Home(props: { searchParams: any }) {
...
       >
-        <AIComponent>
+        <AI.jsx>
           <NaturalLanguageGitHubSearch query={githubTopic} outputFormat="prose" />
-        </AIComponent>
+        </AI.jsx>
       </ResultContainer>
       <ResultContainer
         title={`Natural Language GitHub Search for: "${githubTopic}"`}
-        description="AI has been asked to output HTML, and we write it directly into the DOM. (If the AI generates something other than HTML, this could produce funky results.)"
```
NickHeiner, line 27: Why did you take this out? :smile:

packages/nextjs-demo/src/app/recipe/page.tsx
```diff
@@ -73,33 +74,36 @@ export default function RecipeWrapper({ searchParams }: { searchParams: any }) {
...
+                  <RecipeInstructionList>
+                    <RecipeInstructionListItem>
+                      Combine the ingredients in a large mixing bowl.
+                    </RecipeInstructionListItem>
+                    <RecipeInstructionListItem>
+                      Beat the contents on high speed until soft peaks form.
+                    </RecipeInstructionListItem>
+                    <RecipeIngredientListItem>Keep chilled until serving.</RecipeIngredientListItem>
+                  </RecipeInstructionList>
+                </Recipe>
```
NickHeiner, line 97: The fact that we can still pass pure React as a prop make me feel much better about the "embedding components is not supported" caveat.

packages/nextjs-demo/src/components/ResultContainer.tsx
```diff
@@ -17,7 +17,7 @@ export default function ResultContainer({
             <h3 className="text-base font-semibold leading-6 text-gray-900">{title}</h3>
             <p>{description}</p>
           </div>
-          <div className="bg-white px-4 py-5 sm:px-6">{children}</div>
+          <div className="bg-white px-4 py-5 sm:px-6 whitespace-pre-wrap">{children}</div>
```
NickHeiner, line 20: Ha! Super clever and simple way to make the demo look dramatically better.

NickHeiner (approved):
This is a huge improvement over what we had! Excited to get this landed.

We should include Next (and eventually CRA) samples in our set of GH Template Repos. This will also be a good test to make sure we haven't accidentally done something that only works in the monorepo.

packages/ai-jsx/src/next/index.ts
petersalas: I looked into this with the React rework, unfortunately AFAICT only classes or class methods are allowed to have decorators (in Typescript). But I might encapsulate this a bit better -- I was on the home-stretch of lighting this up and was in hack-and-slash mode. :)

packages/ai-jsx/src/next/index.ts
petersalas: It's an internal invariant being violated (i.e. an AI.jsx bug), either partial rendering something wrong or the boundary render stopping at element other than `AI.React`.

I'll prefix it to try to make it more clear that this is an AI.jsx problem.

packages/ai-jsx/src/next/index.ts
petersalas: They might be a react component, in which case we are assuming they're distinct. Will clarify

packages/ai-jsx/src/next/index.ts
petersalas: Yeah, you can nest `<Suspense>` elements and Next takes care of making sure that an in-progress async RSC suspends the nearest ancestor. :)

packages/ai-jsx/src/react/completion.tsx
petersalas: It is, but:
- Typescript preserves comments by default (and I don't think we want to change that)
- Babel complains about the JSX pragmas

So this is a way to kick the can until the entire `ai-jsx` codebase is aligned on using `AI.createElement` (so as not to require internal pragmas), or even better, moved over to the new `_jsx()`-style conversion (in which case the leftover pragmas will be meaningless because the JSX is already converted by `tsc`).

packages/ai-jsx/src/next/index.ts
petersalas: Adding some more jsdoc/block comments to elaborate :)

packages/ai-jsx/src/next/index.ts
NickHeiner: That's cool – RSC is even better than I gave it credit for!

Where did you learn about that behavior?

packages/ai-jsx/src/next/index.ts
NickHeiner: Cool, that's fine – I agree that only classes or class methods can have decorators, now that I look it up.

What we have here, or following React's pattern for memo (https://react.dev/reference/react/memo), both seem reasonable to me.

packages/ai-jsx/src/react/completion.tsx
NickHeiner: Ok, if this is a can kick, that's fine with me.

packages/ai-jsx/src/next/index.ts
petersalas: tbf previously not all branches returned. but now they do so eslint wins

packages/docs/docs/guides/ai-ui.md
petersalas: It's actually not even a TODO anymore, I just didn't update the readme

packages/nextjs-demo/src/app/nl-gh-search/page.tsx
petersalas: Oops...because at the time the innerHTML'ing didn't work. But now it does, so back it goes.

packages/nextjs-demo/src/app/recipe/page.tsx
petersalas: It actually is supported now! And I'm using it in the "render-to-html" demo.

packages/nextjs-demo/src/components/ResultContainer.tsx
petersalas: who needs HTML when you have "\n"?

packages/nextjs-demo/babel.config.js
petersalas: To be fair I had to implement the `jsx-runtime` stuff. Maybe it breaks on `"classic"` or smth?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -51,7 +51,10 @@ export const openAiClientContext = LLMx.createContext<OpenAIApi>(
   new OpenAIApi(
     new Configuration({
       apiKey: process.env.OPENAI_API_KEY,
-    })
+    }),
+    undefined,
+    // TODO: Figure out a better way to work around NextJS fetch blocking streaming
+    (globalThis as any)._nextOriginalFetch ?? globalThis.fetch
```
petersalas, line 57: FYI @NickHeiner it looks like NextJS's `fetch` patch does something not-streaming friendly. So from peeking at their code I found out how I could get the "real" fetch 😒

NickHeiner: Hmmm, interesting. Good find.

PR #65, opened 2023-06-17
Disable function mangling in CRA demo

petersalas:
Serializing a React tree into text doesn't work well if function names are mangled. (Namely it prevents the LLM from using inferring component semantics from the name.)


NickHeiner (approved):
Notes from Slack:

> Can we try just reading the component's displayName instead of the function name property? I don't think that will be mangled, and React very well may set it under our configuration. (Or maybe that's only done during dev.)

> If we can't do that, then shy of ejecting to have access to webpack config, we may be able to apply some other babel transform that will set displayName.

petersalas:
`displayName` doesn't get set automatically. The alternative would be to manually set `displayName` on components where it matters and use that in favor of `name` if set, but for now I'm going to stick with disabling mangling.

PR #67, opened 2023-06-17
React CSR DocsQA demo against ai.jsx docs

juberti:
Similar to the basic chat example, but this embeds the /brand-new.html page from Docusaurus as Markdown and does docsqa against it.

Summary of changes to existing code:
- docs.tsx: explicitly sets OpenAI key, allows a limit to be specified for KNN, improves the prompt, and changes the interface to return chunks to `docComponent`.
- tsconfig.json: enables es2017 for top-level await


packages/docs/.gitignore
```diff
@@ -17,6 +17,7 @@ data
 
 # Production
 /build
+/docs
```
mdwelsh, line 20: This is not correct. `docs` needs to be there, `docs/api` should be ignored as the latter is autogenerated.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -313,7 +313,8 @@ export class LangChainEmbeddingWrapper implements Embedding {
 }
 
 /** A default embedding useful for DocsQA. Note that this requires an OPENAI_API_KEY. */
-export const defaultEmbedding = new LangChainEmbeddingWrapper(new OpenAIEmbeddings());
+const openAIEmbeddings = new OpenAIEmbeddings({ openAIApiKey: process.env.OPENAI_API_KEY });
```
NickHeiner, line 316: Ah, this wasn't getting picked up automatically?

If this env var is required to work, we should throw an exception if it's not found.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -542,16 +549,17 @@ export async function DocsQA<Doc extends Document>(props: DocsQAProps<Doc>) {
   if (status !== Corpus.LoadingState.COMPLETED) {
     return `Corpus is not loaded. It's in state: ${status.toString()}`;
   }
-  const docs = await props.corpus.search(props.question);
+  const docs = await props.corpus.search(props.question, { limit: props.limit });
   return (
     <ChatCompletion>
       <SystemMessage>
-        You are a customer service agent. Answer questions truthfully. Here is what you know:
+        You are a trained question answerer. Answer questions truthfully, using only the document excerpts below. Do not
```
NickHeiner, line 556: You can replace some of this with `<Prompt persona='trained question answerer' noInternalKnowledge />`.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -542,16 +549,17 @@ export async function DocsQA<Doc extends Document>(props: DocsQAProps<Doc>) {
...
+        use any other knowledge you have about the world. If you don't know how to answer the question, just say "I
+        don't know." Here are the relevant document excerpts you have been given:
         {docs.map((doc) => (
-          // TODO improve types
-          // @ts-expect-error
           <props.docComponent doc={doc} />
         ))}
+        And here is the question you must answer:
```
NickHeiner, line 562: I think this might be unnecessary in a chat model but 🤷

packages/create-react-app-demo/src/docs-chat/ai.tsx
```diff
@@ -0,0 +1,79 @@
...
+import _ from 'lodash';
+
+export class ChatMessage {
+  type: string;
+  content: string;
+  constructor({ type, content }: { type: string; content: string }) {
+    this.type = type;
+    this.content = content;
+  }
+}
```
NickHeiner, line 21: I don't think this needs to be a full-fledged `class`.

```suggestion
export interface ChatMessage {
  type: string;
  content: string;
}
```

packages/create-react-app-demo/src/docs-chat/ai.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const conversationAtom = atom<ChatMessage[]>([]);
+export const modelCallInProgress = atom<boolean>(false);
+
+// For now, we load the ai.jsx docs from a local markdown file. Once we have a HTML->Markdown converter,
+// we can load the docs from the site directly.
+const docText = await fetch(markdownPath).then((res) => res.text());
```
NickHeiner, line 28:
```suggestion
const docResponse = await fetch(markdownPath)
const docText = await docResponse.text()
```

A little cleaner IMO but nbd

packages/create-react-app-demo/src/docs-chat/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+        type: 'user',
+        content: element.value,
+      },
+    ]);
+
+    element.value = '';
+  }
+
+  return (
+    <ResultContainer title="Docs Chat" description="In this demo, you can ask questions based on a document.">
```
NickHeiner, line 35: It would probably be good to link to the doc in question so the demo is more understandable.

NickHeiner (approved):

packages/ai-jsx/src/batteries/docs.tsx
juberti: yeah, I'm not a fan of that particular pattern, I don't think it offers enough to make up for the additional complexity. I think we probably want to consider some other ways of approaching this sort of "prompt reuse".

packages/ai-jsx/src/batteries/docs.tsx
juberti: yeah, I was looking for a citation here but couldn't find one. I'll file a TODO to check the OAI cookbook on this.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -312,8 +312,11 @@ export class LangChainEmbeddingWrapper implements Embedding {
   }
 }
 
-/** A default embedding useful for DocsQA. Note that this requires an OPENAI_API_KEY. */
-export const defaultEmbedding = new LangChainEmbeddingWrapper(new OpenAIEmbeddings());
+/** A default embedding useful for DocsQA. Note that this requires OPENAI_API_KEY to be set. */
+if (!process.env.OPENAI_API_KEY) throw new Error('OPENAI_API_KEY not set');
```
mdepinet, line 316: Does this work? Since the const object is what's exported, I won't expect execution of this line to be related to importing the const value.

packages/create-react-app-demo/src/docs-chat/ai.tsx
mdepinet: nit: class -> interface

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
```diff
@@ -35,11 +35,11 @@ const docs = await Promise.all(
 const corpus = new LocalCorpus(staticLoader(docs), defaultChunker);
 await corpus.startLoading();
 
-function ShowDoc({ doc }: { doc: Document<RedditMetadata> }) {
+function ShowDoc({ doc }: { doc: ScoredChunk }) {
```
mdepinet, line 38: I think it's somewhat important to show that the metadata type still lines up.

```suggestion
function ShowDoc({ doc }: { doc: ScoredChunk<RedditMetadata> }) {
```

I think that means you could continue using the title from the metadata as a first choice.

mdepinet (approved):

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
juberti: I wasn't able to get this to work properly with TS checks, it complained that `RedditMetadata` was not `Jsonifiable`. Do you want to take another run at this?

packages/ai-jsx/src/batteries/docs.tsx
juberti: I see what you mean. I think in practice it shouldn't matter as process.env should be immutable though. Or did you have another idea?

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
mdepinet: Hmm... I had initially defined RedditMetadata as an interface that extends JsonObject to avoid exactly that problem. The "MustBeJsonObject" alternative that exists now was Nick's suggestion.  @NickHeiner Can you help with this?

(I also think it's fine to submit as is and I can clean it up later.)

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: I was thinking you could do something like what Matt wrote and then include this check inside the makeChunker function, but your point about process.env being immutable makes sense too.

juberti: OK. I was going to add a `MarkdownSplitter` too in a followup so I can take another look when I do that.

packages/examples/src/bakeoff/reddit-qa/ai-jsx.tsx
juberti: OK, SGTM.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: This will work, because the entire file is executed whenever anything is `import`ed from it.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: Legit. Looking forward to seeing how things develop and whether the pattern ultimately feels useful.

PR #69, opened 2023-06-17
React DocsQA: retrieve ai.jsx docs dynamically

juberti:
Get rid of the old crufty loader and fetch the docs like an actual human being.


packages/create-react-app-demo/src/docs-chat/ai.tsx
```diff
@@ -19,18 +18,27 @@ export interface ChatMessage {
...
-  {
-    pageContent: [docText],
-    name: 'Guide for AI Newcomers',
-  },
-];
-const corpus = new LocalCorpus(staticLoader(docs), defaultChunker);
-await corpus.startLoading();
+/*
+ * This is a very simple example of how to build a document corpus.
+ * We pull the page content from a URL, convert it to Markdown, and then index it.
```
NickHeiner, line 23: This is fine for the example, but if we're converting to md, should we just request the raw file from GH directly? https://raw.githubusercontent.com/fixie-ai/ai-jsx/main/packages/docs/docs/guides/brand-new.md

NickHeiner (approved):

NickHeiner:
We should delete `packages/create-react-app-demo/src/docs-chat/brand-new.md` as part of this PR.

packages/create-react-app-demo/src/docs-chat/ai.tsx
```diff
@@ -1,32 +1,50 @@
...
+    'getting-started.md': 'Getting Started',
+    'guides/ai-ui.md': 'AI + UI',
+    'guides/brand-new.md': 'Guide for AI Newcomers',
+    'guides/docsqa.md': 'Docs QA: Grounding Answers',
+    'guides/esm.md': 'ESM',
+    'guides/jsx.md': 'JSX: Build System Considerations',
+    'guides/observability.md': 'Observability',
+    'guides/prompting.md': 'Getting the AI to say things',
+    'guides/rules-of-jsx.md': 'Rules of AI.JSX',
+  };
```
NickHeiner, line 23: I feel much better about this demo now that we're reading all the files. :smile:

packages/create-react-app-demo/src/docs-chat/ai.tsx
```diff
@@ -1,32 +1,50 @@
...
 
+/**
+ * Build the corpus on the first query, then use it for all subsequent queries.
+ */
 export async function DocsAgent({ question }: { question: string }) {
-  await corpusLoadedPromise;
-  return <DocsQA question={question} corpus={corpus} limit={5} docComponent={ShowDoc} />;
+  if (!globalCorpus) {
+    globalCorpus = await indexCorpus();
+  }
```
NickHeiner, line 48: Rather than this approach, I'd do:

```jsx
const indexCorpus = _.once(async function indexCorpus() {
  /* ... */
});

<DocsQA question={question} corpus={indexCorpus()} limit={5} docComponent={DefaultFormatter} />;
```

`_.once` creates a function where the wrapper is only invoked once; subsequent calls return the value returned by the first invocation.

PR #70, opened 2023-06-18
Stream more in create-react-app

NickHeiner:
This expands the UI bridge affordance, then updates the Docs Chat example to use it.

As a result, we now have streaming results in Docs Chat. Docs Chat is also vastly simplified.

This will cause some merge conflicts with #69 but I think we mostly changed different parts so it shouldn't be too bad to resolve.

As a follow-up, it would be good to do this for the "Basic Chat" example as well.


packages/ai-jsx/src/react/index.ts
```diff
@@ -92,14 +95,24 @@ export function useAI(children: LLMx.Node) {
...
-export function jsx({ children }: { children: LLMx.Node }, context?: any | LLMx.ComponentContext) {
+export function jsx(
+  {
+    children,
+    onStreamStart,
+    onStreamEnd,
+    loading = '',
+  }: { children: LLMx.Node; onStreamStart?: () => void; onStreamEnd?: () => void; loading?: React.ReactNode },
+  context?: any | LLMx.ComponentContext
+) {
```
NickHeiner, line 106: @petersalas I think it'll be useful for UIs to know when a generation is in progress. Lmk what you think / if you have an alternate suggestion for how to do it.

The `loading` function is the moral equivalent of `Suspense'`s `fallback`. I chose not to use that name because `fallback` to me implies `failed`, as opposed to `not-started`.

packages/create-react-app-demo/src/docs-chat/ai.tsx
```diff
@@ -30,7 +21,8 @@ const docs = [
   },
 ];
 const corpus = new LocalCorpus(staticLoader(docs), defaultChunker);
-await corpus.startLoading();
```
NickHeiner, line 33: @juberti by `await`ing this at the top level, we blocked the entire UI from showing until it was complete. 

The better thing to do is to start the loading ASAP, but only block on it when we actually need to use the results.

packages/ai-jsx/src/react/index.ts
petersalas: I'm wondering if a) useAI should take the value as an "initialValue" and b) components that want to observe the lifecycle should use useAI directly. On first glance that feels more consistent with how other hooks/async operations are modeled in React. WDYT?

NickHeiner: I don't have a strong opinion now but I expect to develop one as we build more apps.

I think some usecases will definitely want to access `useAI` directly, and some (most?) of those usecases will want to observe the lifecycle.

But I also think a purely declarative approach (like what I did in this PR) can righteously benefit from observing the lifecycle as well.

But maybe I'll change my mind on all this later.

re: initial value – I don't think I'm fully following.

PR #71, opened 2023-06-18
Arch and Perf Docs

NickHeiner:


mdwelsh (approved):

packages/docs/docs/guides/architecture.md
```diff
@@ -0,0 +1,168 @@
...
+With a pure client-side approach, you run your UI and AI.JSX logic on the client. However, you still need to call out to external APIs for hosted services like model providers (e.g. GPT-4 from OpenAI) or databases (e.g. Pinecone).
+
+Pros:
+
+- Simple to implement
+- The serialization boundary is at the API layer. The only way to access these APIs is through a serialization boundary, so your architecture isn't adding any new boundaries.
+
+Cons:
+
+- Your API keys are exposed to the client.
```
juberti, line 42: probably add a bit more guidance here, e.g,. this approach is not recommended for public applications

packages/docs/docs/guides/architecture.md
```diff
@@ -0,0 +1,168 @@
...
+
+This is just like the [Run Entirely on the Client](#run-entirely-on-the-client) pattern, except you add a proxy for API calls. You'd traditionally set this proxy up as a serverless or edge function. The proxy stores your API keys, keeping them safe from the client.
+
+Pros:
+
+- There's still only one serialization boundary.
+- API keys are protected from the client.
+
+Cons:
+
```
juberti, line 68: same roundtrip issues as above

packages/docs/docs/guides/architecture.md
```diff
@@ -0,0 +1,168 @@
...
+Pros:
+
+- You can access your AI.JSX logic from multiple clients (web app, native app, etc).
+- API keys are secure.
+- Performance will be less sensitive to roundtrips between AI.JSX and external APIs.
+
+Cons:
+
+- This introduces a new serialization boundary, which sits between the UI and your AI.JSX. This limits how seamless your UI/AI integration can be.
```
juberti, line 99: Maybe flesh this out a bit more? It's kinda abstract rn

packages/docs/docs/guides/architecture.md
```diff
@@ -0,0 +1,168 @@
...
+In this approach, you generate your HTML on the server and stream it to the client. If your app is full-stack JS, you'd typically do this with the help of a server-side rendering framework like [NextJS](https://nextjs.org/).
+
+Pros:
+
+- [Server-side rendering is often a performance win.](https://nextjs.org/docs/getting-started/react-essentials)
+- API keys are secure.
+
+Cons:
+
+- In JS, requires use of a framework like [NextJS](https://nextjs.org/) to get the full benefits. (NextJS is a great choice for new apps, but exisiting codebases may not be able to adopt it immediately.)
```
juberti, line 128:
```suggestion
- In JS, requires use of a framework like [NextJS](https://nextjs.org/) to get the full benefits. (NextJS is a great choice for new apps, but existing codebases may not be able to adopt it immediately.)
```

packages/docs/docs/guides/architecture.md
```diff
@@ -0,0 +1,168 @@
...
+    AI.JSX->>APIs: Model calls (e.g. to OpenAI)
+    AI.JSX->>APIs: Other API calls (e.g. to Pinecone)
+    box rgb(200, 255, 200) Third-Party API
+    participant APIs
+    end
+```
+
+## UI on the client; AI.JSX on the server
+
+In this approach, your AI.JSX logic runs entirely in your backend environment.
```
juberti, line 89: does next.js/RSC enter into the equation here?

packages/docs/docs/guides/performance.md
```diff
@@ -0,0 +1,177 @@
...
+  return (
+    <ChatCompletion>
+      <SystemMessage>Respond in JSON.</SystemMessage>
+      <UserMessage>Write me a profile for a fantasy character</UserMessage>
+    </ChatCompletion>
+  );
+}
+```
+
+If you can get the model to do what you want in a single shot, that'll be more performant. However, asking the model to do more at once decreases reliability. It's more split your workload into simple, discrete tasks for the model. So, there are tradeoffs to balance here. You want the task size to be as complicated as the model can reliably do in a single pass, but no more complicated.
```
juberti, line 107: The "It's more" sentence seems like it's missing something.

packages/docs/docs/guides/performance.md
```diff
@@ -0,0 +1,177 @@
...
+
+With this approach, you're still deferring execution optimally, but you also ensure that each instance of `heroName` will resolve to the same generated result.
+
+## Use a faster model
+
+Different models have different performance profiles. GPT-4 is slower than GPT-3.5-Turbo, for instance. But, unfortunately, the slower models tend to be more correct. So you'll have to find the tradeoff that works for you.
+
+OpenAI's recommendation is to start with the most powerful model, get your app working, then move to faster models if it's possible to do so without sacrificing correctness.
+
+You may also want to consider different model providers; OpenAI vs. [OpenAI-on-Azure](https://azure.microsoft.com/en-us/products/cognitive-services/openai-service) may have different performance and uptime profiles. AI.JSX makes it easy to switch model providers for a part or whole of your app via [context](./rules-of-jsx.md#context).
```
juberti, line 177: probably also want to talk about non-openai models here

packages/docs/docs/guides/performance.md
```diff
@@ -0,0 +1,177 @@
...
+const finalResult = await AI.createRenderContext().render(<MyComponent />, {
+  map: handleIntermediateResult,
+});
+```
+
+For more detail, see [Intermediate Results](./rules-of-jsx.md#intermediate-results).
+
+## Minimize how long the output needs to be
+
+A model generation takes linearly more time as the output length increases, so the shorter your output can be, the faster it'll be. However, this inherently means the model is spending less time "thinking" about your result, [which could degrade accuracy](./brand-new.md#thinking-out-loud). You'll need to balance these tradeoffs.
```
juberti, line 64: maybe also talk about `max_tokens` here

juberti (approved):
lg overall, added a bunch of notes on things that could be discussed further

packages/docs/docs/guides/performance.md
NickHeiner: whoops. Good catch.

packages/docs/docs/guides/architecture.md
NickHeiner: I pushed a change to this doc and others that hopefully addresses this.

PR #68, opened 2023-06-17
Add some simple, step-by-step tutorial examples.

mdwelsh:
This is a first cut at some 'step by step' tutorial examples which I will write docs around and comment extensively in a follow-up PR.

This PR also fixes some typing issues with DocsQA which Justin's PR also attempts to address. I also fix the bakeoff DocsQA example's typing.

PR #72, opened 2023-06-18
Add documentation to go along with tutorials.

mdwelsh:
This PR adds tutorial documentation to go along with the tutorial examples from #68.

I also made the changes to the DocsQA props as discussed here:

https://fixie-ai.slack.com/archives/C05B6EE9JE5/p1687116248841759


packages/docs/docs/tutorial/part2.md
```diff
@@ -0,0 +1,64 @@
...
+import { ChatCompletion, UserMessage } from 'ai-jsx/core/completion';
+import { showInspector } from 'ai-jsx/core/inspector';
+
+const app = (
+  <Inline>
+    <ChatCompletion>
+      <UserMessage>Come up with the name of a mythical forest animal.</UserMessage>
+    </ChatCompletion>
+    {'\n\n'}
+    {(conversation) => (
```
zkoch, line 24: I know you explain it in the text below, but I think adding a comment above here saying that `conversation` is the previous output of the LLM helps explain the code. Otherwise it's not clear where this is coming from

packages/docs/docs/tutorial/part2.md
```diff
@@ -0,0 +1,64 @@
...
+      <ChatCompletion>
+        <UserMessage>Imagine a mythical forest animal called the "{conversation}". Tell me more about it.</UserMessage>
```
zkoch, line 26: Nit: "...called a"?

packages/docs/docs/tutorial/part2.md
```diff
@@ -0,0 +1,64 @@
...
+The third child is a _function_ that takes a single argument -- `conversation` -- and returns
+a `<ChatCompletion>` component that invokes the LLM again. The `conversation` argument contains
+the state of the rendered output of the `<Inline>` component up to this point.
+In this way, you can build up a response consisting of multiple LLM invocations.
+The fourth child of `<Inline>` is another string containing two newline characters.
+Finally, the fifth child of `<Inline>` is another function that takes in the current state
+of the rendered output, and returns a `<ChatCompletion>` component that invokes the LLM
+a third time.
+
+## The Inspector
```
zkoch, line 58: Inspector supports step-by-step debugging, right? Worth talking about that here?

packages/docs/docs/tutorial/part3.md
```diff
@@ -0,0 +1,125 @@
...
+    {(conversation) => (
+      <YamlChatCompletion>
+        <UserMessage>{conversation}</UserMessage>
+      </YamlChatCompletion>
+    )}
+  </Inline>
+);
+```
+
+For example, the `<JsonChatCompletion>` component will emit:
```
zkoch, line 104: Does the `JsonChatCompletion` component allow you to specify the shape of the final JSON string?

packages/docs/docs/tutorial/part4.md
```diff
@@ -0,0 +1,110 @@
...
+
+The implementation of `<DocsQA>` in AI.JSX currently stores the chunks and embedding
+vectors in memory. One could also use a vector database to store the embedding vectors, and
+in the near future we'll be adding support for this to AI.JSX.
+
+The `<DocsQA>` component takes in several props:
+
+- `question` - the query to submit to the LLM about this document corpus.
+- `corpus` - the corpus of documents to search for the answer.
+- `chunkLimit` - the maximum number of document chunks to consider in formulating the response.
```
zkoch, line 43: Are both `chunkLimit` and `chunkFormatter` required? Or is there a default implementation?

zkoch (approved):
These look good!

If I were a greedy man, I would say it'd be amazing to add two additional features:
1.) Example of `<UseTools>`, since I think this is really common.
2.) Use of the natural language router

I feel like you could do something where you're playing a basic fantasy role playing game where there are a variety of tools available (e.g., roll dice, etc)

packages/docs/docs/tutorial/part2.md
mdwelsh: I prefer to show the code first and explain it second - more consistent that way.

packages/docs/docs/tutorial/part3.md
mdwelsh: I don't think so.

packages/docs/docs/tutorial/part4.md
mdwelsh: chunkLimit is optional. There's no default implementation (there probably should be).

mdwelsh:
> 
Sure, let's add some in follow up PRs.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -527,29 +527,33 @@ export interface DocsQAProps {
...
+   * The question to answer.
    */
   question: string;
 
   /**
    *
-   * The maximum number of documents to return.
+   * When processing a DocsQA query, the most relevent chunks are presented to the model. This
+   * field limits the number of chunks to consider. If this value is too large, the size
+   * of the chunks may exceed the token limit of the model.
```
NickHeiner, line 538: I think this and all our docs need to walk a tightrope of "explain concisely to someone who already knows what a `chunk` is" and "provide an on-ramp to people brand new to LLMs who don't." For the latter group, we could link to https://docs.ai-jsx.com/guides/docsqa. (Or maybe we need a glossary.)

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -527,29 +527,33 @@ export interface DocsQAProps {
...
+  chunkLimit?: number;
 
   /**
    * The component used to format document chunks when they're presented to the model.
+   * This can be used to present the model with some metadata about the chunk, in a format
+   * that is appropriate to the type of document.
    *
    * ```tsx
-   *  function MyDocsComponent({ doc }: { doc: ScoredChunk }) {
+   *  function FormatChunk({ chunk }: { chunk: ScoredChunk }) {
```
NickHeiner, line 548: Minor, but typically we'd name a component declaratively, not imperatively. Like, `Timestamp`, rather than `ShowTime`. `UserAvatar` rather than `ShowUserImage`. etc.

So maybe here it's `ChunkTemplate`? Or `ChunkView`?

packages/docs/docs/about.mdx
```diff
@@ -1,39 +1,67 @@
...
+## Getting started
+
+To get started, follow the [Getting Started Guide](./getting-started), or check out the
+[examples](https://github.com/fixie-ai/ai-jsx/tree/main/packages/examples).
+
+AI.JSX can be used to create standalone LLM applications that can
+be deployed anywhere Node.JS is supported, or it can be used as
+part of a larger React application. We have examples that use
+[Next.js](https://github.com/fixie-ai/ai-jsx/tree/main/packages/nextjs-demo)
+and [Create React
```
NickHeiner, line 42: For launch, we're just focusing on Next.js as the supported avenue. CRA will follow shortly thereafter.

I intend to do another pass on the ai-ui doc, and I intend to expand in more detail about CRA being experimental for now.

packages/docs/docs/about.mdx
```diff
@@ -64,26 +92,3 @@ function CharacterGenerator() {
 ```
 
 [![AI.JSX demo](https://asciinema.org/a/m08ORXH8AdcmV8gA95d6UYJzW.svg)](https://asciinema.org/a/m08ORXH8AdcmV8gA95d6UYJzW)
```
NickHeiner, line 94: This looks great – if we could embed the video player directly in the page, that would be even better. (But this doesn't have to be a huge priority.)

packages/docs/docs/getting-started.md
```diff
@@ -2,223 +2,62 @@
...
-> Eldra, Jalara, and Aria found themselves at a crossroads in their lives, each searching for something more. They had all independently heard rumors of a mystical land known as the Cavern of Secrets, a place said to hold untold treasures and knowledge beyond their wildest dreams.
->
-> Eager to prove themselves, the trio decided to set aside their differences and embark on the journey together. Eldra was searching for a new challenge, Jalara hoped to uncover lost magical artifacts, and Aria sought to expand her knowledge and prophetic abilities.
->
-> Together, they journeyed through treacherous terrain and battled against fierce creatures, their skills complementing each other perfectly. Eldra's pugilist abilities kept their enemies at bay, Jalara's sorcery helped them navigate unfamiliar terrain, and Aria's warding spells protected them from harm.
->
-> (Clipped for brevity)
-
-To see the previous characters that were generated, we can look in the logs:
+The following commands should build the entire AI.JSX repo:
```
NickHeiner, line 39: Let's say `build` instead of `should build` to express self-confidence. :smile:

packages/docs/docs/getting-started.md
```diff
@@ -2,223 +2,62 @@
...
-`<ImageGen>` by default uses [Dalle](https://platform.openai.com/docs/guides/images/introduction).
+You can then run the various examples from the top level of the `ai-jsx` checkout using
+`yarn workspace`, like so:
 
-Note that we also used a new component, `memo`. If we simply use `<WriteStory />` twice (even if you assign it to a variable), you will have two separate LLM calls and possibly two different stories as a result.
-For reference see [Memoization](guides/rules-of-jsx.md#Memoization).
+```
+$ yarn workspace tutorial run part1
+$ yarn workspace examples run demo:debate
+$ yarn workspace create-react-app-demo start
```
NickHeiner, line 57: The preferred command, as noted in [the AI-UI page](https://docs.ai-jsx.com/guides/ai-ui), is:

```
yarn turbo run dev --scope create-react-app-demo
```

packages/docs/docs/tutorial/part1.md
```diff
@@ -0,0 +1,63 @@
...
+
+```
+yarn workspace tutorial run part1
+```
+
+## How it works
+
+The first part of the code defines the variable `app`, which
+consists of a JSX component tree that will be rendered by the
+AI.JSX runtime. This is very similar to React, which uses
```
NickHeiner, line 41: You may want to link here to https://docs.ai-jsx.com/is-it-react, which clarifies the similarities and differences from React.

packages/docs/docs/tutorial/part1.md
```diff
@@ -0,0 +1,63 @@
...
+JSX components to generate a DOM tree. Here, we are using JSX
+to describe the structure of the LLM model invocations and prompts
+in a compact, declarative way.
+
+The `<ChatCompletion>` component is the root of the tree, and is responsible
+for generating a single invocation of the LLM -- in this case, OpenAI's
```
NickHeiner, line 47: Uber-nit: let's use an em-dash (–) rather than two hyphens (--). You can produce one on mac by hitting `alt`+`-`.

packages/docs/docs/tutorial/part1.md
```diff
@@ -0,0 +1,63 @@
...
+    <UserMessage>Write a Shakespearean sonnet about AI models.</UserMessage>
+  </ChatCompletion>
+);
+
+const renderContext = LLMx.createRenderContext();
+const response = await renderContext.render(app);
+console.log(response);
+```
+
+You can run this yourself from the AI.JSX tree by running:
```
NickHeiner, line 31: This might be nice as an [admonition](https://docusaurus.io/docs/markdown-features/admonitions).

packages/docs/docs/tutorial/part1.md
```diff
@@ -0,0 +1,63 @@
...
+JSX components to generate a DOM tree. Here, we are using JSX
+to describe the structure of the LLM model invocations and prompts
+in a compact, declarative way.
+
+The `<ChatCompletion>` component is the root of the tree, and is responsible
+for generating a single invocation of the LLM -- in this case, OpenAI's
+ChatGPT model -- and returning the result. The `<UserMessage>` component
+simply tags the prompt as being the "user" component of the prompt -- as opposed
+to `<SystemMessage>`, which would be the "system" component of the prompt. These
+distinctions are used by OpenAI's API to differentiate between the fixed, system aspect
```
NickHeiner, line 51: Let's link to the OpenAI API docs that explain system/user messages here.

One question we might get at this point, which might be good to explain in an admonition: "does this mean AI.JSX only supports OpenAI?"

packages/docs/docs/tutorial/part1.md
```diff
@@ -0,0 +1,63 @@
...
+of the prompt and the variable, user aspect of the prompt.
+
+In order to actually get a result, we need to pass our application to a
+`RenderContext` and call `render()` on it. The `RenderContext` is responsible
+for managing the state of the application and progressively evaluating the state
```
NickHeiner, line 56: For "progressively evaluating", we may wish to link to https://docs.ai-jsx.com/guides/rules-of-jsx#intermediate-results.

NickHeiner:
This is great content. Maybe we want to put `Tutorials` as a top-level sibling of `Docs`?

image

packages/docs/docs/tutorial/part2.md
```diff
@@ -0,0 +1,66 @@
...
+import { ChatCompletion, UserMessage } from 'ai-jsx/core/completion';
+import { showInspector } from 'ai-jsx/core/inspector';
+
+const app = (
+  <Inline>
+    <ChatCompletion>
+      <UserMessage>Come up with the name of a mythical forest animal.</UserMessage>
+    </ChatCompletion>
+    {'\n\n'}
+    {(conversation) => (
```
NickHeiner, line 24: Maybe we should rename `conversation` to `mythicalAnimalName`, and the second `conversation` to `mythicalAnimalNameAndDescription`. That could make it clearer what the content will be.

packages/docs/docs/tutorial/part3.md
```diff
@@ -0,0 +1,125 @@
...
+In the following example, we're going to have the LLM generate a character sheet for a
+fantasty role-playing game, and then ensure that the output is formatted as either
+JSON or YAML.
+
+## Generating a plain-text character sheet
+
+Let's start with a component called `<CharacterGenerator>` that will generate a list of
+fields
+
+```tsx filename="packages/tutorial/src/part3.tsx"
```
NickHeiner, line 20: For some reason, the `filename` doesn't consistently show up. I can't tell if we're doing it wrong or Docusaurus has a bug.

image

packages/docs/docs/tutorial/part3.md
```diff
@@ -0,0 +1,125 @@
...
+function CharacterGenerator() {
+  return (
+    <>
+      The following is a character profile for an RPG game in JSON format:{'\n'}
+      <CharacterField fieldName="name" />
```
NickHeiner, line 25: Should we define the `CharacterField1` component?

packages/docs/docs/tutorial/part3.md
```diff
@@ -0,0 +1,125 @@
...
+## Generating a plain-text character sheet
+
+Let's start with a component called `<CharacterGenerator>` that will generate a list of
+fields
+
+```tsx filename="packages/tutorial/src/part3.tsx"
+function CharacterGenerator() {
+  return (
+    <>
+      The following is a character profile for an RPG game in JSON format:{'\n'}
```
NickHeiner, line 24: Should we say "plain text format" instead of "JSON format"?

packages/docs/docs/tutorial/part3.md
```diff
@@ -0,0 +1,125 @@
...
+achieve this.
+
+In the following example, we're going to have the LLM generate a character sheet for a
+fantasty role-playing game, and then ensure that the output is formatted as either
+JSON or YAML.
+
+## Generating a plain-text character sheet
+
+Let's start with a component called `<CharacterGenerator>` that will generate a list of
+fields
```
NickHeiner, line 18: This sentence should end in punctuation.

packages/docs/docs/tutorial/part3.md
NickHeiner: @farzadab looked at that but couldn't find a DX that felt good. I think this is something we can revisit with OpenAI Functions.

packages/docs/docs/tutorial/part3.md
```diff
@@ -0,0 +1,125 @@
...
+    )}
+    {'\n\n'}
+    And here is a YAML representation of this character profile:{'\n'}
+    {(conversation) => (
+      <YamlChatCompletion>
+        <UserMessage>{conversation}</UserMessage>
+      </YamlChatCompletion>
+    )}
+  </Inline>
+);
```
NickHeiner, line 101: I think I made a similar point elsewhere, but I wouldn't do this with `Inline`. I think it's far more natural to do something like:

```tsx
const character = memo(<CharacterGenerator />)
const app = (
  The following is a JSON representation of this character profile:{'\n'}
  <JsonChatCompletion>
    <UserMessage>{character}</UserMessage>
  </JsonChatCompletion>
  < YamlChatCompletion >
    <UserMessage>{character}</UserMessage>
  </YamlChatCompletion >
)
```

Using `Inline` is also a little funny because we're prompting the `Yaml` component with both the plaintext and JSON output, which seems unnecessary.

packages/docs/docs/tutorial/part3.md
```diff
@@ -0,0 +1,125 @@
...
+    "spells": [
+        "Firebolt",
+        "Mage Armor"
+    ]
+}
+```
+
+Internally, the `<JsonChatCompletion>` and `<YamlChatCompletion>` components prompt the
+LLM to take the input data, format it as JSON or YAML, and check that the resulting objects
+correctly parse in the target format.
```
NickHeiner, line 125: This actually isn't what the component was originally designed to do, and given the prompts that we're providing in [the component](https://github.com/fixie-ai/ai-jsx/blob/d7d91afdbd0b3bc5a20af367b82d6e9f2277d0d9/packages/ai-jsx/src/batteries/constrained-output.tsx), it feels incidental if it does happen to work now. :smile:

That said, the component you're describing here would be a great thing to have.

packages/docs/docs/tutorial/part4.md
```diff
@@ -0,0 +1,110 @@
+---
+sidebar_position: 4
+---
+
+# Tutorial Part 4 - Document Q&A
```
NickHeiner, line 5: All the tutorial elements have a great progression.

packages/docs/docs/tutorial/part4.md
```diff
@@ -0,0 +1,110 @@
...
+
+One of the most powerful capabilities of Large Language Models is the ability to
+answer questions about the contents of a set of documents. AI. JSX provides a powerful
```
NickHeiner, line 8: Extra space in `AI. JSX`.

packages/docs/docs/tutorial/part4.md
```diff
@@ -0,0 +1,110 @@
...
+component called `<DocsQA>` that can answer a question about a corpus of documents
+using the LLM.
+
+Here's an example of how this will look in your app:
+
+```tsx filename="packages/tutorial/src/part4.tsx"
+<DocsQA question="When did the storm occur?" corpus={corpus} chunkLimit={5} chunkFormatter={GetChunk} />
+```
+
+## How does Documents Q&A work?
```
NickHeiner, line 18: I feel like there's a fair bit of overlap here with https://docs.ai-jsx.com/guides/docsqa – can we consolidate?

packages/docs/docs/tutorial/part4.md
```diff
@@ -0,0 +1,110 @@
...
+## Creating a corpus
+
+The first step is to create a corpus that you want the LLM to answer questions about.
+In this example, we'll create a corpus that consists of a single
+Wikipedia article, one about [Hurricane Katrina](https://en.wikipedia.org/wiki/Hurricane_Katrina).
+
+```tsx filename="packages/tutorial/src/part4.tsx"
+// Fetch the HTML for the Wikipedia article.
+const URL = 'https://en.wikipedia.org/wiki/Hurricane_Katrina';
+const html = await fetch(URL).then((response) => response.text());
```
NickHeiner, line 55: minor, but I would do this as:

```ts
const wikiResponse = await fetch(URL)
const html = await wikiResponse.text()
```

packages/docs/docs/tutorial/part4.md
```diff
@@ -0,0 +1,110 @@
...
+
+// Convert the HTML to Markdown using the Turndown library.
```
NickHeiner, line 57: "Why do we need to do this? Why can't we feed in HTML directly?"

PR #74, opened 2023-06-19
Docs update (langchain mostly)

zkoch:


packages/docs/docs/is-it-react.md
```diff
@@ -1,6 +1,6 @@
-# Is This React?
+# Differences with React
```
mdwelsh, line 1: Grammar nit: "Differences from React"

mdwelsh (approved):

packages/ai-jsx/readme.md
```diff
@@ -1,6 +1,127 @@
-# AI.JSX
```
NickHeiner, line 1: Is this just a clone of the root-level README? If so, let's just use a symlink rather than copy/paste.

NickHeiner (approved):

PR #76, opened 2023-06-19
Downgrade to yarn v3

petersalas:
Turbo ostensibly can't parse Yarn v4 lockfiles (and it makes me wonder if this is responsible for incorrect build caching I occasionally see). This downgrades to Yarn v3.


mdwelsh (approved):
Ah this explains the errors I've been seeing from turbo...

PR #78, opened 2023-06-19
Switch everything over to ai-jsx/jsx-runtime

petersalas:
The "new" JSX style has several benefits:
- No need to do an `import * as AI from "ai-jsx"`
- Pragma is a much nicer one-line instead of multiple lines: `/** @jsxImportSource ai-jsx/react */`
- It's compatible with the default NextJS template without overriding Babel settings


packages/ai-jsx/tsconfig-base.json
```diff
@@ -2,9 +2,12 @@
...
   "compilerOptions": {
-    "jsx": "react",
-    "jsxFactory": "AI.createElement",
-    "jsxFragmentFactory": "AI.Fragment",
+    "jsx": "react-jsx",
+    "jsxImportSource": "ai-jsx",
+    "paths": {
+      "ai-jsx/jsx-runtime": ["./src/jsx-runtime.ts"],
+      "ai-jsx/react/jsx-runtime": ["./src/react/jsx-runtime.ts"]
+    },
```
NickHeiner, line 10: Can you update the docs where we talk about our tsconfig settings generally?

NickHeiner (approved):
This is a big win. Thanks for doing it.

PR #81, opened 2023-06-19
next.js: Docs Chat

juberti:
WIP for convo storage but still kinda works


packages/nextjs-demo/src/app/docs-chat/page.tsx
```diff
@@ -0,0 +1,84 @@
...
+      {emoji}: {responseContent}
+    </div>
+  );
+}
+
+/**
+ * We need to memoize this function. Otherwise, every time the parent component of AgentResponse re-renders,
+ * AgentReponse will re-render, which will trigger a new LLM call, which will produce a different result than the
+ * prior call.
+ */
```
NickHeiner, line 27: I think this comment likely does not apply in NextJS.

packages/nextjs-demo/src/app/docs-chat/page.tsx
```diff
@@ -0,0 +1,84 @@
...
+        </AI.jsx>
+      </Suspense>
+    </ConversationItem>
+  );
+});
+
+export default function DocsChat({ searchParams }: { searchParams: any }) {
+  const defaultValue = 'What is AI.JSX?';
+  const query = searchParams.message ?? defaultValue;
+  const userMessages: string[] = [query];
```
NickHeiner, line 43:
```suggestion
  const userMessages = [query];
```

NickHeiner: I don't think this type annotation is necessary. We want to let type inference do as much as it can.

NickHeiner (approved):

petersalas (approved):

PR #83, opened 2023-06-19
Clean up some Corpus rough edges and add easy LangChain support.

mdepinet:


packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -370,15 +400,19 @@ export interface Corpus<ChunkMetadata extends Jsonifiable = Jsonifiable> {
   startLoading: () => Promise<Corpus.Stats>;
```
juberti, line 400: Can we get rid of this and just `await corpus.waitForLoadingToComplete` (or maybe `await corpus.createIndex()`)

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -325,6 +326,30 @@ export const defaultEmbedding = new LangChainEmbeddingWrapper(
   new OpenAIEmbeddings({ openAIApiKey: process.env.OPENAI_API_KEY })
 );
 
+/** A function converting documents into EmbeddedChunks. Also see `asVectorizer`. */
```
NickHeiner, line 329: You can use a `@see` TSDoc tag to make it a bit more structured.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -325,6 +326,30 @@ export const defaultEmbedding = new LangChainEmbeddingWrapper(
...
+  ChunkMetadata extends Jsonifiable = Jsonifiable
+>(chunker: Chunker<DocumentMetadata, ChunkMetadata>, embedding: Embedding) {
+  return async (docs: Document<DocumentMetadata>[]) => {
+    const chunks: Chunk<ChunkMetadata>[] = [];
+    await Promise.all(
+      docs.map(async (doc) => {
+        const docChunks = await chunker(doc);
+        chunks.push(...docChunks);
+      })
+    );
```
NickHeiner, line 347:
```suggestion
    const chunks = _.flat(await Promise.all(
      docs.map(async (doc) => chunker(doc))
    ));
```

I think this will work.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: WDYT of this?

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: What is the underscore here? Is that something I need to import?  (This didn't seem to work without any new imports fwiw.)

NickHeiner: Yep! It's not a special symbol although I see how it could look that way. :smile:

```tsx
import _ from 'lodash`
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -351,166 +380,213 @@ export interface EmbeddedChunk<ChunkMetadata extends Jsonifiable = Jsonifiable>
...
-  }
+/** An indication of what steps are necessary to make this corpus ready for serving. */
+export enum CorpusLoadingState {
+  /** Documents have not been loaded. Call `startLoading` to begin loading them. */
+  NOT_STARTED = 'NOT_STARTED',
+  /** Documents are currently being loaded. */
+  IN_PROGRESS = 'IN_PROGRESS',
+  /** Documents were successfully loaded. The corpus is ready for serving. */
+  COMPLETED = 'COMPLETED',
+  /** Documents failed to load. See details for your implementation to try to resolve. */
```
NickHeiner, line 412: We may wish to link to https://docs.ai-jsx.com/guides/observability as well.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -351,166 +380,213 @@ export interface EmbeddedChunk<ChunkMetadata extends Jsonifiable = Jsonifiable>
...
+
+/** Statistics about a corpus, including its LoadingState. */
+export interface CorpusStats {
+  loadingState: CorpusLoadingState;
+  completedPartitions: number;
+  activePartitions: number;
+  numDocuments: number;
+  numChunks: number;
+  /** If loadingState == FAILED, the error that caused it. */
+  loadingError?: Error;
```
NickHeiner, line 419: We can use TS to encode that `loadingError` will only appear when `loadingState === failed`:

https://www.typescriptlang.org/play?#code/PQKhAIEEDtwS2gEzgYwIYBc4HtbYGbgDuAFpuAM4YCmADheGgE7XjTUrUUXMCe4GbOAC2aANasMJOAxTYmtAK4MWaRP3zzK1JgDcEAcwB04EMABQ1AB615GcNWiLh4ADLY1hgMoZMrAN7m4OCgEAAi2CjOjhgMZLqs0Nj2AEbUjuAANh6I1IgmAMJomZngAAZUzBjuntAGZQJCaQYIWTmGAiTUwiZmQeAAcgDyACoA+l4jkABKIwCiYeAAvOAA5MPjkzPzYasANP2h4BFRwjEMzKxRTCzQGJn8aR3Zanm9FsEAkgNjAArTQwA4tM5l4vMs1t8-gDgaCvPtDmBjpFoncGEQdKwKIoUJxuPhFCV+C9cvlwCMuuA5AplPAVNQ1BotBQdPo6u9+gUhgBZX6uOY7CGrLm8-k7BHBI4nVGxcD4NBwTJ5RptNQmLzpcC5XyKhiaJjgXjYRQGuDCWhKs53TA4WCCARMfj2lgUbCZBIc4IAMUgn35ixWqx9foWCIAvuZzEcfDaqKgLiljfY0FT5EoKHt4NAUJlFMg6vBZTV8wYYzQOQgaEx5ZxwAAhNAsssMQLBEneXw0ABcbnadTL1AA3OYI+ZKzoa6xmwB1OBSOY3LTWGhIBgNpudlv9dv9zvUHvFjt+IzB-3Dtt9gwLpjyHvX+TD0fj6toWszuckJP3g3LxyINeNtQzbgK2qolgOd5WDmebUAAPIeu5+JmCGlnuJ6+v6AB8I6RhgvC0FOm4Qu+86LgaAA+4AkZ+igYN+kbAMA5LSHqCqZAwaToMorAYqsLAiDIFDPJe9FyNAVCUD2wErKBO6oX4B6XgO6EhmEOFRkxFIyHKbEcRwaDccQrDxKwcnfuAKS0ZQNGZIg0CrBg5hiRJFAAExSURMnbkpe6KbU8nlhsExTLMCwHBe-nfj27BEOA34ABSrPKYirAAlOpjHMdpRDyGIFBObgLkAMweZgDBeRF4G+b2-nKaeYXeZFZHRdQsUJUl4hpThQA

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -519,11 +590,69 @@ export class LocalCorpus<
...
+    return searchVectorStore(this.vectorstore, query, params);
+  }
+}
+
+async function searchVectorStore<ChunkMetadata extends Jsonifiable = Jsonifiable>(
+  vectorStore: VectorStore,
+  query: string,
+  params?: { limit?: number; score_threshold?: number }
+): Promise<ScoredChunk<ChunkMetadata>[]> {
+  const scoredLcDocs = await vectorStore.similaritySearchWithScore(query, params?.limit ?? 4, params);
```
juberti, line 637: suggest making the `4` a constant somewhere rather than buried in here

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -552,17 +681,22 @@ export interface DocsQAProps {
...
-  const status = (await props.corpus.getStats()).loadingState;
-  if (status !== Corpus.LoadingState.COMPLETED) {
-    return `Corpus is not loaded. It's in state: ${status.toString()}`;
+export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(props: DocsQAProps<ChunkMetadata>) {
+  let docs: ScoredChunk<ChunkMetadata>[];
+  try {
+    docs = await props.corpus.search(props.question, { limit: props.chunkLimit });
+  } catch (e) {
+    if (e instanceof CorpusNotReadyError) {
+      return e.message;
```
NickHeiner, line 695: I'm slightly inclined to say we should just throw the error and force the user to handle it. They can use an `ErrorBoundary` if they want.

packages/docs/docs/guides/docsqa.md
```diff
@@ -52,33 +53,34 @@ With vectors in hand, your text chunks can be added to a vector database. When r
...
 ```
 
-Typically you'll want to make sure a corpus has completed loading before exposing it to your users, but for small corpora loading just in time can work as well. (If you invoke the DocsQA primitive with a corpus that hasn't completed loading, you'll just get a message about the corpus not being ready instead of a real response from the LLM.)
-
-:::note Note:
-
-`LocalCorpus` simplifies this flow a bit for the simple use cases it supports by having a `COMPLETED` `LoadingState` by the time the promise from `startLoading()` has resolved.
-
-:::
+Typically you'll want to make sure a corpus has completed loading before exposing it to your users, but for small corpora loading just in time can work as well. (If you invoke the DocsQA primitive with a LoadableCorpus that hasn't completed loading, you'll just get a message about the corpus not being ready instead of a real response from the LLM.)
```
NickHeiner, line 83:
```suggestion
Typically you'll want to make sure a corpus has completed loading before exposing it to your users, but for small corpora loading just in time can work as well. (If you invoke the DocsQA primitive with a `LoadableCorpus` that hasn't completed loading, you'll just get a message about the corpus not being ready instead of a real response from the LLM.)
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -325,6 +327,28 @@ export const defaultEmbedding = new LangChainEmbeddingWrapper(
   new OpenAIEmbeddings({ openAIApiKey: process.env.OPENAI_API_KEY })
 );
 
+/**
+ * A function converting documents into EmbeddedChunks.
```
juberti, line 331:
```suggestion
 * A function that converts documents into EmbeddedChunks.
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -351,166 +375,213 @@ export interface EmbeddedChunk<ChunkMetadata extends Jsonifiable = Jsonifiable>
...
+  maxRequests: number = 50
+): Promise<CorpusStats> {
+  const activePartitionsToToken = new Map<CorpusPartition['name'] | null, string | null>();
+  const completedPartitions = new Set<CorpusPartition['name']>();
+  let numDocuments = 0;
+  let numChunks = 0;
+  let numRequests = 0;
+
+  activePartitionsToToken.set(null, null);
+  while (activePartitionsToToken.size > 0 && numRequests++ < maxRequests) {
```
NickHeiner, line 510: Is `maxRequests` a protection against the "crawl all of wikipedia" case?

NickHeiner: If a crawl stops due to hitting max requests, as opposed to exhausting all the partitions, is that visible to the dev?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -552,17 +681,22 @@ export interface DocsQAProps {
    *  }
    * ```
    */
-  chunkFormatter: (props: { doc: ScoredChunk }) => Node;
+  chunkFormatter: (props: { doc: ScoredChunk<ChunkMetadata> }) => Node;
```
juberti, line 684: can you make `DefaultFormatter` the default?

juberti (approved):
Nice! LG with nits

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -351,166 +375,213 @@ export interface EmbeddedChunk<ChunkMetadata extends Jsonifiable = Jsonifiable>
...
+  let numRequests = 0;
+
+  activePartitionsToToken.set(null, null);
+  while (activePartitionsToToken.size > 0 && numRequests++ < maxRequests) {
+    const [partition, pageToken] = activePartitionsToToken.entries().next().value;
+    activePartitionsToToken.delete(partition);
+
+    const response = await loader({ partition, pageToken } as CorpusLoadRequest);
+    if (!response.page && !response.partitions) {
+      throw Error('Loader responses must include a page, new partitions, or both.');
```
NickHeiner, line 516: Can we update this to be more actionable to the dev? This is clearly the dev's fault, right? 
```suggestion
      throw Error('Loader responses must include a page, new partitions, or both. However, the loader returned ${JSON.stringify(response)}. Update your loader to return the right type.');
```

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: I'm not sure I understand what you're getting at here.  Are you just wanting to point out where logs could be?

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: This is super cool. Thanks for sharing!

Unfortunately this seems to keep me from updating individual fields on the stats object, which makes it a bit clunky to use unless there's a trick I'm missing.

NickHeiner:
I don't have time to review this any further so no need to wait for me.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: sg, done.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: That and to prevent an infinite loop. I suppose it's not our resources though, so I upped the default to 5000 requests.

Devs could infer max requests were hit by viewing the corpus stats.  Non-zero active partitions with COMPLETED state implies max requests limit was reached.

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: I think so?  @NickHeiner Is this a reasonable way to do this?

mdepinet: Oh, just saw your comment.  Well it works, so let's go with it!

PR #85, opened 2023-06-19
Many updates to API documentation.

mdwelsh:
This PR brings in the other public endpoints into the API docs, and cleans up a bunch of the API documentation in various places.

Something that @petersalas did in #78 seems to have broken docusaurus's resolution of types in the top level index.ts file. I am not sure why this is happening yet.


packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -1,7 +1,9 @@
 /**
  * This module defines affordances for constraining the output of the model
  * into specific formats, such as JSON, YAML, or Markdown.
+ * @packageDocumentation
```
NickHeiner, line 4: Ah nice! This seems like a nice way to mark what's part of the external API.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -13,9 +15,11 @@ interface ValidationResult {
 
 // TODO: schema
 /**
- * A ChatCompletion component that constrains the output to be a valid JSON string.
+ * A {@link ChatCompletion} component that constrains the output to be a valid JSON string.
```
NickHeiner, line 18: Nice! This works well.

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -236,6 +275,8 @@ export function Completion(
  *      <SystemMessage>You are a nice person.</SystemMessage>
  *      {/* This is fine, because MyUserMessage renders to a UserMessage. *}
```
NickHeiner, line 276:
```suggestion
 *      {/* This is fine, because MyUserMessage renders to a UserMessage. */}
```

Whoops – now that we're rendering the docs, we see that this comment was missing a closing /.

packages/ai-jsx/src/core/error-boundary.ts
```diff
@@ -5,9 +5,11 @@ import { Node, RenderContext } from '../index.js';
  *
  * If the fallback itself throws an error, that error will be propagated. (Just like if your `catch` block throws an error.)
  *
- * ```
+ * @example
+ * ```tsx
  *  <ErrorBoundary fallback="User data could not be fetched.">
  *    {fetchUserData()}
+ *
```
NickHeiner, line 12: Hmm, do we want this? Looks a bit funny to me.

image

Also it's weird to me that "Example" is in code formatting, but not worth spending more than 30 seconds thinking about. 
image

https://ai-jsx-docs-git-mdw-api-docs-fixie-ai.vercel.app/api/modules/core_error_boundary

packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -1,11 +1,19 @@
...
 
+/**
+ * Represents properties passed to the {@link ImageGen} component.
+ */
 export interface ImageGenProps {
-  /** defines how many image samples should be produced */
+  /** The number of image samples to produce. */
   numSamples?: number;
-  /** defines the image resolution */
+  /** The image resolution. */
```
NickHeiner, line 16: Much better!

packages/ai-jsx/src/core/inline.tsx
```diff
@@ -6,10 +6,10 @@ type InlineChild = Node | InlineFn;
 type InlineChildren = InlineChild | InlineChildren[];
 
 /**
- * Use this to do a completion where you have access to the previously-rendered content.
+ * The Inline component renders each of its children in sequence, optionally passing the previously rendered content to the next child.
```
NickHeiner, line 9: I think it would be nice to link to the tutorial section that talks about `Inline`.

packages/ai-jsx/src/index.ts
```diff
@@ -1,48 +1,84 @@
...
+ *
+ * See: https://ai-jsx.com for more details.
+ *
+ * @packageDocumentation
+ */
+
 import { v4 as uuidv4 } from 'uuid';
 import { BoundLogger, NoOpLogImplementation, LogImplementation, Logger, PinoLogger } from './core/log.js';
 
+/** @hidden */
```
NickHeiner, line 12: This actually is something we'd like to expose. (Sorry, I know it's a bit guess-and-check.)

packages/docs/docusaurus.config.js
```diff
@@ -21,7 +21,7 @@ const config = {
   organizationName: 'fixie-ai', // Usually your GitHub org/user name.
   projectName: 'ai-jsx', // Usually your repo name.
 
-  onBrokenLinks: 'throw',
+  onBrokenLinks: 'warn',
```
NickHeiner, line 24: Why are we changing this?

NickHeiner (approved):
There's more to do to tighten this up, but this is a huge improvement over where we are now, so 🚢 🇮🇹 .

## Future improvement
The `destructured` stuff isn't super helpful:
image

Nor is showing the input TS expression – it would be better to show up the final output. Maybe we can use `Simplify`?
image

Some pages are kinda empty / useless:
image
image

It might be nice if we could group different API sections. Like, there's enough in DocsQA that it could be its own section.

NickHeiner:
image

not a blocker: this page isn't super useful – feels redundant with the left-hand nav.

packages/ai-jsx/src/core/error-boundary.ts
mdwelsh: That is how typedoc renders the `@example` tag. I think it's useful to have an example.

packages/docs/docusaurus.config.js
mdwelsh: I'm having some weird problems building with this branch and believe something must have broken with link resolution, so this works around that temporarily while I figure out what's going on.

PR #92, opened 2023-06-20
Add ability to stream over an API

petersalas:
- Add `ai-jsx/stream` export with streaming helpers
- Add `useAIStream` client-side React hook for streaming AI.JSX UI from an API
- Add a demo to the NextJS project

Still TODO:
- Support delta streaming so that requests are not O(n^2)
- Support non-text UI serialization/hydration
- Facilitate persistence


packages/ai-jsx/src/react/index.ts
```diff
@@ -64,6 +65,54 @@ export function useAI(children: AI.Node, onStreamStart?: () => void, onStreamEnd
...
+    fetch(...fetchArguments).then((response) => {
+      if (response.ok && response.body) {
+        setCurrentStream(fromStreamResponse(response.body) as ReadableStream<ReactModule.ReactNode>);
+      }
+    });
+  }
+
+  ReactModule.useEffect(() => {
+    let shouldStopValue = false;
+    const shouldStop = () => shouldStopValue;
```
NickHeiner, line 82:
```suggestion
    let shouldStop = false;
    const getShouldStop = () => shouldStop;
```

NickHeiner (approved):

PR #98, opened 2023-06-20
Nicer formatting for the inspector.

mdwelsh:
This PR makes a few formatting tweaks to the inspector:

https://asciinema.org/a/PYy3BhpSLwtpwm6OoCbl6eVCa

* Adds a header
* Puts borders around the two panes
* Adds a status bar with info on key shortcuts
* Adds "q" to quit
* Adds "1" to jump to first frame, "9" to jump to last frame


NickHeiner (approved):
Nice! This is much better.

Another thing that would be cool is if the frames on the final text output could sync up with the frames of the debug tree. But right now we don't have any way to do that.

PR #103, opened 2023-06-21
Allow logging to be overridden via AI.JSX tags

mdepinet:


mdwelsh:
Thanks for doing this. I guess my preference would be that this is controlled by an environment variable and it's off by default, unless someone turns it on by setting AIJSX_LOG_LEVEL or something like that. This way we can control it in any environment and not have to special-case NextJS on Vercel. WDYT?

petersalas:
Along the lines of what @mdwelsh is saying I think logging to a file is probably too-opinionated of a default.

@NickHeiner should we change the default to log to stderr at `"info"`? and/or add an optional env variable that specifies a file path?

mdepinet:
Given that things are currently broken, I'm inclined to submit now and have this debate later.  I think we'll want to keep the ability for users to override our default either way, so adding the logger prop seems like a step forward.  Thoughts?

petersalas:
@mdepinet the problem is that with arch 3 it's still broken

NickHeiner (approved):

PR #105, opened 2023-06-21
Compute delta events to make data transfer of typical cases O(n)

petersalas:
In `ai-jsx/stream` compute delta events so that the data transfer is O(n) rather than O(n^2) when doing token streaming.


juberti (approved):
Nice efficient approach!

PR #106, opened 2023-06-21
A basic NextJS tutorial app.

mdwelsh:
This app uses "Architecture #3" in which there is a Vercel Edge Function running at `/api/completion` that takes in the user message, system message, and assistant message as input. The Edge Function runs AI.JSX with these components of the LLM prompt, renders the result, and passes it back as a plain text string.

This is *not* streaming -- would appreciate advice on how to make the result stream back to the client.

This app is intentionally minimal since it's part of the tutorial series.


packages/docs/docs/tutorial/part5.md
```diff
@@ -0,0 +1,162 @@
...
+function Poem({ about }: { about: string }) {
+  const [poem, setPoem] = useState('');
+
+  useEffect(() => {
+    if (poem !== '') {
+      return;
+    }
+    const prompt = 'Write a poem about ' + about + '.';
+
+    const doCompletion = () => {
```
juberti, line 137: you'll want to use `AI.useAIStream` here instead to get streaming behavior

packages/tutorial-nextjs/src/app/page.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * The main page for the AI.JSX Next.js App Demo.
+ */
+export default function Home() {
+  return (
+    <main className={styles.main}>
+      <div>
+        <h2>AI.JSX Next.js App Demo</h2>
+      </div>
+      <div>
+        <PoemGenerator />
```
farzadab, line 52: Nit: Can you add a Suspense component so that users know something is going to load?

PR #108, opened 2023-06-21
Use AI streaming in tutorial

petersalas:


mdwelsh (approved):
I think it's a lot harder to follow this way, i.e., the actual prompt is stuffed into the default of the `useState` call, etc.) but whatever -- we can make it simpler. Is the AI stream thingy checked in?

mdwelsh:
@petersalas Can you do a `vercel deploy` on this so I can see it live?

petersalas:
@mdwelsh when I run `vercel deploy` it barfs

petersalas:
@mdwelsh re:
> I think it's a lot harder to follow this way, i.e., the actual prompt is stuffed into the default of the useState call, etc.) but whatever -- we can make it simpler. Is the AI stream thingy checked in?

If we want to optimize for simplicity there's nothing better than #4. But the problem is that one can't build anything interactive with it, so I think it's actually pretty important to show how you're expected to deal with interactivity.

petersalas:
Gonna merge because I can't figure out how to `vercel deploy` -- we can take stuff out as we see fit.

PR #122, opened 2023-06-21
Add tests for docs.tsx

mdepinet:
Also fix some jest issues that didn't show up until I needed to import from one of our source files.


NickHeiner:
> Also fix some jest issues that didn't show up until I needed to import from one of our source files.

heh, yes, I was wondering if this would cause a problem.

packages/ai-jsx/src/batteries/docs.test.ts
```diff
@@ -1,3 +1,132 @@
-test('adds 1 + 2 to equal 3', () => {
-  expect(1 + 2).toBe(3);
+import { describe, expect, test } from '@jest/globals';
```
NickHeiner, line 1: We shouldn't need to do this. Jest tests are typically written with the presumption that these globals are in scope.

packages/ai-jsx/src/batteries/docs.test.ts
```diff
@@ -1,3 +1,132 @@
...
+} from '../../dist/cjs/batteries/docs.cjs';
+
+test('defaultParser handles simple text', async () => {
+  const rawDoc = {
+    pageContent: new TextEncoder().encode('Hello world!'),
+    name: 'greeting.txt',
+    encoding: 'utf-8',
+    mimeType: 'text/plain',
+  };
+  const rawLoader: RawLoader = (_) => Promise.resolve({ page: { documents: [rawDoc] } });
```
NickHeiner, line 21:
```suggestion
  const rawLoader: RawLoader = () => Promise.resolve({ page: { documents: [rawDoc] } });
```

packages/ai-jsx/src/batteries/docs.test.ts
```diff
@@ -1,3 +1,132 @@
...
+    { name: 'doc1.txt', pageContent: ['Hello world1!'] },
+    { name: 'doc2.txt', pageContent: ['Hello world2!'] },
+    { name: 'doc3.txt', pageContent: ['Hello world3!'] },
+    { name: 'doc4.txt', pageContent: ['Hello world4!'] },
+    { name: 'doc5.txt', pageContent: ['Hello world5!'] },
+  ];
+  const loader = staticLoader(docs, 2);
+  const chunker = (doc: Document) => {
+    const chunkTexts: Array<string> = [];
+    doc.pageContent.map((content) => chunkTexts.push(...content.split(' ')));
```
NickHeiner, line 62: I would do something like:

```js
return Promise.resolve(doc.pageContent
  .map((content) => content.split(' '))
  .flat()
  .map((text) => ({ content: text, documentName: doc.name })))
```

NickHeiner (approved):

packages/ai-jsx/.eslintignore
```diff
@@ -1,3 +1,4 @@
 dist
 jest.config.ts
 scripts/move-cjs-to-dist.js
+src/**/*.test.ts
```
NickHeiner, line 4: We don't want to ignore all our tests. But we do want to do path-specific overrides in `.eslintrc.cjs` to configure the linter to be aware of the tests.

What issue were you seeing that made you do this? I can suggest the config change we need.

packages/ai-jsx/src/batteries/docs.test.ts
mdepinet: See https://github.com/Microsoft/TypeScript/issues/31226

I'm open to other ways of doing this, but essentially it seems we have to choose between also distributing our tests (and getting them to compile and run from dist) or adding this import.  I don't think adding the import is really a big deal so I opted for not needlessly increasing our distributed binary size.

packages/ai-jsx/.eslintignore
mdepinet: ESLint was complaining about the mismatch between our tsconfig and this.

ai-jsx:lint: /home/mike/repos/ai-jsx/packages/ai-jsx/src/batteries/docs.test.ts
ai-jsx:lint:   0:0  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/src/batteries/docs.test.ts` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.json
ai-jsx:lint: However, that TSConfig does not include this file. Either:
ai-jsx:lint: - Change ESLint's list of included files to not include this file
ai-jsx:lint: - Change that TSConfig to include this file
ai-jsx:lint: - Create a new TSConfig that includes this file and include it in your parserOptions.project

packages/ai-jsx/src/batteries/docs.test.ts
NickHeiner: We definitely don't want to distribute the tests, so if this import is the way to avoid that, I'm fine with it.

I'm slightly surprised that we'd need this import to avoid that, but :shrug:. :smile:

packages/ai-jsx/.eslintignore
NickHeiner: Ok, you can leave it as-is for now, and I'll revisit later.

PR #132, opened 2023-06-21
Lifting W&B Integration

farzadab:
This PR simply moves the W&B integration that we had into the official package so it can be imported.


packages/ai-jsx/src/batteries/logging-integrations.tsx
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  return result;
+}
+
+/**
+ * Exporting a forked version of wandb.
+ * You need to import this version of wandb to use the integration.
+ */
+export { wandb };
```
farzadab, line 35: @NickHeiner is there more info we should add here? E.g. "this is a temporary solution until X".
Also not sure if this is the right approach. I'm exporting this because when I tried importing it twice, it raised an error, so I did this.

packages/ai-jsx/src/batteries/logging-integrations.tsx
```diff
@@ -0,0 +1,126 @@
+/**
```
farzadab, line 1: Thoughts on the module name?
Alternatively we can add this to `core/logs` as well.

NickHeiner: I'm fine with this for now.

packages/ai-jsx/src/batteries/logging-integrations.tsx
NickHeiner: Ah, I was thinking this component would call `init` and `finish` for the user. @petersalas what do you think?

That way, we wouldn't have to export this, which is awkward.

NickHeiner:
Can you add a section to https://docs.ai-jsx.com/guides/observability about this?

packages/ai-jsx/src/batteries/logging-integrations.tsx
farzadab: I'll take another look. The `init` I feel like I can do. For `finish` I'm not quite sure.

farzadab: Also what is the difference between your fork and the official W&B package? Are you going to create a PR for them?

farzadab: I pushed `init` inside, but I don't quite know how to do it for `finish` (since I don't 100% understand what is even going on here).

petersalas: From skimming their SDK it looks like it's effectively global, so I don't think we want to call `finish` within the render itself.

Can we have the consumer pass in the `wandb` object as a prop and let them take responsibility for `init`/`finish`?

farzadab: > Can we have the consumer pass in the `wandb` object

Hmm, is that so much better? It's not like `wandb.init()` creates a new object, right?

Also there's a question for Nick: What's the difference between your fork and the original? Potentially we can't let users supply their own `wandb` if it doesn't have have Nick needed to get it to work here.

petersalas: I think the things Nick had to change were largely exporting types...but since we end up just calling `toJSON()` on it I think we could construct the objects entirely on our side without even importing the SDK.

I think that's probably the best option here? It lets us avoid needing to rely on the fork and defers the lifetime questions to the consumer.

farzadab: Not sure I understand. Which object should we create then?
I'll setup a quick meeting to go over this. The conversation is dragging on and it'll probably take more of your time anyway.

farzadab:
After talking to Peter, I believe the PR is ready. I'll merge it tomorrow at 9am if there are no comments.

petersalas (approved):
LGTM!

PR #135, opened 2023-06-21
Support REACT_APP_ env var lookups

NickHeiner:
This allows us to support people on create-react-app.

We now also see that the create-react-app-demo works without the need to eject and modify `env.cjs`.


packages/examples/src/simple-chat.tsx
```diff
@@ -1,5 +1,6 @@
 import { ChatCompletion, SystemMessage, UserMessage } from 'ai-jsx/core/completion';
-import { showInspector } from 'ai-jsx/core/inspector';
+import * as AI from 'ai-jsx';
+// import { showInspector } from 'ai-jsx/core/inspector';
```
petersalas, line 3: cruft

petersalas (approved):

PR #136, opened 2023-06-21
ImageGen in JIT UI

farzadab:
This PR:
- creates an `<Image>` tag, that renders based on the environment (URL in command line and `` tag in browser)
- adds an image to the recipe demo (react and NextJS)


packages/create-react-app-demo/src/recipe/page.tsx
```diff
@@ -102,9 +117,18 @@ export default function RecipeWrapper() {
...
-            </ChatCompletion>
+            {recipe}
+            {'\n'}
+            Now here's a link to the banner URL for the recipe:{' '}
+            <ImageGen>
+              Generate an image for a recipe name
+              <ChatCompletion>
+                <UserMessage>
+                  In two to three sentences, describe how the following recipe would look like when prepared by a chef:
+                  {recipe}
```
juberti, line 128: maybe want to try doing this based on {query}, to avoid introducing a 3rd pass?

juberti:
have a screenshot of how this works in practice?

farzadab:
Here are some examples:
image
image
image

packages/create-react-app-demo/src/recipe/page.tsx
farzadab: Yeah I did that originally and it works easily.
The main issue is image not matching the recipe. The query along is usually not descriptive enough.

At the very least, I should change the default from "beans" to an actual dish name 😅

NickHeiner:
damn now I want beans

farzadab:
I give up on trying to make this work with the `nextjs-experimental` demo. I did fix the color issue that Nick mentioned though.

packages/ai-jsx/package.json
```diff
@@ -15,7 +15,6 @@
         "default": "./dist/esm/index.js"
       },
       "require": {
-        "types": "./dist/cjs/index.d.cts",
```
farzadab, line 18: I had some issues with `yarn build` and as part of the fix, we also removed the `.cts` files that don't actually exist.

packages/nextjs-experimental/next.config.js
```diff
@@ -9,4 +9,16 @@ const nextConfig = {
       },
     ];
   },
+  images: {
+    remotePatterns: [
+      {
+        protocol: 'https',
+        hostname: 'oaidalleapiprodscus.blob.core.windows.net',
```
farzadab, line 16: I had to add this so that we can show Dalle images.

packages/nextjs-experimental/next.config.js
```diff
@@ -9,4 +9,16 @@ const nextConfig = {
...
+        port: '',
+        pathname: '**',
+      },
+    ],
+  },
 };
+
+module.exports = nextConfig;
```
farzadab, line 24: Turns out we weren't exporting this.
I did notice an issue though: after this fix, the `yarn workspace docs start` command loads the `localhost:3000/basic-completion` page by default which is not found. Not sure how they are related or how to fix TBH.

packages/nextjs-experimental/tailwind.config.js
```diff
@@ -7,10 +7,24 @@ module.exports = {
...
   theme: {
     extend: {
+      fontFamily: {
+        sans: ['Source Sans Pro', 'sans-serif'],
+      },
       backgroundImage: {
         'gradient-radial': 'radial-gradient(var(--tw-gradient-stops))',
         'gradient-conic': 'conic-gradient(from 180deg at 50% 50%, var(--tw-gradient-stops))',
       },
+      colors: {
```
farzadab, line 17: Fixing colors not showing up: https://fixie-ai.slack.com/archives/C05B6EE9JE5/p1687460043947939

packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -77,14 +79,45 @@ export function ImageGenProvider<T extends ImageGenPropsWithChildren>(
...
+    <AIR.React>
+      {urls.split('\n').map((url) =>
+        ReactModule.createElement('img', {
+          src: url,
+          alt: alt ?? 'Generated image',
+          width: width ?? (props.size ? props.size.split('x')[0] : undefined),
+          height: height ?? (props.size ? props.size.split('x')[1] : undefined),
+        })
+      )}
+    </AIR.React>
```
farzadab, line 122: Any thoughts on whether this use is Kosher @NickHeiner?

packages/nextjs-experimental/tailwind.config.js
NickHeiner: Nice!

packages/nextjs-experimental/src/app/recipe/page.tsx
```diff
@@ -65,16 +68,32 @@ export function RecipeInstructionListItem({ children }: { children: React.ReactN
...
+  );
+
   return (
     <>
       <InputPrompt label="What would you like a recipe for?" defaultValue={defaultValue} />
 
       <ResultContainer title={`AI comes up with a recipe for "${query}"`}>
         <Suspense fallback={'Loading...'}>
           <AI.jsx>
+            <ImageGenHTML size="256x256">
```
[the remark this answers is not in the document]
NickHeiner: Ugh, lol, very interesting. Maybe we can tweak the prompt below to make it less likely to generate "problematic" prompts?

packages/nextjs-experimental/next.config.js
NickHeiner: I think I only notice that when I have the docs and nextjs servers running at the same time (which it generally seems like I shouldn't be doing. :smile:)

packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -77,14 +79,46 @@ export function ImageGenProvider<T extends ImageGenPropsWithChildren>(
...
+  },
+  { render }: RenderContext
+) {
+  const urls = await render(<ImageGen {...props}>{children}</ImageGen>);
+  return (
+    <AIR.React>
+      {urls.split('\n').map((url) =>
+        ReactModule.createElement('img', {
+          src: url,
+          alt: alt ?? 'Generated image',
```
NickHeiner, line 117: If we happened to know the prompt, it would be nice to put here. But that's probably too tricky to wire up.

packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -77,14 +79,46 @@ export function ImageGenProvider<T extends ImageGenPropsWithChildren>(
...
+          width: width ?? (props.size ? props.size.split('x')[0] : undefined),
+          height: height ?? (props.size ? props.size.split('x')[1] : undefined),
```
NickHeiner, line 119: `props.size.split('x')[1]` is this part necessary? If `width` and `height` are unspecified, I think the `img` may default to the actual size of the image, so setting it explicitly by looking at the `size` input param would be redundant.

packages/ai-jsx/src/core/image-gen.tsx
NickHeiner: I might call it `AIReact` but this is internal code so nbd.

packages/ai-jsx/src/core/image-gen.tsx
```diff
@@ -77,14 +79,46 @@ export function ImageGenProvider<T extends ImageGenPropsWithChildren>(
...
+  }: ImageGenPropsWithChildren & {
+    /** Width for the HTML image tag. */ width?: string;
+    /** Height for the HTML image tag. */ height?: string;
+    /** Alternative text for the HTML image tag.  */ alt?: string;
+  },
+  { render }: RenderContext
+) {
+  const urls = await render(<ImageGen {...props}>{children}</ImageGen>);
+  return (
+    <AIR.React>
```
petersalas, line 113: Per our conversation this morning I think this is backwards -- I feel strongly that users shouldn't have to select the right `ImageGenX` variant with knowledge of the eventual rendering surface. (This basically means any component that wants to do an `ImageGen` internally needs to either know where it's going to end up or invent an affordance to plumb it down.)

- My first choice is to add code at the React/AI integration boundary to handle this
- My second choice would be to use context providers to allow ancestors to override the rendering behavior

At a high level I think it's quite important to decouple the abstract representation of the image in AI.JSX (where it's not hard to imagine components that would pull the image and read the bytes) from the way it gets rendered in eventual UI.

packages/ai-jsx/src/core/image-gen.tsx
petersalas: If we had an abstract image representation I think this would be straightforward :)

packages/ai-jsx/src/core/image-gen.tsx
petersalas: to be precise, https://github.com/fixie-ai/ai-jsx/pull/160 is what I'm proposing

NickHeiner: I feel slightly queasy about that but I also see the appeal of what you're saying, and I think if we chatted about it more, I'd update in your direction. So I'm fine to go with this approach.

packages/ai-jsx/src/core/image-gen.tsx
farzadab: removed.

packages/ai-jsx/src/core/image-gen.tsx
farzadab: Going with Peter's approach as discussed.

packages/ai-jsx/src/core/image-gen.tsx
farzadab: Done. The `<Image>` tag now holds a `prompt` and a `modelName` property. These are never rendered however and can only be used for debugging at the moment.

farzadab:
PTAL @juberti. I'll update the changelog and version right before merging.

packages/docs/docs/changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [0.5.8]
```
NickHeiner, line 3: We shouldn't make things a link if there's nothing to link to.

PR #142, opened 2023-06-23
Remove need for module resolution esnext

NickHeiner:
the same thing we do every day, pinky: muck with ESM

Previously, consumers need to set `"moduleResolution": "nodenext"` in their `tsconfig.json`. Otherwise, importing from `ai-jsx` fails.

I confirmed the original problem, and that this PR fixes it, in https://github.com/NickHeiner/esm-cjs-test.

The reason for the problem: in the legacy module resolution scheme, an import to `ai-jsx/stream` would be resolved by looking for these files:

```
node_modules/ai-jsx/stream.js
node_modules/ai-jsx/stream/index.js
```

We don't have files at that location, because we use the `package.json` `exports` field. But that field isn't respected in the legacy resolution scheme, so we got TS build errors.

To fix it, I added a script that writes proxy files in the locations the legacy resolution algo expects to find them.


packages/ai-jsx/package.json
```diff
@@ -69,16 +69,6 @@
         "default": "./dist/cjs/batteries/use-tools.cjs"
       }
     },
-    "./batteries/langchain-wrapper": {
```
NickHeiner, line 72: This was a dead reference. One benefit of our script is that dead references will now be caught.

packages/ai-jsx/package.json
```diff
@@ -295,19 +285,21 @@
     "eslint-config-nth": "^2.0.1",
     "globby": "^13.1.4",
     "jest": "^29.5.0",
+    "load-json-file": "^7.0.1",
     "node-fetch": "^3.3.1",
     "pino-pretty": "^10.0.0",
     "react": "^18.2.0",
     "ts-jest": "^29.1.0",
     "ts-node": "^10.9.1",
-    "tsx": "^3.12.7",
```
NickHeiner, line 303: This was unused.

PR #143, opened 2023-06-23
Update observability.md

larryatfixie:
Clarified the logging docs since we default to console now instead of a file, and need to set level='debug'


packages/docs/docs/guides/observability.md
```diff
@@ -46,14 +47,14 @@ function App() {
 console.log(
   await AI.createRenderContext({
     // highlight-next-line
-    logger: new PinoLogger(),
+    logger: new PinoLogger(pino({ level: 'debug' }),  //default level is 'info'
```
NickHeiner, line 50: This line is missing a `)`.

packages/docs/docs/guides/observability.md
```diff
@@ -46,14 +47,14 @@ function App() {
...
   }).render(<App />)
 );
 ```
 
-Now, when you run, you'll see a file created called `ai-jsx.log`. This file will have a bunch of lines in it like:
+Now, when you run, you'll see something like this on the console:
 
-```json file="ai-jsx.log"
+```json file="logfile.log"
```
NickHeiner, line 57: We shouldn't be using the `file` label if the contents aren't actually a file.

PR #144, opened 2023-06-23
Add FixieCorpus implementation capable of using existing Fixie corpora.

mdepinet:


packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -600,6 +600,45 @@ export class LocalCorpus<
   }
 }
 
+/** A fully mananged {@link Corpus} served by Fixie. */
```
NickHeiner, line 603: It's fine if it's a follow-up PR, but let's add a guide/tutorial handholding the reader through setting up a Fixie corpus + using it here. Maybe @mdwelsh can help with that.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -600,6 +600,45 @@ export class LocalCorpus<
...
+export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implements Corpus<ChunkMetadata> {
+  private static readonly FIXIE_API_URL = 'https://app.fixie.ai/api';
+
+  constructor(private readonly corpus_id: string, private readonly fixie_api_key?: string) {
+    if (!fixie_api_key) {
+      this.fixie_api_key = getEnvVar('FIXIE_API_KEY');
+      if (!this.fixie_api_key) {
+        throw new Error(
+          'You must provide a Fixie API key to access Fixie corpora. Find yours at https://app.fixie.ai/profile'
```
NickHeiner, line 612:
```suggestion
          'You must provide a Fixie API key to access Fixie corpora. Find yours at https://app.fixie.ai/profile.'
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -600,6 +600,45 @@ export class LocalCorpus<
   }
 }
 
+/** A fully mananged {@link Corpus} served by Fixie. */
+export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implements Corpus<ChunkMetadata> {
+  private static readonly FIXIE_API_URL = 'https://app.fixie.ai/api';
+
+  constructor(private readonly corpus_id: string, private readonly fixie_api_key?: string) {
+    if (!fixie_api_key) {
+      this.fixie_api_key = getEnvVar('FIXIE_API_KEY');
```
NickHeiner, line 609:
```suggestion
      this.fixie_api_key = getEnvVar('FIXIE_API_KEY', false);
```

Pass `shouldThrow = false` here since you want to handle the error yourself.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -600,6 +600,45 @@ export class LocalCorpus<
   }
 }
 
+/** A fully mananged {@link Corpus} served by Fixie. */
+export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implements Corpus<ChunkMetadata> {
+  private static readonly FIXIE_API_URL = 'https://app.fixie.ai/api';
```
NickHeiner, line 605: I think we'll want to make this readable via env var as well so people can proxy it if they want, just like they proxy OpenAI calls in [Arch 2](https://docs.ai-jsx.com/guides/architecture?_highlight=arch#ui--aijsx-on-the-client-api-calls-on-the-server).

NickHeiner: See also https://docs.ai-jsx.com/guides/openai#set-a-proxy-env-var.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -600,6 +600,45 @@ export class LocalCorpus<
...
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${this.fixie_api_key}`,
+      },
+      body: JSON.stringify({ query_string: query, chunk_limit: params?.limit }),
+    });
+    if (response.status !== 200) {
+      throw new Error(`Fixie API returned status ${response.status}: ${await response.text()}`);
+    }
+    const api_results = await response.json();
```
NickHeiner, line 630:
```suggestion
    const apiResults = await response.json();
```

We use camelcase in JS. :smile:

NickHeiner (approved):

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: Yep, will do (though any reasonable directions depend on my next backend PR).

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: Whoops, thanks. It'd be awesome if our formatter or linter could catch that!

PR #147, opened 2023-06-23
A tutorial for the NaturalLanguageRouter.

mdwelsh:
This is intended to be part 6 of the tutorial, showing how to use the NaturalLanguageRouter component as part of a simple 20-questions game about animals.

The idea is that the AI thinks of an animal and the user gets to ask up to 20 yes/no questions. We use NaturalLanguageRouter twice: Once to filter out bad questions from the user, and another time to match the question
against the secret animal.

I'll write documentation about this once we decide the code is good.


packages/ai-jsx/src/batteries/natural-language-router.tsx
```diff
@@ -27,7 +27,7 @@ const noMatch = 'None of the routes match what the user said.';
  *      <Route when='the user would like to upgrade'>
  *        <Upgrade />
  *      </Route>
- *      <Route noMatch>
+ *      <Route unmatched>
```
NickHeiner, line 30: 😅 thanks!

packages/tutorial/src/part6.tsx
```diff
@@ -0,0 +1,124 @@
...
+const { prompt } = enquirer;
+
+/** Prompt the user for their question. */
+async function getQuestion(questionNumber: number, maxQuestions: number) {
+  const response = await prompt({
+    type: 'input',
+    name: 'question',
+    message: `[Question ${questionNumber + 1}/${maxQuestions}] Your question`,
+  });
+  // @ts-ignore
```
NickHeiner, line 16:
```suggestion
  // @ts-expect-error
```

We always want to use `ts-expect-error`. That way, if the underlying error goes away, we get a warning, and can remove the comment.

packages/tutorial/src/part6.tsx
```diff
@@ -0,0 +1,124 @@
...
+            right! The animal I was thinking of was..."
+          </SystemMessage>
+          <UserMessage>🎉 You got it right! The animal I was thinking of was {animal}.</UserMessage>
+        </ChatCompletion>
+      </Route>
+      <Route when={correctRoute}>
+        <ChatCompletion>
+          <SystemMessage>
+            Answer the following question about the animal "{animal}" in the affirmative, using the form: "👍 Yes, the
+            animal I am thinking of ...". Do not, under any circumstances, reveal any other information, especially not
```
NickHeiner, line 69: > Do not, under any circumstances, reveal any other information,

hehehehe I'm guessing this was a problem? These models are SO HELPFUL.

packages/tutorial/src/part6.tsx
```diff
@@ -0,0 +1,124 @@
...
+  console.log(`${answer}\n`);
+  numQuestions++;
+  if (answer.includes('You got it right!')) {
+    console.log('You win!');
+    break;
+  }
+}
+
+if (numQuestions === MAX_QUESTIONS) {
+  console.log(`I'm sorry, the animal I was thinking of was a ${animal}.`);
```
NickHeiner, line 123: minor: We could do this with model call as well to get the `a` vs `an` phrasing correct.

NickHeiner (approved):
I didn't run this locally, but the code looks great.

PR #148, opened 2023-06-23
[WIP] Integrate with Vercel AI

NickHeiner:


packages/ai-jsx/src/core/completion.tsx
```diff
@@ -194,6 +195,20 @@ export function AssistantMessage({ children }: { children: Node }) {
...
+  return messages.map((message) => {
+    if (message.role === 'assistant') {
+      return <AssistantMessage>{message.content}</AssistantMessage>;
+    }
+    if (message.role === 'user') {
+      return <UserMessage>{message.content}</UserMessage>;
+    }
+    if (message.role === 'function') {
+      return <FunctionCall name={message.function_call!.name!} args={JSON.parse(message.function_call!.arguments!)} />;
+    }
```
NickHeiner, line 208: I'm not totally sure I did the right thing here.

packages/ai-jsx/src/stream/index.ts
```diff
@@ -94,6 +94,7 @@ export function toStreamResponse(renderable: Renderable): Response {
       status: 200,
       headers: {
         'Content-Type': 'text/event-stream',
+        'x-powered-by': 'ai-jsx',
```
NickHeiner, line 97: This feels more like a vanity header than something that gets interpreted by the client. I was thinking of setting `Content-Type` to something like `text/event-stream-ai-jsx`, but that seems bad because this is a valid SSE stream, so I want to leave this header as-is.

Any other suggestions for what header to set?

PR #150, opened 2023-06-26
Rename tutorial files.

mdwelsh:
As we add more tutorials it may not be clear what order these should go in, so rather than have the files named "part1", "part2", etc. I just renamed them to be more descriptive. We can migrate away from the "partN" naming for the tutorial docs as well if we like, for now I'm just renaming the source files.


NickHeiner (approved):
Looks great! This will be more legible.

PR #151, opened 2023-06-26
Test Cleanup

NickHeiner:
1. Remove the need to manually `import` Jest globals.
2. Introduce `eslint-plugin-jest`
3. Introduce a dedicated tsconfig for tests so the tests are typechecked
4. Fix newly-found lint errors in test code, now that the tests being linted.


packages/ai-jsx/test/README.md
```diff
@@ -0,0 +1,3 @@
+Originally, our test files lived alongside the code they tested. However, this confused VS Code – it couldn't figure out which `tsconfig.json` to use.
```
mdepinet, line 1: Consider describing where we expect test files to live instead.

```suggestion
Originally, our test files lived alongside the code they tested. However, this confused VS Code – it couldn't figure out which `tsconfig.json` to use. Instead we place test files in a test-only sibling directory. Each source file src/foo/bar/baz.ts{x} should have its test at test/foo/bar/baz_test.ts{x}.
```

packages/ai-jsx/test/docs.ts
mdepinet, line 1: Should this be at packages/ai-jsx/test/batteries/docs_test.ts?

Reasoning:
Added "batteries" directory because I don't think we want a flat directory for tests so mirroring the source tree makes sense.
Added "_test" to the name so that searching by file name makes it easy to distinguish between source and its test.

packages/ai-jsx/test/util.ts
mdepinet, line 1: Ditto re: name and location.

mdepinet (approved):

packages/ai-jsx/test/docs.ts
NickHeiner: Heh, I chose a flat directory for simplicity for now. But I'm happy to mirror the source tree.

When I search by file name, I use `cmd+p`, which searches the whole path, which already contains `test`, so there's no need to put `test` in the path again. But I'm fine to change it.

PR #152, opened 2023-06-26
Tutorial for UseTools.

mdwelsh:
This is a simple tutorial showing how to use the UseTools component.

As with all of the tutorials, the goal here is not to be completely rock-solid as an end-user application, but rather to demonstrate the AI.JSX capability in a simple way.

In this case we show UseTools being used to answer questions about stock prices, and to graph historical stock price data.


packages/tutorial/src/tools.tsx
```diff
@@ -0,0 +1,64 @@
...
+import yahooFinance from 'yahoo-finance2';
+import asciichart from 'asciichart';
+import enquirer from 'enquirer';
+
+const { prompt } = enquirer;
+
+function StockAgent(props: { query: string }) {
+  async function checkStockPrice(symbol: string) {
+    const quote = await yahooFinance.quote(symbol);
+    return quote.regularMarketPrice ?? 'Unknown';
```
farzadab, line 13: I kept getting an error here when I didn't use a proper symbol. You could check if `quote` is undefined or not.

PR #153, opened 2023-06-27
Add `ConversationalHistory` component

NickHeiner:
This is a subset of #148. I'm re-opening this PR with the upstream as the compare (rather than my fork), since Vercel deployments don't seem to work when sending a PR from a fork.


packages/ai-jsx/src/core/completion.tsx
```diff
@@ -203,6 +204,20 @@ export function AssistantMessage({ children }: { children: Node }) {
   return children;
 }
 
+export function ConversationHistory({ messages }: { messages: ChatCompletionResponseMessage[] }) {
+  return messages.map((message) => {
+    if (message.role === 'assistant') {
+      return <AssistantMessage>{message.content}</AssistantMessage>;
+    }
+    if (message.role === 'user') {
+      return <UserMessage>{message.content}</UserMessage>;
```
petersalas, line 213: Assign name as well

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -203,6 +204,20 @@ export function AssistantMessage({ children }: { children: Node }) {
   return children;
 }
 
+export function ConversationHistory({ messages }: { messages: ChatCompletionResponseMessage[] }) {
+  return messages.map((message) => {
+    if (message.role === 'assistant') {
```
petersalas, line 209: System?

petersalas (approved):

packages/ai-jsx/src/core/completion.tsx
NickHeiner: I see `name` on `UserMessage` but I'm not sure what it means or what it would be used for.

petersalas: You can assign names to users in the conversation history to make it clear to the model that there are different users. Assuming the JSON type you're mapping is from OpenAI it should be there too?

petersalas:
Idk what happened to the comment thread (thanks GH) but I see you're using the response type which maybe explains why name isn't on the type? I think it's on the input type but not a huge deal

packages/ai-jsx/src/core/completion.tsx
NickHeiner: Hmm, I don't see that in the type. But maybe that's because it's the `openai-edge` type which lags behind.

PR #156, opened 2023-06-27
Implement basic UI streaming

petersalas:
This implements basic UI "streaming" with Next and replicates the `Recipe` example via "arch 3".

Some caveats:
- `<AI.JSX>` nested within `<AI.React>` is not yet supported (which means that there's no way to have AI.JSX content streaming "inside" a React component delivered by the API)
- Components need to be registered in a `ComponentMap` that is used both by the API to serialize and the client to rehydrate.


NickHeiner:
Let's make sure the docs for this are updated before you're OOO. (Maybe I'll push to this branch to do it.)

packages/ai-jsx/src/experimental/next/index.ts
```diff
@@ -145,3 +147,8 @@ export const jsx = asJsxBoundary(function jsx(
   }) as JSX.Element;
 });
 export const JSX = jsx;
+
+export function toReactStream(componentMap: ComponentMap<any>, renderable: AI.Renderable): Response {
+  const renderResult = AI.createRenderContext().render(renderable, { stop: (e) => e.tag == AI.React, map: (x) => x });
```
NickHeiner, line 152: minor: is there a reason not to make the identity function the default value for `map`?

packages/ai-jsx/src/react/core.ts
```diff
@@ -35,3 +38,25 @@ export function React(
...
   return children;
 }
+
+/**
+ * Creates a seralizer function serializes React components.
+ */
+export function createElementSerializer(map: ComponentMap<any>): ElementSerializer {
+  return (element: AI.Element<any>) => {
+    if (isJsxBoundary(element.tag)) {
+      throw new Error('Serializing AI.JSX components within React components is not yet supported.');
```
NickHeiner, line 48: Can you use the new `AIJSXError` for this? :smile:

packages/ai-jsx/src/react/index.ts
```diff
@@ -85,6 +87,12 @@ export interface UseAIStreamOpts {
    * An event handler that runs when the request fails.
    */
   onError?: (error: Error) => void;
+
+  /**
+   * A map of component names to React components. This is used to deserialize
+   * the stream.
```
NickHeiner, line 93: This is a user-facing type, right? Let's add a bit more detail / an example here. Imagine the user doesn't know anything about AI.JSX.

packages/ai-jsx/src/react/index.ts
```diff
@@ -107,6 +115,37 @@ export interface UseAIStreamResult {
...
+
+      if (typeof parsed.$$component === 'string') {
+        return ReactModule.createElement(parsed.$$component, parsed.props);
+      }
+
+      if (typeof parsed.$$component === 'object') {
+        const id = (parsed.$$component as Record<string, any>).id;
+        const component = componentMap?.idToComponent.get(id);
+        if (component === undefined) {
+          throw new Error(`Unknown component ${id}`);
```
NickHeiner, line 137: Please use `AIJSXError` here.

packages/nextjs-demo/src/app/recipe/page.tsx
```diff
@@ -0,0 +1,19 @@
+import InputPrompt from '@/components/InputPrompt';
+import ResultContainer from '@/components/ResultContainer';
+import { RecipeGenerator } from '@/components/Recipe';
+
+export default function BasicCompletion({ searchParams }: { searchParams: any }) {
```
juberti, line 5: maybe call this something else other than `BasicCompletion`?

packages/ai-jsx/src/react/serialize.tsx
```diff
@@ -20,7 +20,7 @@ export function Serialize({ children }: { children: React.ReactNode }): AI.Rende
     const typeName = reactComponentName(child.type);
     // TODO: support prop serialization
     if (typeof child.props === 'object' && child.props && 'children' in child.props) {
-      return [`<${typeName}>`, <Serialize>{child.props.children}</Serialize>, `</${typeName}`];
+      return [`<${typeName}>`, <Serialize>{child.props.children}</Serialize>, `</${typeName}>`];
```
NickHeiner, line 23: lol whoops

packages/ai-jsx/src/stream/index.ts
```diff
@@ -99,13 +139,42 @@ export function toStreamResponse(renderable: Renderable): Response {
...
+  renderResult: RenderResult<PartiallyRendered[], PartiallyRendered[]>,
+  serializer: ElementSerializer
+): Response {
+  const jsonReplacer = jsonReplacerFromSerializer(serializer);
+  return new Response(
+    toEventStream(renderResult)
+      .pipeThrough(
+        new TransformStream<StreamEvent<PartiallyRendered>, string>({
+          transform(streamEvent, controller) {
+            controller.enqueue(`data: ${JSON.stringify(streamEvent, jsonReplacer)}\n\n`);
```
NickHeiner, line 156: It's fun to actually get to use the `replacer` arg for once.

NickHeiner:
I'm getting some of this when I run locally:

```
AIJSXError: The model did not produce a valid JSON object, even after 3 attempts.
```

So @hessamb / @farzadab it would be great to get OpenAI functions working to see if that helps with this.

NickHeiner:
> This implements basic UI "streaming"

heh, I see now why 'streaming' was in quotes.

packages/ai-jsx/src/experimental/next/index.ts
NickHeiner, line 1: In this PR or a follow-up: move anything required for Arch 3 out of `experimental`.

packages/ai-jsx/src/stream/index.ts
```diff
@@ -1,27 +1,43 @@
...
     }
-  | { type: 'replace'; index: number; content: PartiallyRendered }
-  | { type: 'multiple'; events: StreamEvent[] }
-  | { type: 'append'; index: number; content: string }
+  | { type: 'replace'; index: number; content: T }
+  | { type: 'multiple'; events: StreamEvent<T>[] }
+  | { type: 'append'; index: number; content: T }
   | { type: 'complete' };
 
+export type ElementSerializer = (element: Element<any>) => Jsonifiable;
```
NickHeiner, line 23: We may need the `@hidden` tag here.

packages/ai-jsx/src/stream/index.ts
```diff
@@ -56,14 +71,38 @@ async function* renderToJsonEvents(renderable: Renderable): AsyncGenerator<Strea
...
+function jsonReplacerFromSerializer(serializer: ElementSerializer) {
+  return (key: string, value: unknown) => {
+    let currentValue: unknown = value;
+
+    if (currentValue !== null && typeof currentValue === 'object') {
+      while (isIndirectNode(currentValue)) {
+        currentValue = getReferencedNode(currentValue);
+      }
+
+      if (isElement(currentValue)) {
```
NickHeiner, line 86: rename to `isAIElement`?

packages/ai-jsx/src/stream/index.ts
```diff
@@ -56,14 +71,38 @@ async function* renderToJsonEvents(renderable: Renderable): AsyncGenerator<Strea
     }
   }
 
-  return { type: 'complete' } as StreamEvent;
+  return { type: 'complete' };
+}
+
+function jsonReplacerFromSerializer(serializer: ElementSerializer) {
```
NickHeiner, line 77: Maybe add comment about this function normalizing everything to AI.JSX

packages/ai-jsx/src/stream/index.ts
```diff
@@ -56,14 +71,38 @@ async function* renderToJsonEvents(renderable: Renderable): AsyncGenerator<Strea
...
+        return serializer(currentValue);
+      }
+    }
+
+    return currentValue;
+  };
+}
+
+function jsonReviverFromDeserializer<T>(deserializer: ElementDeserializer<T>) {
+  return (key: string, value: Jsonifiable) => deserializer(value) || value;
```
NickHeiner, line 96:
```suggestion
  return (key: string, value: Jsonifiable) => deserializer(value);
```

Maybe this can be removed?

packages/ai-jsx/src/stream/index.ts
```diff
@@ -80,10 +119,11 @@ export function toEventStream(renderable: Renderable): ReadableStream<StreamEven
  * content as SSE events.
  */
 export function toStreamResponse(renderable: Renderable): Response {
+  const renderResult = createRenderContext().render(renderable, { stop: () => false, map: (x) => x });
```
NickHeiner, line 122: `stop` is maybe unneeded here.

NickHeiner (approved):

packages/nextjs-demo/src/components/Recipe.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const { current, fetchAI } = useAIStream({
+    componentMap: RecipeMap,
+    onComplete: (x) => {
+      setIsLoading(false);
+      return x;
+    },
+  });
+
+  if (activeTopic !== topic) {
```
NickHeiner, line 77: Is there a reason not to use `useEffect` here?

packages/nextjs-demo/src/components/Recipe.map.tsx
```diff
@@ -0,0 +1,4 @@
+import { makeComponentMap } from 'ai-jsx/react/map';
+import * as RecipeComponents from './Recipe';
+
+export default makeComponentMap(RecipeComponents);
```
NickHeiner, line 4: fyi @petersalas

packages/nextjs-demo/src/app/recipe/page.tsx
NickHeiner: Good catch!

packages/ai-jsx/src/experimental/next/index.ts
petersalas: Sadly yes -- it's a speed-bump to make it harder to accidentally do the wrong thing with partial rendering. (I'm inclined to condemn partial rendering to a lower-level API in the future because it's IMO not worth the ambiguity around attempting to make it ergonomic.)

packages/ai-jsx/src/stream/index.ts
petersalas: Unforunately not because the presence of a `stop` is what indicates to `render` that the result should be an array of strings rather than a string. (See other comment about my regrets on the altitude of this API.)

packages/nextjs-demo/src/components/Recipe.map.tsx
petersalas: Nice!

packages/nextjs-demo/src/components/Recipe.tsx
petersalas: Probably not? I think I was trying to avoid the effect being fired multiple times in strict mode but that was likely misguided/wrong

petersalas:
@NickHeiner re:

> AIJSXError: The model did not produce a valid JSON object, even after 3 attempts.

I went and switched this back to `ChatCompletion`, empirically it is much more reliable. (At least with gpt-3.5)

packages/ai-jsx/src/stream/index.ts
petersalas: I left this way for symmetry with `ElementSerializer` and for clarity of types -- `ElementDeserializer` is responsible for deserializing anything that was serialized by `ElementSerializer` but otherwise returns `undefined` to let things deserialize as-is. (Namely `StreamEvent` needs to be preserved.) I started to change the semantics but realized that it was going to nerf the type signature.

packages/ai-jsx/src/experimental/next/index.ts
petersalas: Will lazily/selfishly leave this for a followup

PR #158, opened 2023-06-27
Pinecone DocsQA example with LangChainCorpus

farzadab:
This PR adds an example of using a Pincone database through a LangChainCorpus to the tutorials.

### DIFF
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -706,8 +706,7 @@ async function searchVectorStore<ChunkMetadata extends Jsonifiable = Jsonifiable
   params?: { limit?: number; score_threshold?: number }
 ): Promise<ScoredChunk<ChunkMetadata>[]> {
   const k = params?.limit ?? defaultLangchainChunkLimit;
-  delete params?.limit;
-  const scoredLcDocs = await vectorStore.similaritySearchWithScore(query, k, params);
+  const scoredLcDocs = await vectorStore.similaritySearchWithScore(query, k, _.omit(params, 'limit'));
   return scoredLcDocs.map((lcDocAndScore) => {
     const lcDoc = lcDocAndScore[0];
     return {
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -705,11 +705,9 @@ async function searchVectorStore<ChunkMetadata extends Jsonifiable = Jsonifiable
   query: string,
   params?: { limit?: number; score_threshold?: number }
 ): Promise<ScoredChunk<ChunkMetadata>[]> {
-  const scoredLcDocs = await vectorStore.similaritySearchWithScore(
-    query,
-    params?.limit ?? defaultLangchainChunkLimit,
-    params
-  );
+  const k = params?.limit ?? defaultLangchainChunkLimit;
+  delete params?.limit;
+  const scoredLcDocs = await vectorStore.similaritySearchWithScore(query, k, params);
   return scoredLcDocs.map((lcDocAndScore) => {
     const lcDoc = lcDocAndScore[0];
     return {
--- a/packages/docs/docs/tutorial/part4-docsqa.md
+++ b/packages/docs/docs/tutorial/part4-docsqa.md
@@ -120,48 +120,47 @@ To do so, you can use [LangchainCorpus](../api/classes/batteries_docs.LangChainC
 const corpus = new LangchainCorpus(await getVectorStore());
 ```
 
-Here is an example where we build a DocsQA component from AI.JSX documentation using Pinecone:
+Here is an example where we build a DocsQA component from an existing Pinecone index:
 
 ```tsx
 async function getVectorStore(): Promise<VectorStore> {
   const client = new PineconeClient();
   await client.init({
-    apiKey: pineconeConfig.apiKey,
-    environment: pineconeConfig.environment,
-  });
-  const pineconeIndex = client.Index(pineconeConfig.index);
-
-  // For this example, we will index the ai-jsx docs.
-  const loader = new DirectoryLoader('../../packages/docs/docs/', {
-    '.md': (path) => new TextLoader(path),
+    apiKey: process.env.PINECONE_API_KEY,
+    environment: process.env.PINECONE_ENVIRONMENT,
   });
 
-  const splitter = new RecursiveCharacterTextSplitter({ chunkSize: 1000, chunkOverlap: 100 });
-  const docs = await loader.loadAndSplit(splitter);
-
-  const vectorStore = await PineconeStore.fromDocuments(docs, new OpenAIEmbeddings(), {
-    pineconeIndex,
-    namespace: pineconeConfig.namespace,
+  const vectorStore = await PineconeStore.fromExistingIndex(new OpenAIEmbeddings(), {
+    pineconeIndex: client.Index(process.env.PINECONE_INDEX);,
+    namespace: process.env.PINECONE_NAMESPACE,
   });
+  return vectorStore;
 }
 ```
 
-Once we have a `VectorStore` object, we wrap it with a `LangChainCorpus` and we can use it as we did before:
+Once we have a `VectorStore` object, we wrap it with a `LangChainCorpus`:
 
 ```tsx
 const corpus = new LangchainCorpus(await getVectorStore());
+```
 
+The above assumes that the vector store is already populated, but you can also use the same syntax as before to add documents to it:
+
+```tsx
+// or use a loadable corpus
+const corpus = new LoadableLangchainCorpus(await getVectorStore(), staticLoader(docs), makeChunker(600, 100));
+await corpus.load();
+```
+
+Once you have the `corpus` object, you can ask questions from it just as before:
+
+```tsx
 function App() {
   return (
     <>
-      <DocsQA
-        question="What is the advantage of using JIT UI?"
-        corpus={corpus}
-        chunkLimit={4}
-        chunkFormatter={GetChunk}
-      />
+      <DocsQA question="What was Hurricane Katrina?" corpus={corpus} chunkLimit={5} chunkFormatter={GetChunk} />
       {'\n\n'}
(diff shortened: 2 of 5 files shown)


packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -705,11 +705,9 @@ async function searchVectorStore<ChunkMetadata extends Jsonifiable = Jsonifiable
...
   params?: { limit?: number; score_threshold?: number }
 ): Promise<ScoredChunk<ChunkMetadata>[]> {
-  const scoredLcDocs = await vectorStore.similaritySearchWithScore(
-    query,
-    params?.limit ?? defaultLangchainChunkLimit,
-    params
-  );
+  const k = params?.limit ?? defaultLangchainChunkLimit;
+  delete params?.limit;
+  const scoredLcDocs = await vectorStore.similaritySearchWithScore(query, k, params);
```
farzadab, line 710: @mdepinet I feel like this was a bug, but I defer that to you. The issue was that `limit` was being passed as a `filter` to Pinecone and I kept getting an empty response. It took me too long to figure out the issue was on our side, not in Pinecone/LangChain 😅

packages/docs/docs/tutorial/part4-docsqa.md
```diff
@@ -108,3 +109,62 @@ The `<GetChunk>` component is used to format each document chunk as it is presen
...
+        chunkFormatter={GetChunk}
+      />
+      {'\n\n'}
+      <DocsQA question="How can I contribute to AI.JSX?" corpus={corpus} chunkLimit={4} chunkFormatter={GetChunk} />
+    </>
+  );
+}
+```
+
+<!-- Alternatively, [Fixie](https://www.fixie.ai) also provides a fully-managed Corpus solution you could drop in instead. -->
```
farzadab, line 170: @mdepinet we maywant to add the Fixie Corpus example here too.

packages/docs/docs/tutorial/part4-docsqa.md
```diff
@@ -108,3 +109,62 @@ The `<GetChunk>` component is used to format each document chunk as it is presen
...
+
+  // For this example, we will index the ai-jsx docs.
+  const loader = new DirectoryLoader('../../packages/docs/docs/', {
+    '.md': (path) => new TextLoader(path),
+  });
+
+  const splitter = new RecursiveCharacterTextSplitter({ chunkSize: 1000, chunkOverlap: 100 });
+  const docs = await loader.loadAndSplit(splitter);
+
+  const vectorStore = await PineconeStore.fromDocuments(docs, new OpenAIEmbeddings(), {
```
mdepinet, line 142: Can we avoid doing all the loading in this example and use fromExistingIndex instead?  It's a bit of a distraction for anyone who already has a Pinecone VDB set up. It also doesn't match how you'd typically do this with ai.jsx (though obviously it works also if you're familiar with LangChain).
If you think we need to show loading, could you use LoadableLangchainCorpus instead?

packages/docs/docs/tutorial/part4-docsqa.md
mdepinet: Yep, that'd be great.  I can do that after this is submitted. :)

mdepinet (approved):

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: I would tend to do this to avoid mutating `params`, since someone else in this function could expect `limit` to still be defined.

```suggestion
  const scoredLcDocs = await vectorStore.similaritySearchWithScore(query, k, _.omit(params, 'limit'));
```

Or better yet, explicitly pass the known params we want to pass to pinecone, rather than letting everything through except known bad params:

```suggestion
  const scoredLcDocs = await vectorStore.similaritySearchWithScore(query, k, _.pick(params, 'score_threshold'));
```

packages/docs/docs/tutorial/part4-docsqa.md
```diff
@@ -26,6 +26,7 @@ that are individually processed by the LLM. Each chunk is passed into the LLM,
 and the _text embedding_ representing the chunk is computed. (For more details on text
 embeddings, check out the [OpenAI text embeddings API docs](https://platform.openai.com/docs/guides/embeddings).) The text embedding is a multidimensional vector, calculated by the LLM,
 representing the semantic content of the chunk. A _corpus_ consists of a set of document chunks and their corresponding embedding vectors.
+For a more detailed explanation of how a corpus is prepared and used, see the guide [DocsQA: Grounding Answers with a Source of Truth](https://docs.ai-jsx.com/guides/docsqa#overview).
```
NickHeiner, line 29: Use a relative link to the markdown file in the monorepo, rather than an absolute link to the hosted file. That'll be more robust.

packages/docs/docs/tutorial/part4-docsqa.md
```diff
@@ -108,3 +109,62 @@ The `<GetChunk>` component is used to format each document chunk as it is presen
 to the LLM. In this case, we just return the raw text of the chunk, but if we wanted
 to include some additional metadata, or transformed the chunks in some way before
 processing them, we could have done so.
+
+## Using a Pinecone database
+
+Until now, we have been using an in-memory Corpus, which is good for a demo, but in practice you might want to use a vector database like [Pinecone](https://www.pinecone.io/) or [Chroma](https://www.trychroma.com/) instead.
```
NickHeiner, line 115:
```suggestion
Until now, we have been using an in-memory Corpus, which is good for a demo, but in practice you might want to use a vector database like [Pinecone](https://www.pinecone.io/) or [Chroma](https://www.trychroma.com/) instead. A vector database allows you to scale to much larger datasets.
```

packages/docs/docs/tutorial/part4-docsqa.md
```diff
@@ -108,3 +109,62 @@ The `<GetChunk>` component is used to format each document chunk as it is presen
...
+
+To do so, you can use [LangchainCorpus](../api/classes/batteries_docs.LangChainCorpus) to integrate with any [VectorStore from LangChain.js](https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/) like so:
```
NickHeiner, line 117:
```suggestion
To do so, you can use [LangchainCorpus](../api/classes/batteries_docs.LangChainCorpus) to integrate with any [VectorStore from LangChain.js](https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/):
```

packages/docs/docs/tutorial/part4-docsqa.md
```diff
@@ -108,3 +109,62 @@ The `<GetChunk>` component is used to format each document chunk as it is presen
...
+  const client = new PineconeClient();
+  await client.init({
+    apiKey: pineconeConfig.apiKey,
+    environment: pineconeConfig.environment,
+  });
+  const pineconeIndex = client.Index(pineconeConfig.index);
+
+  // For this example, we will index the ai-jsx docs.
+  const loader = new DirectoryLoader('../../packages/docs/docs/', {
+    '.md': (path) => new TextLoader(path),
```
NickHeiner, line 136: We should include the `import`s in the example.

packages/tutorial/src/docsqa_pinecone.tsx
```diff
@@ -0,0 +1,109 @@
+/**
+ * This example shows how to use the PineconeStore with a LangChainCorpus for DocsQA.
+ * This is a more advanced version of `docsqa.tsx`, which uses a {@link LocalCorpus} so
```
NickHeiner, line 3:
```suggestion
 * This is a more advanced version of `docsqa.tsx`, which uses a {@link LocalCorpus}, so
```

packages/tutorial/src/docsqa_pinecone.tsx
```diff
@@ -0,0 +1,109 @@
...
+  namespace: 'ai-jsx-tutorial-part4',
+};
+
+function GetChunk({ doc }: { doc: ScoredChunk }) {
+  return doc.chunk.content;
+}
+
+/**
+ * Creates a PineconeStore from the ai-jsx docs.
+ * If the namespace exists however, it will not re-index the documents.
```
NickHeiner, line 40:
```suggestion
 * If the namespace exists, it will not re-index the documents.
```

I don't totally follow what this means / what the implications of this are.

packages/tutorial/src/docsqa_pinecone.tsx
```diff
@@ -0,0 +1,109 @@
...
+    const docs = await loader.loadAndSplit(splitter);
+
+    const vectorStore: VectorStore = await PineconeStore.fromDocuments(docs, embedder, {
+      pineconeIndex,
+      namespace: pineconeConfig.namespace,
+    });
+    return vectorStore;
+  }
+
+  console.log("Namespace exists. Loading it's index ...");
```
NickHeiner, line 84:
```suggestion
  console.log("Namespace exists. Loading its index ...");
```

NickHeiner (approved):

mdwelsh (approved):

packages/tutorial/src/docsqa_pinecone.tsx
farzadab: I did this so that you don't have to delete and re-index all the documents every time. It used to take longer, but now it's honestly not that slow, but still.

packages/docs/docs/tutorial/part4-docsqa.md
farzadab: Oops, nice catch.

packages/docs/docs/tutorial/part4-docsqa.md
farzadab: Good point. That's only for external imports, right?

packages/ai-jsx/src/batteries/docs.tsx
farzadab: Not sure what should and should not be included, so I'm just gonna `omit` the `limit` parameter for now.

packages/docs/docs/tutorial/part4-docsqa.md
NickHeiner: Basically, I want people to be able to paste examples into their editor and have it work. For pieces that are internal and common, like `ChatCompletion`, it's probably generally unnecessary. But yeah, for external imports, it seems good.

PR #159, opened 2023-06-27
Add support for append-only text streaming

petersalas:
This adds underlying support for rendering to an append-only stream.

Components that will yield in an append-only manner can do so by yielding a special symbol, after which point all yielded values will be appended to, rather than replace, previously yielded values. (This indicates to rendering that it is safe to present intermediate content when append-only rendering is requested.)

This adds an API endpoint that demonstrates responding with a text stream that is compatible with Vercel's AI SDK but does not add any client-side demos.

### DIFF
--- a/packages/ai-jsx/src/core/core.ts
+++ b/packages/ai-jsx/src/core/core.ts
@@ -316,20 +316,20 @@ async function* renderStream(
     });
 
     const currentValue = () => {
-      if (!appendOnly) {
-        return inProgressRenders.flatMap((r) => r.currentValue);
-      }
-
-      let currentValue = [] as PartiallyRendered[];
-      for (const inProgressRender of inProgressRenders) {
-        currentValue = currentValue.concat(inProgressRender.currentValue);
-        if (inProgressRender.currentPromise !== null) {
-          // This node is still rendering, so we can't include any ones beyond it.
-          break;
+      if (appendOnly) {
+        let currentValue = [] as PartiallyRendered[];
+        for (const inProgressRender of inProgressRenders) {
+          currentValue = currentValue.concat(inProgressRender.currentValue);
+          if (inProgressRender.currentPromise !== null) {
+            // This node is still rendering, so we can't include any ones beyond it.
+            break;
+          }
         }
+
+        return currentValue;
       }
 
-      return currentValue;
+      return inProgressRenders.flatMap((r) => r.currentValue);
     };
     const remainingPromises = () => inProgressRenders.flatMap((r) => (r.currentPromise ? [r.currentPromise] : []));
 
--- a/packages/ai-jsx/src/core/core.ts
+++ b/packages/ai-jsx/src/core/core.ts
@@ -23,6 +23,12 @@ export type Component<P> = (props: P, context: ComponentContext) => Renderable;
  */
 export type Literal = string | number | null | undefined | boolean;
 
+/**
+ * A value that can be yielded by a component to indicate that each yielded value should
+ * be appended to, rather than replace, the previously yielded values.
+ */
+export const AppendOnlyStream = Symbol('AI.appendOnlyStream');
+
 const attachedContext = Symbol('AI.attachedContext');
 /**
  * An Element represents an instance of an AI.JSX component, with an associated tag, properties, and a render function.
@@ -55,7 +61,10 @@ export type Node = Element<any> | Literal | Node[] | IndirectNode;
  * A RenderableStream represents an async iterable that yields {@link Renderable}s.
  */
 export interface RenderableStream {
-  [Symbol.asyncIterator]: () => AsyncIterator<Renderable, Renderable>;
+  [Symbol.asyncIterator]: () => AsyncIterator<
+    Renderable | typeof AppendOnlyStream,
+    Renderable | typeof AppendOnlyStream
+  >;
 }
 
 /**
@@ -76,7 +85,8 @@ export type PartiallyRendered = string | Element<any>;
 export type StreamRenderer = (
   renderContext: RenderContext,
   renderable: Renderable,
-  shouldStop: ElementPredicate
+  shouldStop: ElementPredicate,
+  appendOnly: boolean
 ) => AsyncGenerator<PartiallyRendered[], PartiallyRendered[]>;
 
 const contextKey = Symbol('AI.contextKey');
@@ -101,6 +111,11 @@ interface RenderOpts<TIntermediate = string, TFinal = string> {
    * @returns The value to yield.
    */
   map?: (frame: TFinal) => TIntermediate;
+
+  /**
+   * Indicates that the stream should be append-only.
+   */
+  appendOnly?: boolean;
 }
 
 /**
@@ -244,12 +259,14 @@ export const LoggerContext = createContext<LogImplementation>(new NoOpLogImpleme
 async function* renderStream(
   context: RenderContext,
   renderable: Renderable,
-  shouldStop: ElementPredicate
+  shouldStop: ElementPredicate,
+  appendOnly: boolean
 ): AsyncGenerator<PartiallyRendered[], PartiallyRendered[]> {
   // If we recurse, propagate the stop function but ensure that intermediate values are preserved.
   const recursiveRenderOpts: RenderOpts<PartiallyRendered[], PartiallyRendered[]> = {
     stop: shouldStop,
     map: (frame) => frame,
+    appendOnly,
   };
 
   if (typeof renderable === 'string') {
@@ -298,7 +315,22 @@ async function* renderStream(
       return inProgressRender;
     });
 
-    const currentValue = () => inProgressRenders.flatMap((r) => r.currentValue);
+    const currentValue = () => {
+      if (!appendOnly) {
+        return inProgressRenders.flatMap((r) => r.currentValue);
+      }
+
+      let currentValue = [] as PartiallyRendered[];
+      for (const inProgressRender of inProgressRenders) {
+        currentValue = currentValue.concat(inProgressRender.currentValue);
+        if (inProgressRender.currentPromise !== null) {
+          // This node is still rendering, so we can't include any ones beyond it.
+          break;
+        }
+      }
+
+      return currentValue;
+    };
     const remainingPromises = () => inProgressRenders.flatMap((r) => (r.currentPromise ? [r.currentPromise] : []));
 
     // Wait for each sub-generator to yield once.
@@ -337,10 +369,7 @@ async function* renderStream(
     try {
       return yield* renderingContext.render(
         renderable.render(renderingContext, new BoundLogger(logImpl, renderId, renderable)),
-        {
-          stop: shouldStop,
-          map: (frame) => frame,
-        }
+        recursiveRenderOpts
       );
     } catch (ex) {
       logImpl.logException(renderable, renderId, ex);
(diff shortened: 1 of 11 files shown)


packages/ai-jsx/src/core/core.ts
```diff
@@ -298,7 +315,22 @@ async function* renderStream(
       return inProgressRender;
     });
 
-    const currentValue = () => inProgressRenders.flatMap((r) => r.currentValue);
+    const currentValue = () => {
+      if (!appendOnly) {
```
NickHeiner, line 319: minor: I think it's generally more readable to have positive assertions, so I'd flip this to:

```tsx
if (appendOnly) { /* ... */ }

return inProgressRenders.flatMap((r) => r.currentValue);
```

packages/ai-jsx/src/core/memoize.tsx
```diff
@@ -109,11 +112,12 @@ export function memo(renderable: Renderable): Node {
           return finalResult;
         } else if (nextPromise == null) {
           nextPromise = generator.next().then((result) => {
+            const memoized = result.value === AI.AppendOnlyStream ? result.value : memo(result.value);
```
NickHeiner, line 115: Is this because it would be incorrect to memoize an append only stream, or because it's unnecessary and thus we should save the bloat from the DebugTree / rest of the system?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -248,13 +248,13 @@ export async function* OpenAICompletionModel(
 
   for await (const event of openAiEventsToJson<CreateCompletionResponse>(responseIterator)) {
     logger.trace({ event }, 'Got createCompletion event');
+    yield event.choices[0].text;
     resultSoFar += event.choices[0].text;
-    yield resultSoFar;
   }
 
   logger.debug({ completion: resultSoFar }, 'Finished createCompletion');
```
NickHeiner, line 255: Hmm – is this log line redundant with the built-in render log for this component? If we remove this, we can also remove `resultsSoFar`, right?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -248,13 +248,13 @@ export async function* OpenAICompletionModel(
 
   for await (const event of openAiEventsToJson<CreateCompletionResponse>(responseIterator)) {
     logger.trace({ event }, 'Got createCompletion event');
+    yield event.choices[0].text;
```
NickHeiner, line 251: It feels nice to remove the `+=` responsibility from this component. And it feels good for components to be able to explicitly say they'll be `appendOnly`, in cases where that can be a simplifying assumption.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -393,7 +393,7 @@ export async function* OpenAIChatModel(
     if (delta.content) {
       currentMessage.content = currentMessage.content ?? '';
       currentMessage.content += delta.content;
-      yield currentMessage.content;
+      yield delta.content;
     }
     if (delta.function_call) {
```
NickHeiner, line 398: Is it weird that we're never `yield`ing the function call?

NickHeiner:
When I ran this locally, I got this error, but it's not clear to me that it's the fault of this PR:

```
yarn workspace ai-jsx build && yarn workspace examples demo:append-only
This example renders an AI.JSX tree in append-only mode, where rendering guarantees that each frame will append to the previous one.
Here's some static text before!

Why is the sky blue? The sky appears blue to the human eye because of a phenomenon called Rayleigh scattering. This occurs when sunlight enters the Earth's atmosphere and collides with air molecules. The shorter blue wavelengths of light scatter more easily in all directions, making the sky look blue to us. This is also why the sky can appear red or orange during sunrises and sunsets, as the longer wavelengths of light are more easily scattered.

Why is Mars red? {"level":50,"time":1687899833531,"pid":7242,"hostname":"Nicks-MBP","name":"ai-jsx","exception":{},"renderId":"c116700c-5617-447c-a40c-2d92d0f091f5","element":"<OpenAIChatModel>","msg":"Rendering element <OpenAIChatModel> failed with exception: TypeError: Cannot read properties of undefined (reading '0')"}
file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/lib/openai.js:245
        finishReason = finishReason ?? deltaMessage.choices[0].finish_reason;
                                                           ^

TypeError: Cannot read properties of undefined (reading '0')
    at OpenAIChatModel (file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/lib/openai.js:245:60)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async renderStream (file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:178:26)
    at async file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:228:34
    at async renderStream (file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:165:20)
    at async file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:228:34
    at async renderStream (file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:165:20)
    at async file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:228:34
    at async renderStream (file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:165:20)
    at async file:///Users/nth/code/ai-jsx/.yarn/__virtual__/ai-jsx-virtual-09e3b1257a/1/packages/ai-jsx/dist/esm/core/core.js:228:34

Node.js v18.16.0
```

This was after hanging for a long time at "Why is Mars red?"

NickHeiner:
I ran a few more times and couldn't repro that issue.

NickHeiner:
Do we consider it to be valid for a generator to emit `appendOnly` midway through its sequence? At that point, would subsequently returned values be appended to the previously-emitted value?

petersalas:
@NickHeiner I've seen that on various branches as well and suspect it's a regression/bug in `OpenAIChatModel`? Will see if I can root cause this afternoon

NickHeiner:
> @NickHeiner I've seen that on various branches as well and suspect it's a regression/bug in `OpenAIChatModel`? Will see if I can root cause this afternoon

Don't worry about it if you have higher-priority things.

packages/ai-jsx/src/stream/index.ts
```diff
@@ -99,6 +99,27 @@ export function toStreamResponse(renderable: Renderable): Response {
   );
 }
 
+/**
+ * Converts a {@link Renderable} to a {@link ReadableStream} that will stream the rendered
+ * content as UTF-8 encoded text.
```
NickHeiner, line 104: If this is user-facing, it would be good to add a sentence about the tradeoffs of doing this versus other approaches, and what those alternatives are.

packages/ai-jsx/src/stream/index.ts
```diff
@@ -99,6 +99,27 @@ export function toStreamResponse(renderable: Renderable): Response {
...
+ */
+export function toTextStream(renderable: Renderable): ReadableStream<Uint8Array> {
+  let previousValue = '';
+  const generator = createRenderContext().render(renderable, { appendOnly: true })[Symbol.asyncIterator]();
+  return new ReadableStream({
+    async pull(controller) {
+      const next = await generator.next();
+      const delta = next.value.slice(previousValue.length);
```
NickHeiner, line 112: Did we consider an alternative where `appendOnly` causes the generator to return deltas directly?

packages/nextjs-demo/src/app/basic-completion/api/text-stream/route.tsx
```diff
@@ -0,0 +1,25 @@
+/** @jsxImportSource ai-jsx */
+import { toTextStream } from 'ai-jsx/stream';
+import { NextRequest } from 'next/server';
+import { ChatCompletion, UserMessage } from 'ai-jsx/core/completion';
+import { StreamingTextResponse } from 'ai';
+
+export async function POST(request: NextRequest) {
+  const { topic } = await request.json();
+
+  return new StreamingTextResponse(
```
NickHeiner, line 10: Is this actually doing anything for us?

NickHeiner (approved):

packages/ai-jsx/src/core/memoize.tsx
petersalas: Just that `memo` takes a `Renderable`, and the symbol isn't itself a `Renderable`.

packages/ai-jsx/src/lib/openai.tsx
petersalas: Nothing (currently anyway) automatically logs the result of rendering, so it's not redundant. (Perhaps they should be logged though.)

packages/nextjs-demo/src/app/basic-completion/api/text-stream/route.tsx
petersalas: Not really, I just wanted to use it to show that it works with their AI SDK. Originally I was just creating a response myself but when I saw that their RSC streamer takes a `ReadableStream` I opted to make that the abstraction instead.

packages/ai-jsx/src/stream/index.ts
petersalas: I've previously tried to go down similar paths and wasn't sure I could make it work in time :)

I have a (shelved, for now) branch that rewrites rendering and will give code that wants it (e.g. this, nested UI/AI streams, partial rendering) much finer-grained visibility/control over rendering so that we can actually propagate fine-grained deltas out of the tree. (Currently the in-progress tree is only implicitly represented on co-routine callstacks which makes things tricky.) Depending on priorities I might take a stab at finishing that when I get back.

packages/ai-jsx/src/core/core.ts
petersalas: I agree

(Though, all things equal I also prefer "short blocks first" especially if I can't indent the mutually exclusive else block BECAUSE OF LINT RULES! 🤪)

packages/ai-jsx/src/lib/openai.tsx
petersalas: It happens to be harmless but I agree that it's weird -- fixed.

packages/ai-jsx/src/lib/openai.tsx
petersalas: Yeah those were always a little icky. Now the ickiness is one layer closer to being removed :)

petersalas:
> Do we consider it to be valid for a generator to emit `appendOnly` midway through its sequence? At that point, would subsequently returned values be appended to the previously-emitted value?

Yeah that's valid. (Whether it's useful or not is a reasonable question but I didn't go out of my way to forbid it.)

packages/ai-jsx/src/stream/index.ts
petersalas: Good call

packages/ai-jsx/src/core/core.ts
NickHeiner: I agree "short blocks first" is also a good heuristic.

Boo hiss upon vacuous `else`. :smile:

PR #160, opened 2023-06-27
Add demo of image rendering

petersalas:
- Define an abstract representation of images that by default renders to its URL
- Add logic at the React boundary to represent images as HTML images

### DIFF
--- a/packages/ai-jsx/src/experimental/next/index.ts
+++ b/packages/ai-jsx/src/experimental/next/index.ts
@@ -6,10 +6,15 @@ import { asJsxBoundary } from '../../react/jsx-boundary.js';
 import { AIJSXError, ErrorCode } from '../../core/errors.js';
 import { toSerializedStreamResponse } from '../../stream/index.js';
 import { ComponentMap } from '../../react/map.js';
+import { Image } from '../../core/image-gen.js';
 export * from '../../react/core.js';
 
 function unwrapReact(partiallyRendered: AI.PartiallyRendered): ReactModule.ReactNode {
   if (AI.isElement(partiallyRendered)) {
+    if (partiallyRendered.tag === Image) {
+      return ReactModule.createElement('img', { src: partiallyRendered.props.url });
+    }
+
     // This should be an AI.React element.
     if (partiallyRendered.tag !== AI.React) {
       throw new AIJSXError(
@@ -73,7 +78,7 @@ export const jsx = asJsxBoundary(function jsx(
   }
 
   const renderResult = AI.createRenderContext(props ?? {}).render(children, {
-    stop: (e) => e.tag === AI.React,
+    stop: (e) => e.tag === AI.React || e.tag === Image,
     map: (frame) => frame.map(unwrapReact),
   });
   const asyncIterator = renderResult[Symbol.asyncIterator]();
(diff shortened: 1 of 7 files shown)


packages/ai-jsx/src/react/core.ts
```diff
@@ -53,6 +54,14 @@ export function createElementSerializer(map: ComponentMap<any>): ElementSerializ
       );
     }
 
+    if (element.tag === Image) {
+      return {
```
petersalas, line 58: It's unfortunate that the current state of the React integration has three different ways AI.JSX elements manifest as React elements (corresponding to architectures 2, 3, and 4) but I think it's overwhelmingly preferable to have additional marginal complexity in places that are already complex (with the _same type of complexity_ IMO!) than spread the complexity elsewhere.

petersalas:
https://ai-jsx-nextjs-demo-git-psalas-images-fixie-ai.vercel.app/recipe

image

@juberti see images via arch 3 ☝️

packages/ai-jsx/src/experimental/next/index.ts
```diff
@@ -6,10 +6,15 @@ import { asJsxBoundary } from '../../react/jsx-boundary.js';
 import { AIJSXError, ErrorCode } from '../../core/errors.js';
 import { toSerializedStreamResponse } from '../../stream/index.js';
 import { ComponentMap } from '../../react/map.js';
+import { Image } from '../../core/image-gen.js';
 export * from '../../react/core.js';
 
 function unwrapReact(partiallyRendered: AI.PartiallyRendered): ReactModule.ReactNode {
   if (AI.isElement(partiallyRendered)) {
+    if (partiallyRendered.tag === Image) {
+      return ReactModule.createElement('img', { src: partiallyRendered.props.url });
```
juberti, line 15: should we be using https://nextjs.org/docs/pages/api-reference/components/image here?

packages/ai-jsx/src/react/core.ts
```diff
@@ -53,6 +54,14 @@ export function createElementSerializer(map: ComponentMap<any>): ElementSerializ
...
+        $$type: 'element',
+        $$component: 'img',
```
juberti, line 60: and here (https://nextjs.org/docs/pages/api-reference/components/image)

packages/ai-jsx/src/react/index.ts
```diff
@@ -5,10 +5,15 @@ import { AIJSXError, ErrorCode } from '../core/errors.js';
 import { Deserialized, fromStreamResponse } from '../stream/index.js';
 import { Jsonifiable } from 'type-fest';
 import { ComponentMap } from './map.js';
+import { Image } from '../core/image-gen.js';
 export * from './core.js';
 
 function unwrapReact(partiallyRendered: AI.PartiallyRendered): ReactModule.ReactNode {
   if (AI.isElement(partiallyRendered)) {
+    if (partiallyRendered.tag === Image) {
+      return ReactModule.createElement('img', { src: partiallyRendered.props.url });
```
juberti, line 14: and here? (https://nextjs.org/docs/pages/api-reference/components/image)

packages/ai-jsx/src/experimental/next/index.ts
petersalas: Here, perhaps. But I don't think for the others, which are either hydrated or running on the client. (But I also don't totally know all that the NextJS Image does.)

juberti: can dalle just return the data as bytes? we can then use a data URI.

farzadab: Tried fixing the issue with the help of Nick, but we still couldn't figure it out. We decide to use the `` tag for now and figure out `Image` later if need be.

NickHeiner:
Closing since the changes were folded into #136.

PR #164, opened 2023-06-28
Wire up <UseTools /> to use OpenAI function calls if available

hessamb:
This wires OpenAI function calls through to <UseTools /> to rely on them for function calling and consuming their responses.

Thanks to @farzadab for walking through the hurdles of typescript on this with me!


farzadab (approved):
LGTM

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -183,6 +190,85 @@ export interface UseToolsProps {
...
+  messages.push(<UserMessage>{props.query}</UserMessage>);
+
+  do {
+    const model_response = (
+      <ChatCompletion
+        functionDefinitions={Object.entries(props.tools).map(([toolName, tool]) => ({
+          name: toolName,
+          description: tool.description,
+          parameters: tool.parameters,
+        }))}
```
NickHeiner, line 227: Should we just map the type of `props.tools` precisely the same as `functionDefinitions`?

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -183,6 +190,85 @@ export interface UseToolsProps {
...
+    if (e.code === ErrorCode.ChatModelDoesntSupportFunctions) {
+      return <UseToolsPromptEngineered {...props} />;
+    }
+    throw e;
+  }
+}
+
+/**
+ * An implementation of <UseTools/> that uses {@link ChatCompletion}'s {@link FunctionDefinition} capability to call
+ * functions. The chat model in scope must support Function Calls for this component.
```
NickHeiner, line 207: Is this a user-facing API? If so, let's link out to the tutorials page / docs content for using tools. If not, let's mark it as `@hidden`.

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -183,6 +190,85 @@ export interface UseToolsProps {
...
+  ];
+  if (props.fallback) {
+    messages.push(
+      <SystemMessage>Here's the fallback strategy/message if something failed: {props.fallback}</SystemMessage>
+    );
+  }
+  messages.push(<UserMessage>{props.query}</UserMessage>);
+
+  do {
+    const model_response = (
```
NickHeiner, line 221:
```suggestion
    const modelResponse = (
```

Please don't make me turn on the lint rule enforcing camelCase. :smile:

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -183,6 +190,85 @@ export interface UseToolsProps {
...
+        {messages}
+      </ChatCompletion>
+    );
+
+    const renderResult = yield* render(model_response, { stop: (el) => el.tag == FunctionCall });
+
+    if (isElement(renderResult[0])) {
+      // Model has generated a <FunctionCall/> element.
+      if (renderResult.length > 1 || renderResult[0].tag != FunctionCall) {
+        throw new AIJSXError(
```
NickHeiner, line 238: You can also use the `metadata` argument to the `AIJSXError` constructor to pass the `renderResult` directly.

packages/ai-jsx/src/core/errors.ts
```diff
@@ -20,6 +20,7 @@ export enum ErrorCode {
   NestedAIUIStreamsAreNotSupported = 1015,
   UnknownUIComponentId = 1016,
   UnknownSerializedComponentType = 1017,
+  ChatModelDoesntSupportFunctions = 1018,
```
NickHeiner, line 23:
```suggestion
  ChatModelDoesNotSupportFunctions = 1018,
```

NickHeiner (approved):
I did a pretty quick skim of this but it seems good!

NickHeiner:
Do we need to update the docs/tutorials in any way to reflect this?

packages/ai-jsx/src/batteries/use-tools.tsx
hessamb: Looks like `metadata` needs to be Jsonifable; whereas `renderResult` is not.

packages/ai-jsx/src/batteries/use-tools.tsx
hessamb: Good callout. I changed functionDefinitions to match with props.tools

PR #165, opened 2023-06-28
Add Anthropic support

NickHeiner:
See the docs changes for more details.


packages/ai-jsx/package.json
```diff
@@ -180,6 +180,16 @@
         "default": "./dist/cjs/lib/openai.cjs"
       }
     },
+    "./lib/anthropic": {
```
mdwelsh, line 183: Dumb question out of curiosity: Why is it necessary to do this for each new source file in the tree? This seems kinda weird to me...

packages/ai-jsx/src/core/errors.ts
```diff
@@ -20,6 +20,10 @@ export enum ErrorCode {
   NestedAIUIStreamsAreNotSupported = 1015,
   UnknownUIComponentId = 1016,
   UnknownSerializedComponentType = 1017,
+  AnthropicDoesNotSupportCompletionModels = 1018,
```
juberti, line 23: This seems likely to get out of hand. Maybe just `CompletionModelNotSupported` that can be used for any provider?

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -0,0 +1,178 @@
...
+type AnthropicModelChoices = ChatOrCompletionModelOrBoth<ValidChatModel, ValidCompletionModel>;
+
+/**
+ * An AI.JSX component that invokes an Anthropic Large Language Model.
+ * @param children The children to render.
+ * @param chatModel The chat model to use.
+ * @param completionModel The completion model to use.
+ * @param client The OpenAI client.
+ */
+export function OpenAI({
```
mdwelsh, line 52: Is this function called the right thing?

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -0,0 +1,178 @@
...
+  })
+);
+
+type ValidCompletionModel = never;
+/**
+ * The set of valid Claude models.
+ *
+ * @see https://docs.anthropic.com/claude/reference/complete_post.
+ */
+type ValidChatModel =
```
juberti, line 30: I'm coming to the conclusion that this is not a good pattern. I don't think we want to have to update our SDK every time some vendor releases a new model.

packages/ai-jsx/src/lib/anthropic.tsx
juberti, line 52: OpenAI?

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -0,0 +1,178 @@
...
+  });
+  yield AI.AppendOnlyStream;
+  const messages = await Promise.all(
+    messageElements.filter(AI.isElement).map(async (message) => {
+      switch (message.tag) {
+        case UserMessage:
+          return `${Anthropic.HUMAN_PROMPT}: ${await render(message)}`;
+        case AssistantMessage:
+          return `${Anthropic.AI_PROMPT}: ${await render(message)}`;
+        case SystemMessage:
```
mdwelsh, line 109: Man, that's painful. Is there any way we can have an option so that the System Message is appended to the AI_PROMPT or something, so people swapping models don't need to update their code?

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -0,0 +1,178 @@
...
+          throw new AIJSXError(
+            'Anthropic models do not support SystemMessage. Change your user message to instruct the model what to do.',
```
juberti, line 111: should we just jam this in as a UserMessage then?

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -0,0 +1,178 @@
...
+            ErrorCode.ChatCompletionUnexpectedChild,
+            'internal'
+          );
+      }
+    })
+  );
+
+  if (!messages.length) {
+    throw new AIJSXError(
+      "ChatCompletion must have at least child that's a SystemMessage, UserMessage, AssistantMessage, FunctionCall, or FunctionResponse, but no such children were found.",
```
mdwelsh, line 134: "at least a child".

Also suggest updating error to make it clear that this only supports UserMessage and AssistantMessage.

juberti, line 134: "one child"? Also, this is kind of confusing when the model doesn't support some of those types.

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -0,0 +1,178 @@
...
+      'user'
+    );
+  }
+
+  messages.push(Anthropic.AI_PROMPT);
+
+  const anthropic = getContext(anthropicClientContext);
+  const anthropicCompletionRequest: Anthropic.CompletionCreateParams = {
+    prompt: messages.join('\n\n'),
+    max_tokens_to_sample: props.maxTokens ?? 1000,
```
mdwelsh, line 145: Does it make sense to use a named constant instead of hardcoding 1000 here?

packages/ai-jsx/src/lib/model.tsx
```diff
@@ -0,0 +1,3 @@
+export type ChatOrCompletionModelOrBoth<ValidChatModel extends string, ValidCompletionModel extends string> =
```
mdwelsh, line 1: Can you add a doc string to this?

mdwelsh (approved):

packages/ai-jsx/package.json
NickHeiner: It's not every new source file; it's just the source files which are exported. Which is a lot right now. 🙃 

If we moved to a model where we had separate npm packages (e.g. `npm install ai-jsx @ai-jsx/tools @ai-jsx/router`), then we'd have less need to do this. But the monopackage pushes us more in this direction.

I agree it's not great and in the future we may change it.

LC does something similar fwiw https://github.com/hwchase17/langchainjs/blob/main/langchain/package.json.

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: NOPE. Good catch.

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: > people swapping models don't need to update their code

That's definitely a good goal. My initial thought was that I wasn't sure we could always do that 100% reliably, so it would be better to give users control rather than try to be magical and fail.

But maybe that would be a good follow-up PR for @hessamb or @farzadab. Maybe it's actually pretty straightforward to do "system message via prompt engineering".

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: 💯

packages/ai-jsx/src/core/errors.ts
NickHeiner: Good question. With the error codes already here, I had to choose between:

1. Error codes reflect semantic situations and may be emitted from multiple callsites.
2. Error codes reflect specific callsites.

Ultimately, I went with (2), because I thought that would make it easier to track down what was happening with bug reports. But it does have the downside you're identifying here. We very well may feel that this gets out of hand and reshuffle down the line. Or maybe we should do it now.

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: 🤦

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: I agree that's a pain, but the type completion is realllly nice. :smile:

The types are just a static check that can be bypassed, so if someone wanted to use a model that we didn't have in this list, they could just do:

```tsx
// @ts-expect-error
model='gpt-4-5-turbo'
```

which is gross, but at least it's not a hard blocker.

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: https://github.com/fixie-ai/ai-jsx/pull/165#discussion_r1246036009

PR #168, opened 2023-06-29
Observability + make streaming demo look better

NickHeiner:


packages/ai-jsx/src/core/core.ts
```diff
@@ -367,10 +367,13 @@ async function* renderStream(
...
     const renderId = uuidv4();
     try {
-      return yield* renderingContext.render(
+      logImpl.log('debug', renderable, renderId, 'Start element');
+      const finalResult = yield* renderingContext.render(
         renderable.render(renderingContext, new BoundLogger(logImpl, renderId, renderable)),
         recursiveRenderOpts
       );
+      logImpl.log('debug', renderable, renderId, { finalResult }, 'Rendered element');
+      return finalResult;
```
NickHeiner: @petersalas lmk if there's a reason this is not the right way to do it.

packages/ai-jsx/src/core/core.ts
```diff
@@ -385,6 +388,8 @@ async function* renderStream(
     while (true) {
       const next = await iterator.next();
       if (next.value === AppendOnlyStream) {
+        // TODO: I'd like to emit a log here indicating that an element has chosen to AppendOnlyStream,
+        // but I'm not sure what the best way is to know which element/renderId produced `renderable`.
```
NickHeiner, line 392: @petersalas lmk if you have suggestions for this.

packages/ai-jsx/src/core/core.ts
```diff
@@ -428,7 +433,7 @@ export function createRenderContext(opts?: { logger?: LogImplementation }) {
   });
 }
 
-function createRenderContextInternal(render: StreamRenderer, userContext: Record<symbol, any>): RenderContext {
+function createRenderContextInternal(renderStream: StreamRenderer, userContext: Record<symbol, any>): RenderContext {
```
NickHeiner, line 436: This code is pretty tough to parse because the word `render` is overused. I changed this param name to match the name of the value that's passed into it, but there's more we can do here.

packages/ai-jsx/src/experimental/next/index.ts
```diff
@@ -160,8 +160,12 @@ export const jsx = asJsxBoundary(function jsx(
...
 export const JSX = jsx;
 
-export function toReactStream(componentMap: ComponentMap<any>, renderable: AI.Renderable): Response {
-  const renderResult = AI.createRenderContext().render(renderable, {
+export function toReactStream(
+  componentMap: ComponentMap<any>,
+  renderable: AI.Renderable,
+  renderContextOpts?: Parameters<typeof AI.createRenderContext>[0]
+): Response {
+  const renderResult = AI.createRenderContext(renderContextOpts).render(renderable, {
```
NickHeiner, line 168: Whoops! `toReactStream` had no way to pass render opts.

packages/docs/docs/guides/rules-of-jsx.md
```diff
@@ -50,14 +49,48 @@ async function CustomerData() {
...
+  // highlight-next-line
+  yield '';
+
+  await new Promise((resolve) => setTimeout(resolve, 4000));
+  return 'slow result';
 }
 ```
 
+This is not ideal and we plan to improve it in the future.
+:::
```
NickHeiner, line 92: I know I used to know this, but I forgot it, and it took me some time today to re-learn. :smile:

packages/nextjs-demo/src/components/Recipe.tsx
```diff
@@ -82,5 +83,12 @@ export function RecipeGenerator({ topic }: { topic: string }) {
     });
   }, [topic]);
 
-  return <div className="whitespace-pre-line">{isLoading ? 'Loading...' : current}</div>;
```
NickHeiner, line 85: lol, this line also killed any sort of incremental streaming, because nothing would be shown until `isLoading` was `false`.

packages/docs/docs/guides/rules-of-jsx.md
petersalas: Two alternative approaches that come to mind are:
1. A `<Suspense>` analogue that lets component consumers declare what should be shown before an async component starts to render
2. Have the runtime impose that _any_ asynchronous component will have an initial `""` frame

The reason I don't love the latter is that components no longer have full control over their presentation.

NickHeiner: I get that for the latter, but concretely, I'm having a hard time coming up with an example of when it'd actually be problematic.

PR #172, opened 2023-07-04
Streaming JIT UI

farzadab:
This PR:
- Adds schemas to JSON completion;
- Enables streaming JIT UI.

### What is UI streaming?
We had _streaming for text_, but not when you generate a UI (so it seems slow to load). I guess you can simply compare [Nick's video](https://fixie-ai.slack.com/archives/C05B6EE9JE5/p1688140097512069) (without streaming) and [mine](https://www.loom.com/share/383e49e3ce1e49aca9530cb7b1dbff70) (with streaming) from Friday.

### Is it a good idea?
Not sure. People had reservations about it in the meeting, but the response to the video seemed positive so I rolled with it. I won't merge it before we get more positive feedback though.

### How does it work?
The idea is that as soon as we have a "valid" prefix of the output, we try to generate & show the UI. With YAML it makes a lot of sense. With JSON, it can only work if you can somehow **"heal" or "untruncate"** or in anyway fix the end. Fortunately there are [JS libraries for that](https://github.com/dphilipson/untruncate-json) (I tried another one and it wasn't as good)!

### Yaml vs JSON:
- Minified JSON: not gonna use it, because apparently the response is less reliable. I think the extra tabs/newlines are actually helping the model quite a bit. Same is true of minified YAML.
- YAML: it's basically the same as JSON, because the model most of the time does return a JSON (as JSON seems to be accepted as YAML). At that point the difference is negligible.
   - Also, with YAML I kept seeing the word `name` at the end of the responses. This is because it thought the start of a new component (which starts with `name:` is a string itself). This doesn't happen with JSON purely out of luck as `":` seems to be a single token and we never get `"name"` in the output :sweat_smile:
- Also, with JSON (and minified json), the `untruncate-json` does a good job of healing a cut-off JSON, so if they're not that different let's keep JSON (I found no similar package for YAML)
- Another reason: JSON is compatible with FunctionCall, so we can easily use that if we decide it's more reliable than prompt engineering.

TL/DR for formats: unfortunately minified versions aren't as reliable. JSON with `untruncate-json` is better than YAML. Also YAML sometimes returns a small incorrect part at the end (`name`) that JSON luckily avoids.

### DIFF
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -171,10 +171,10 @@ async function* OneShotObjectCompletion(
       for (const validator of validatorsAndSchema) {
         validator(object);
       }
-    } catch (e) {
+    } catch (e: any) {
       if (partial.done) {
         logger.warn(
-          { output: partial.value, cleaned: partialResultCleaner ? str : undefined },
+          { output: partial.value, cleaned: partialResultCleaner ? str : undefined, errorMessage: e.message },
           "ObjectCompletion failed. The final result either didn't parse or didn't validate."
         );
         throw e;
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -22,7 +22,7 @@ export type ObjectCompletion = ModelPropsWithChildren & {
   /** Validators are used to ensure that the final object looks as expected. */
   validators?: ((obj: object) => void)[];
   /**
-   * The object Schema that is required. This is a type of validator with the excpetion
+   * The object Schema that is required. This is a type of validator with the exception
    * that the schema description is also provided to the model.
    */
   schema?: z.Schema;
@@ -42,7 +42,7 @@ export type TypedObjectCompletion = ObjectCompletion & {
   /**
    * The intermediate results that get yielded might be incomplete and as such
    * a cleaning/healing/untruncating step might be required.
-   * For example, see the `untrunctate-json` package.
+   * For example, see the `untruncate-json` package.
    */
   partialResultCleaner?: (str: string) => string;
 };
@@ -241,7 +241,7 @@ async function* ObjectCompletionWithRetry(
     }
 
     logger.debug(
-      { atempt: retryIndex + 1, expectedFormat: props.typeName, output },
+      { attempt: retryIndex + 1, expectedFormat: props.typeName, output },
       `Output did not validate to ${props.typeName}.`
     );
   }
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -5,39 +5,90 @@
  */
 
 import * as AI from '../index.js';
-import { ChatCompletion, SystemMessage, AssistantMessage, UserMessage } from '../core/completion.js';
+import {
+  ChatCompletion,
+  SystemMessage,
+  AssistantMessage,
+  UserMessage,
+  ModelPropsWithChildren,
+} from '../core/completion.js';
 import yaml from 'js-yaml';
 import { AIJSXError, ErrorCode } from '../core/errors.js';
+import z from 'zod';
+import { zodToJsonSchema } from 'zod-to-json-schema';
+import untruncateJson from 'untruncate-json';
 
-interface ValidationResult {
-  success: boolean;
-  error: string;
-}
+export type ObjectCompletion = ModelPropsWithChildren & {
+  /** Validators are used to ensure that the final object looks as expected. */
+  validators?: ((obj: object) => void)[];
+  /**
+   * The object Schema that is required. This is a type of validator with the excpetion
+   * that the schema description is also provided to the model.
+   */
+  schema?: z.Schema;
+  /** Any output example to be shown to the model. */
+  example?: string;
+  // TODO (@farzad): better name/framing for example.
+};
 
-// TODO: schema
+export type TypedObjectCompletion = ObjectCompletion & {
+  /** Human-readable name of the type, e.g. JSON or YAML. */
+  typeName: string;
+  /**
+   * Object parser: creates the object to be evaluated given the string.
+   * Note: the parser is only used for validation. The final output is still a string.
+   */
+  parser: (str: string) => object;
+  /**
+   * The intermediate results that get yielded might be incomplete and as such
+   * a cleaning/healing/untruncating step might be required.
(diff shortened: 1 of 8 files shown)


packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -5,169 +5,255 @@
...
+  props: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
+  { render }: AI.ComponentContext
+) {
+  return yield* render(
+    <ObjectCOmpletionWithRetry
+      {...props}
+      typeName="JSON"
+      parser={JSON.parse}
+      // TODO: can we remove .default?
+      partialResultCleaner={untruncateJson.default}
```
farzadab, line 90: Not sure why I had to do this. I tried different variations of the `import` but couldn't make it work right.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -5,169 +5,255 @@
...
-  error: string;
-}
+export type ObjectCompletion = ModelPropsWithChildren & {
+  /** Validators are used to ensure that the final object looks as expected. */
+  validators?: ((obj: object) => void)[];
+  /**
+   * The object Schema that is required. This is a type of validator with the excpetion
+   * that the schema description is also provided to the model.
+   */
+  schema?: z.Schema;
```
farzadab, line 28: Seems like the doc strings are not showing up in the docs site. Should I be using `interface` instead? But `interface` didn't let me easily do `&` and other stuff. I didn't try too much either tbh :)

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -5,169 +5,255 @@
...
+          : ''}
+        {example ? `For example: \n${example}` : ''}
+        Respond with only the {typeName} object. Do not include any explanatory prose. Do not include ```
+        {typeName.toLowerCase()} ``` code blocks.
       </SystemMessage>
     </ChatCompletion>
   );
+  const renderGenerator = render(childrenWithCompletion)[Symbol.asyncIterator]();
+
+  // TODO: we can possibly add a timelimit here so we don't emit too many times.
```
farzadab, line 164: I'm worried that calling parser, schema, and all the validators for each returned result can be a bit too much. We can potentially add a time limit for it too.

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -100,5 +80,35 @@ export async function* UICompletion(
...
-  return <AI.React>{toComponent(JSON.parse(modelResult))}</AI.React>;
+  const Element: z.Schema = z.object({
+    name: z.string(),
+    children: z.array(z.union([z.string(), z.lazy(() => Element)])),
+  });
+
+  const modelRenderGenerator = render(
+    <JsonChatCompletion
+      schema={Element}
+      example={'<SomeComponent /> becomes: { "name": "SomeComponent", "children": [] }'}
```
farzadab, line 91: The `example` prop does seem a bit wishy-washy. Gotta think about a better name/framing for it.

zkoch:
Hey @farzadab any way you could produce a quick screen recording showing what this enables? I thought we already had streaming UI, but maybe I'm thinking about something different.

farzadab:
Good point. I updated the PR description to explain that.

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -100,5 +80,35 @@ export async function* UICompletion(
     return <Component>{children.map(toComponent)}</Component>;
   }
 
-  return <AI.React>{toComponent(JSON.parse(modelResult))}</AI.React>;
+  const Element: z.Schema = z.object({
+    name: z.string(),
```
farzadab, line 84: I tried to restrict the `name` only to `reactComponents`, but my knowledge of `zod` failed me. Any thoughts @NickHeiner? Is it possible to do this if the list is created dynamically?

farzadab, line 84: I tried to restrict the list of `names` to `validComponents` only, but my knowledge of `zod` is failing me. @NickHeiner is it possible to create a union of literals based on a dynamic array in `zod`? I can add a `.refine`, but the model won't see it in the first pass.

zkoch:
> Not sure. People had reservations about it in the meeting, but the response to the video seemed positive so I rolled with it. I won't merge it before we get more positive feedback though.

What were the reservations? This seems like something we should definitely merge so people can start trying it for real

farzadab:
Okay let's YOLO this since people are not here and we can always fix any issues in the follow up PRs.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: Ugh. ESM stuff can be a huge pain.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: > Seems like the doc strings are not showing up in the docs site.

You mean it's not showing up in the generated markdown docs? Hmm. It seems like it should, based on what you're doing.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: Unless I'm misunderstanding what those functions do, I strongly suspect their overhead will be trivial relative to the model latency.

packages/ai-jsx/src/react/completion.tsx
NickHeiner: This seems like it should work, although it gives me a type error:

```ts
  const names = ['foo', 'bar'];
  const u = z.union([names.map(name => z.literal(name))]);
```

I suspect the type error is noise can can be ignored.

packages/ai-jsx/src/batteries/constrained-output.tsx
farzadab: Yeah I think so too.

packages/ai-jsx/src/react/completion.tsx
farzadab: I think I tried it and it was not happy that the list was dynamic. In this code we know what `names` is at compile time, but in my case we don't. AFAIK Zod also wants to support compile time checks somehow (is that what `z.infer` does?) and because of that it only supports static types/validations. I can of course be wrong.

NickHeiner: No that could be right. I vaguely remember trying to do this earlier and it not working. :smile:

PR #175, opened 2023-07-05
JsonCompletion with FunctionCall

farzadab:
This PR creates a version of `JsonChatCompletion` that uses/abuses OpenAI's `FunctionCall`s.
It also expands the definition of a `FunctionParameter` to accept both
the existing simplified version as well as a full-on `z.ZodObject` Schema that
enables more complex scenarios (e.g. recursive types).

### DIFF
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -273,7 +273,7 @@ async function* ObjectCompletionWithRetry(
  * It (ab)uses OpenAI function calls to generate the JSON string.
  *
  * @returns A string that is a valid JSON or throws an error.
- * 
+ *
  * @hidden
  */
 async function* JsonChatCompletionFunctionCall(
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -302,7 +302,6 @@ export async function JsonChatCompletionFunctionCall(
 
   const lastYield = output[output.length - 1];
   if (AI.isElement(lastYield) && lastYield.tag == FunctionCall) {
-    console.log('args', lastYield.props.args);
     // TODO: is validation necessary? If yes, how to handle errors?
     schema.parse(lastYield.props.args);
     return JSON.stringify(lastYield.props.args);
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -10,7 +10,6 @@ import {
   SystemMessage,
   AssistantMessage,
   UserMessage,
-  FunctionCall,
   ModelPropsWithChildren,
 } from '../core/completion.js';
 import yaml from 'js-yaml';
@@ -25,8 +24,11 @@ export type ObjectCompletion = ModelPropsWithChildren & {
   /**
    * The object Schema that is required. This is a type of validator with the exception
    * that the schema description is also provided to the model.
+   *
+   * @note To match OpenAI function definition specs, the schema must be a Zod object.
+   * Arrays and other types should be wrapped in a top-level object in order to be used.
    */
-  schema?: z.Schema;
+  schema?: z.ZodObject<any>;
   /** Any output example to be shown to the model. */
   example?: string;
   // TODO (@farzad): better name/framing for example.
@@ -66,8 +68,12 @@ export type TypedObjectCompletionWithRetry = TypedObjectCompletion & { retries?:
  *   })
  * );
  *
+ * const RootFamilyTree: z.ZodObject<any> = z.object({
+ *   tree: FamilyTree,
+ * });
+ *
  * return (
- *    <JsonChatCompletion schema={FamilyTree}>
+ *    <JsonChatCompletion schema={RootFamilyTree}>
  *     <UserMessage>
  *      Create a nested family tree with names and ages.
  *      It should include a total of 5 people.
@@ -117,9 +123,12 @@ export async function* JsonChatCompletion(
  *     children: z.lazy(() => FamilyTree).optional(),
  *   })
  * );
+ * const RootFamilyTree: z.ZodObject<any> = z.object({
+ *   tree: FamilyTree,
+ * });
  *
  * return (
- *    <YamlChatCompletion schema={FamilyTree}>
+ *    <YamlChatCompletion schema={RootFamilyTree}>
  *     <UserMessage>
  *      Create a nested family tree with names and ages.
  *      It should include a total of 5 people.
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -84,7 +84,6 @@ export async function* JsonChatCompletion(
 ) {
   if (schema) {
     try {
-      // TODO: do not merge with this. This screws up progressive loading.
       return yield* render(<JsonChatCompletionFunctionCall schema={schema} {...props} />);
     } catch (e: any) {
       if (e.code !== ErrorCode.ChatModelDoesNotSupportFunctions) {
@@ -275,12 +274,13 @@ async function* ObjectCompletionWithRetry(
  *
  * @returns A string that is a valid JSON or throws an error.
  */
-export async function JsonChatCompletionFunctionCall(
+export async function* JsonChatCompletionFunctionCall(
   { schema, children, ...props }: ModelPropsWithChildren & { schema: z.Schema },
   { render }: AI.ComponentContext
 ) {
   const childrenWithCompletion = (
     <ChatCompletion
+      experimental_streamFunctionCallOnly
       {...props}
       functionDefinitions={{
         print: {
@@ -291,23 +291,15 @@ export async function JsonChatCompletionFunctionCall(
     >
       {children}
       <SystemMessage>
-        Your response must use the `print` function that is provided. No other explanation needed.
+        Your response must use the `print` function that is provided. No other explanation needed. do not respond with
+        an assistant message. Just call the function.
       </SystemMessage>
     </ChatCompletion>
   );
 
-  const output = await render(childrenWithCompletion, { stop: (e) => e.tag == FunctionCall });
-
-  // TODO: can we make this work with progressive loading? This will require changes to core/completion.
-
-  const lastYield = output[output.length - 1];
-  if (AI.isElement(lastYield) && lastYield.tag == FunctionCall) {
-    // TODO: is validation necessary? If yes, how to handle errors?
-    schema.parse(lastYield.props.args);
-    return JSON.stringify(lastYield.props.args);
+  const frames = render(childrenWithCompletion);
+  for await (const frame of frames) {
+    yield JSON.stringify(JSON.parse(frame).arguments);
   }
-  // TODO: retry maybe?
-  throw new AIJSXError('Expected a function call to `print`.', ErrorCode.ModelOutputDidNotMatchConstraint, 'runtime', {
-    output: JSON.stringify(output),
-  });
+  return JSON.stringify(JSON.parse(await frames).arguments);
 }
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -79,9 +79,19 @@ export type TypedObjectCompletionWithRetry = TypedObjectCompletion & { retries?:
  *    Intermediate results that are valid, are also yielded.
  */
 export async function* JsonChatCompletion(
-  props: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
+  { schema, ...props }: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
   { render }: AI.ComponentContext
 ) {
+  if (schema) {
+    try {
+      // TODO: do not merge with this. This screws up progressive loading.
+      return yield* render(<JsonChatCompletionFunctionCall schema={schema} {...props} />);
+    } catch (e: any) {
+      if (e.code !== ErrorCode.ChatModelDoesNotSupportFunctions) {
+        throw e;
+      }
+    }
+  }
   return yield* render(
     <ObjectCompletionWithRetry
       {...props}
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -67,7 +67,7 @@ export type TypedObjectCompletionWithRetry = TypedObjectCompletion & { retries?:
  * );
  *
  * return (
- *    <JsonChatCompletion>
+ *    <JsonChatCompletion schema={FamilyTree}>
  *     <UserMessage>
  *      Create a nested family tree with names and ages.
  *      It should include a total of 5 people.
@@ -119,7 +119,7 @@ export async function* JsonChatCompletion(
  * );
  *
  * return (
- *    <YamlChatCompletion>
+ *    <YamlChatCompletion schema={FamilyTree}>
  *     <UserMessage>
  *      Create a nested family tree with names and ages.
  *      It should include a total of 5 people.
@@ -171,7 +171,6 @@ async function* OneShotObjectCompletion(
   );
   const renderGenerator = render(childrenWithCompletion)[Symbol.asyncIterator]();
 
-  // TODO: we can possibly add a timelimit here so we don't emit too many times.
   let lastYieldedLen = 0;
   while (true) {
     const partial = await renderGenerator.next();
@@ -276,7 +275,7 @@ async function* ObjectCompletionWithRetry(
  *
  * @hidden
  */
-async function* JsonChatCompletionFunctionCall(
+export async function* JsonChatCompletionFunctionCall(
   { schema, children, ...props }: ModelPropsWithChildren & { schema: z.Schema },
   { render }: AI.ComponentContext
 ) {
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -13,10 +13,11 @@ import {
   ModelPropsWithChildren,
 } from '../core/completion.js';
 import yaml from 'js-yaml';
-import { AIJSXError, ErrorCode } from '../core/errors.js';
+import { AIJSXError, ErrorCode, ErrorBlame } from '../core/errors.js';
+import { Jsonifiable } from 'type-fest';
 import z from 'zod';
 import { zodToJsonSchema } from 'zod-to-json-schema';
-import untruncateJson from 'untruncate-json';
+import { patchedUntruncateJson } from '../lib/util.js';
 
 export type ObjectCompletion = ModelPropsWithChildren & {
   /** Validators are used to ensure that the final object looks as expected. */
@@ -27,6 +28,12 @@ export type ObjectCompletion = ModelPropsWithChildren & {
    *
    * @note To match OpenAI function definition specs, the schema must be a Zod object.
    * Arrays and other types should be wrapped in a top-level object in order to be used.
+   *
+   * For example, to describe a list of strings, the following is not accepted:
+   * `const schema: z.Schema = z.array(z.string())`
+   *
+   * Instead, you can wrap it in an object like so:
+   * `const schema: z.ZodObject = z.object({ arr: z.array(z.string()) })`
    */
   schema?: z.ZodObject<any>;
   /** Any output example to be shown to the model. */
@@ -88,13 +95,11 @@ export async function* JsonChatCompletion(
   { schema, ...props }: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
   { render }: AI.ComponentContext
 ) {
-  if (schema) {
-    try {
-      return yield* render(<JsonChatCompletionFunctionCall schema={schema} {...props} />);
-    } catch (e: any) {
-      if (e.code !== ErrorCode.ChatModelDoesNotSupportFunctions) {
-        throw e;
-      }
+  try {
+    return yield* render(<JsonChatCompletionFunctionCall schema={schema ?? z.object({}).nonstrict()} {...props} />);
+  } catch (e: any) {
+    if (e.code !== ErrorCode.ChatModelDoesNotSupportFunctions) {
+      throw e;
     }
   }
   return yield* render(
@@ -102,8 +107,7 @@ export async function* JsonChatCompletion(
       {...props}
       typeName="JSON"
       parser={JSON.parse}
-      // TODO: can we remove .default?
-      partialResultCleaner={untruncateJson.default}
+      partialResultCleaner={patchedUntruncateJson}
     />
   );
 }
@@ -148,6 +152,16 @@ export async function* YamlChatCompletion(
   );
 }
 
+export class CompletionError extends AIJSXError {
+  constructor(
+    message: string,
+    public readonly blame: ErrorBlame,
+    public readonly metadata: Jsonifiable & { output: string; validationError: string }
+  ) {
+    super(message, ErrorCode.ModelOutputDidNotMatchConstraint, blame, metadata);
+  }
+}
+
 /**
  * A {@link ChatCompletion} component that constrains the output to be a valid object format (e.g. JSON/YAML).
  *
@@ -159,7 +173,7 @@ export async function* YamlChatCompletion(
  */
 async function* OneShotObjectCompletion(
   { children, typeName, validators, example, schema, parser, partialResultCleaner, ...props }: TypedObjectCompletion,
-  { render, logger }: AI.ComponentContext
+  { render }: AI.ComponentContext
 ) {
   // If a schema is provided, it is added to the list of validators as well as the prompt.
   const validatorsAndSchema = schema ? [schema.parse, ...(validators ?? [])] : validators ?? [];
@@ -170,9 +184,9 @@ async function* OneShotObjectCompletion(
       <SystemMessage>
         Respond with a {typeName} object that encodes your response.
         {schema
-          ? `The ${typeName} object should match this JSON Schema: ${JSON.stringify(zodToJsonSchema(schema))}`
+          ? `The ${typeName} object should match this JSON Schema: ${JSON.stringify(zodToJsonSchema(schema))}\n`
           : ''}
-        {example ? `For example: \n${example}` : ''}
+        {example ? `For example: ${example}\n` : ''}
         Respond with only the {typeName} object. Do not include any explanatory prose. Do not include ```
         {typeName.toLowerCase()} ``` code blocks.
       </SystemMessage>
@@ -191,11 +205,11 @@ async function* OneShotObjectCompletion(
       }
     } catch (e: any) {
       if (partial.done) {
-        logger.warn(
-          { output: partial.value, cleaned: partialResultCleaner ? str : undefined, errorMessage: e.message },
-          "ObjectCompletion failed. The final result either didn't parse or didn't validate."
-        );
-        throw e;
+        throw new CompletionError(`The model did not produce a valid ${typeName} object`, 'runtime', {
+          typeName,
+          output: partial.value,
+          validationError: e.message,
+        });
       }
       continue;
     }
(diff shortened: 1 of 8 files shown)


packages/ai-jsx/src/core/completion.tsx
```diff
@@ -40,18 +42,50 @@ export type ModelComponent<T extends ModelPropsWithChildren> = Component<T>;
  */
 export interface FunctionDefinition {
   description?: string;
-  parameters: Record<string, FunctionParameter>;
+  parameters: FunctionParameters;
+}
+
+export function getParametersSchema(parameters: FunctionParameters) {
```
farzadab, line 48: We are allowing both the simplified function signature as well as the more complex `zod` Schema.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -78,9 +79,19 @@ export type TypedObjectCompletionWithRetry = TypedObjectCompletion & { retries?:
...
  */
 export async function* JsonChatCompletion(
-  props: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
+  { schema, ...props }: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
   { render }: AI.ComponentContext
 ) {
+  if (schema) {
+    try {
+      // TODO: do not merge with this. This screws up progressive loading.
+      return yield* render(<JsonChatCompletionFunctionCall schema={schema} {...props} />);
```
farzadab, line 88: As the note says, we shouldn't merge until we figure this part out.

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -40,18 +42,50 @@ export type ModelComponent<T extends ModelPropsWithChildren> = Component<T>;
...
+  if (parameters instanceof z.Schema) {
+    const jsonSchema = zodToJsonSchema(parameters);
+    if (!('type' in jsonSchema) || jsonSchema.type !== 'object') {
+      throw new Error('Function parameters must be an object');
+    }
```
farzadab, line 53: @NickHeiner is there a sub-type of `z.Schema` that ensures the top-level to be an object? I tried a few things and asked Perplexity, but I couldn't get it to work.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -355,21 +356,7 @@ export async function* OpenAIChatModel(
...
-            (map: Record<string, any>, paramName) => ({
-              ...map,
-              [paramName]: {
-                type: functionDefinition.parameters[paramName].type,
-              },
-            }),
-            {}
-          ),
-        },
+        parameters: getParametersSchema(functionDefinition.parameters),
```
farzadab, line 359: This is an improvement to `FunctionCall` that both enables this PR and makes FunctionCall stronger.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -5,169 +5,310 @@
...
-    return { success: true, error: '' } as ValidationResult;
-  } catch (e: any) {
-    return { success: false, error: e.message } as ValidationResult;
+  const output = await render(childrenWithCompletion, { stop: (e) => e.tag == FunctionCall });
+
+  // TODO: can we make this work with progressive loading? This will require changes to core/completion.
+
+  const lastYield = output[output.length - 1];
+  if (AI.isElement(lastYield) && lastYield.tag == FunctionCall) {
+    console.log('args', lastYield.props.args);
```
hessamb, line 305: Forgot to debug info?
```suggestion
```

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -5,169 +5,310 @@
...
+    // TODO: is validation necessary? If yes, how to handle errors?
```
hessamb, line 306: maybe similar to how you have `<ObjectCompletionWithRetry>` and `<OneShotObjectCompletion>`, we can have the same separation here and we pass the parsing error as a `<FunctionResponse>` back to the model.

packages/ai-jsx/src/batteries/constrained-output.tsx
hessamb: Does this screw up progressive loading because chat completion doesn't stream partials of a `<FunctionCall />` or is there something else going on here?

packages/ai-jsx/src/batteries/constrained-output.tsx
farzadab: Yup. I'm guessing I'll need more clean up still.

packages/ai-jsx/src/core/completion.tsx
NickHeiner: I think this does what you're looking for.

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBALzgMyhEcDkCIBNMDcAUEQMYQB2AzvAIYUCeA8gEYBWApqTAMqkAWHELTgBeRADooXaLgAUCCQFcKAawoQA7hTkBKXcRgMwHOABUxk4BWQcoAHiMmIyOPWbsuvAUNoA+EnJqeAgALnNLAG9kCDCsFlooTABfYiIAenS4AHVTTXp4GH5gKjgYCBRaYAAbMkoaOAgAJnCLcRgoJQ4CIA

https://www.perplexity.ai/search/23eea325-9d11-45af-b4d2-7daa067cc7dd?s=c

NickHeiner:
heh, great to see the "files changed" drop from 11 to 4. :smile:

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -257,3 +268,46 @@ async function* ObjectCompletionWithRetry(
...
   );
 }
+
+/**
+ * A {@link ChatCompletion} component that constrains the output to be a valid JSON string.
+ * It (ab)uses OpenAI function calls to generate the JSON string.
+ *
+ * @returns A string that is a valid JSON or throws an error.
+ */
+export async function JsonChatCompletionFunctionCall(
```
NickHeiner, line 278: Is this intended to be used publicly? Or are people supposed to use `JsonChatCompletion`, which delegates to this if possible?

farzadab: The idea was to delegate. But open to suggestions.

farzadab: On that note, yes we should probably remove the `export` keyword.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -257,3 +267,39 @@ async function* ObjectCompletionWithRetry(
...
+        an assistant message. Just call the function.
+      </SystemMessage>
+    </ChatCompletion>
+  );
+
+  const frames = render(childrenWithCompletion);
+  for await (const frame of frames) {
+    yield JSON.stringify(JSON.parse(frame).arguments);
+  }
+  return JSON.stringify(JSON.parse(await frames).arguments);
```
NickHeiner, line 304: @petersalas, is this the right pattern to modify the result of an inner component as it's being passed up the render tree? Or am I just supposed to use the `map` arg here?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -420,11 +432,20 @@ export async function* OpenAIChatModel(
...
+        });
+      }
     }
   }
 
   logger.debug({ message: currentMessage }, 'Finished createChatCompletion');
 
+  if (props.experimental_streamFunctionCallOnly) {
+    return JSON.stringify(currentMessage.function_call);
+  }
```
NickHeiner, line 448: @petersalas now that you see what I had in mind, lmk what you think.

packages/examples/src/validated-generation.tsx
```diff
@@ -10,21 +13,51 @@ const FamilyTree: z.Schema = z.array(
...
+
+// let lastValue = '';
+// const rendering = AI.createRenderContext().render(<App />, { appendOnly: true });
+// for await (const frame of rendering) {
+//   process.stdout.write(frame.slice(lastValue.length));
+//   lastValue = frame;
+// }
+
+// const finalResult = await rendering;
+// process.stdout.write(finalResult.slice(lastValue.length));
```
NickHeiner, line 63: We can clean this up before merging.

packages/examples/src/validated-generation.tsx
```diff
@@ -10,21 +13,51 @@ const FamilyTree: z.Schema = z.array(
...
 function App() {
   const query = 'Create a nested family tree with names and ages. It should include a total of 5 people.';
   return (
     <>
       JSON generation example:{'\n'}
-      <JsonChatCompletion schema={FamilyTree}>
-        <UserMessage>{query}</UserMessage>
-      </JsonChatCompletion>
-      {'\n\n'}
+      <ChatProvider model="gpt-4">
```
NickHeiner, line 25: To make this work, I had to switch to a model that supports functions.

packages/examples/src/validated-generation.tsx
```diff
@@ -10,21 +13,51 @@ const FamilyTree: z.Schema = z.array(
...
+  level: process.env.loglevel ?? 'debug',
+  transport: {
+    target: 'pino-pretty',
+    options: {
+      colorize: true,
+    },
+  },
+});
+
 showInspector(<App />);
```
NickHeiner, line 50: When we run the inspector, we see the results streaming.

packages/ai-jsx/src/batteries/constrained-output.tsx
petersalas: Either works

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -297,7 +321,9 @@ export async function* OpenAIChatModel(
       e.tag == FunctionCall ||
       e.tag == FunctionResponse,
   });
-  yield AI.AppendOnlyStream;
+  if (!props.experimental_streamFunctionCallOnly) {
+    yield AI.AppendOnlyStream;
```
petersalas, line 325: Should we skip the `yield delta.content` on line 425 if this is set?

packages/ai-jsx/src/lib/openai.tsx
petersalas: it's about what I expected :)

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -78,9 +79,18 @@ export type TypedObjectCompletionWithRetry = TypedObjectCompletion & { retries?:
  *    Intermediate results that are valid, are also yielded.
  */
 export async function* JsonChatCompletion(
-  props: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
+  { schema, ...props }: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
   { render }: AI.ComponentContext
 ) {
+  if (schema) {
+    try {
+      return yield* render(<JsonChatCompletionFunctionCall schema={schema} {...props} />);
```
petersalas, line 87: feels like there's probably a formulation with `<ErrorBoundary>` here that's a little nicer but I think it amounts to the same thing

packages/examples/src/validated-generation.tsx
petersalas: doesn't gpt 3.5 support function calls?

petersalas: Either way consider `<OpenAI chatModel="...">` instead

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: And, if it's internal only, I was expecting to see a `@hidden` tag, to prevent it from showing up in the generated API docs.

packages/examples/src/validated-generation.tsx
NickHeiner: Ah, looks like it does actually.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: Hmm. I switched to this:

```tsx
  return render(childrenWithCompletion, {
    map: (frame) => JSON.stringify(JSON.parse(frame).arguments)
  });
```

And it doesn't apply the mapping function to the final result. That seems consistent with the other places we call `render`, but it also seems like it'll be a footgun to people with less context.

I then tried a few ways to also act on the final result, but it didn't work – what's the recommended approach here?

packages/ai-jsx/src/lib/openai.tsx
NickHeiner: Yes, good point.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -78,9 +79,18 @@ export type TypedObjectCompletionWithRetry = TypedObjectCompletion & { retries?:
  *    Intermediate results that are valid, are also yielded.
```
NickHeiner, line 79: @farzadab can we update the JSDoc example for `JsonChatCompletion` to include the `schema` prop? Right now the `FamilyTree` var is not used.

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -40,18 +42,56 @@ export type ModelComponent<T extends ModelPropsWithChildren> = Component<T>;
...
-  parameters: Record<string, FunctionParameter>;
+  parameters: FunctionParameters;
+}
+
+export function getParametersSchema(parameters: FunctionParameters) {
+  if (parameters instanceof z.Schema) {
+    const jsonSchema = zodToJsonSchema(parameters);
+    if (!('type' in jsonSchema) || jsonSchema.type !== 'object') {
+      throw new AIJSXError(
+        `Function parameters param must be an object, but was: ${JSON.stringify(jsonSchema, null, 2)}`,
```
NickHeiner, line 53: It's a bit confusing to say "input must be an object" when the input `{a: true}` would trigger this error.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner, line 1: @farzadab in this file, the error that emits `ErrorCode.ModelOutputDidNotMatchConstraint` is missing `validationError` from the metadata.

packages/ai-jsx/src/core/completion.tsx
NickHeiner: I think the current form is a footgun, because things like this don't work:

```tsx
const Elements = z.union([Element, z.array(Element)]);

<JsonChatCompletion schema={Elements} retries={1}>
```

OpenAI requires that the top-level type be an object. However, it seems more natural in Zod to pass in any Zod type.

The way I've been solving this manually is to wrap with an outer object:

```tsx
  const Elements = z.union([Element, z.array(Element)]);
  const elementsWrapper = z.object({root: Elements });
<JsonChatCompletion schema={elementsWrapper} retries={1}>
```

But this is kind of annoying for me and will be confusing for newcomers.

Maybe the `JsonChatCompletion` component should do the wrapping / unwrapping automatically? I'm slightly concerned it would be too magical, but I think on net it's worth eliminating the footgun.

At the very least, we could update the error message to tell the dev exactly what they need to do.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: Perhaps relatedly, something odd is going on here.

When I run `yarn workspace examples demo:validated-generation`, this works fine as-is.

However, when I rebase https://github.com/fixie-ai/ai-jsx/pull/186 to be on top of this branch and run it, the render result emitted from `JsonChatCompletionFunctionCall` looks good, but the render result of `JsonChatCompletion` is a doubly-quoted string: `"\"{a: 123}\""`. This means that when the caller calls `JSON.parse`, it gets a string, not an object.

I'm confused as to why the output of `JCCFC` looks good and the output of `JCC` looks bad, given that this is all `JCC` does:

```
      return yield* render(<JsonChatCompletionFunctionCall schema={schema} {...props} />);
```

packages/ai-jsx/src/core/completion.tsx
farzadab: If we can make it work every time that's good, but sometimes the user might describe the JSON in their prompt (since it doesn't usually follow the schema too well). In that case the difference between user's view and the actual schema can cause confusion.

farzadab: If we do that, we'll have to come up with a generic name for it, but names matter with LLMs and it can be annoying for users who don't want this extra step.

This is kind of an issue with OpenAI, not us, so maybe we don't need to address it? Maybe they'll fix it, or another provider will come up with a better interface. Handling this will put the burden on us to make sure it is the best thing to do.

packages/ai-jsx/src/batteries/constrained-output.tsx
petersalas:
```ts
return map(yield* render(..., { map }));
```

```ts
const iterator = render(...)[Symbol.asyncIterator]();
while (true) {
    const next = await iterator.next();
    const value = map(next.value);
    if (next.done) {
        return value;
    } else {
        yield value;
    }
}
```

```ts
const rendered = render(...);
for await (const frame of rendered) {
   yield map(frame);
}
return map(await rendered);
```

all should be equivalent. (Not super ergonomic, I know 😦)

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -291,23 +293,15 @@ export async function JsonChatCompletionFunctionCall(
...
-    return JSON.stringify(lastYield.props.args);
+  const frames = render(childrenWithCompletion);
+  for await (const frame of frames) {
+    yield JSON.stringify(JSON.parse(frame).arguments);
   }
-  // TODO: retry maybe?
-  throw new AIJSXError('Expected a function call to `print`.', ErrorCode.ModelOutputDidNotMatchConstraint, 'runtime', {
-    output: JSON.stringify(output),
-  });
+  return JSON.stringify(JSON.parse(await frames).arguments);
```
farzadab, line 306: I'm not a fan of how we have so many `JSON.parse` and `JSON.stringify`'s, but this is okay for now.

packages/ai-jsx/src/batteries/constrained-output.tsx
farzadab: There was a `JSON.parse` missing on the arguments. Should get fixed.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: Yeah it's not ideal – feel free to change it if you have a better suggestion. :smile:

packages/ai-jsx/src/core/completion.tsx
NickHeiner: > If we can make it work every time that's good, but sometimes the user might describe the JSON in their prompt (since it doesn't usually follow the schema too well).

Great point about the difference with the prompt itself causing a problem.

But is it still true that the model doesn't follow the schema well with function calls? I thought OpenAI was claiming that it did.

NickHeiner: > This is kind of an issue with OpenAI, not us

I disagree. OpenAI says, "give us any JSON schema that's an object", and that's easy to follow.

We say, "give us any Zod schema", but then at runtime, Zod schemas that aren't objects at the top-level fail. And we can see how counter-intuitive this is, because both you and I wrote Zod schemas that aren't objects at the top level in these PRs. :smile:

But I agree with you that it's bad if the user does their own prompt engineering and then the precise form doesn't match. So for now, let's just make an explicit note in the docs.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -78,9 +79,18 @@ export type TypedObjectCompletionWithRetry = TypedObjectCompletion & { retries?:
  *    Intermediate results that are valid, are also yielded.
  */
 export async function* JsonChatCompletion(
-  props: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
+  { schema, ...props }: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
   { render }: AI.ComponentContext
 ) {
+  if (schema) {
```
NickHeiner, line 85: I think it would actually be nice to always use `JsonChatCompletionFunctionCall`. If no `schema` is provided, that means the user just wants JSON of any form, right? In that case, we can create a very generic Zod schema that matches any JSON (maybe just `zod.object()`? 

https://www.perplexity.ai/search/56487aec-baf4-4590-bd66-c43a713b0127?s=c

packages/ai-jsx/src/core/completion.tsx
farzadab: Oh I agree that we should make it a compile time error and not a runtime one. That's why I asked you before. I'm gonna try your idea here now (sorry got called into a meeting):
https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=3&pc=1#code/JYWwDg9gTgLgBALzgMyhEcDkCIBNMDcAUEQMYQB2AzvAIYUCeA8gEYBWApqTAMqkAWHELTgBeRADooXaLgAUCCQFcKAawoQA7hTkBKXcRgMwHOABUxk4BWQcoAHiMmIyOPWbsuvAUNoA+EnJqeAgALnNLAG9kCDCsFlooTABfYiIAenS4AHVTTXp4GH5gKjgYCBRaYAAbMkoaOAgAJnCLcRgoJQ4CIA

packages/examples/src/validated-generation.tsx
```diff
@@ -24,11 +25,12 @@ function App() {
       </JsonChatCompletion>
       {'\n\n'}
       YAML generation example:{'\n'}
-      <YamlChatCompletion schema={FamilyTree}>
+      <YamlChatCompletion schema={RootFamilyTree}>
```
NickHeiner, line 28: nice

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -437,7 +437,7 @@ export async function* OpenAIChatModel(
       if (props.experimental_streamFunctionCallOnly) {
         yield JSON.stringify({
           ...currentMessage.function_call,
-          arguments: untruncateJson.default(currentMessage.function_call.arguments ?? '{}'),
+          arguments: JSON.parse(untruncateJson.default(currentMessage.function_call.arguments || '{}')),
```
NickHeiner, line 440: ah, was this the problem with my "sometimes it's a multiply-nested string" issue?

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: hehehe yeah that first one is the one I'd reach for in terms of concision but damn would it be hard for an uninformed dev to come up with that on their own.

packages/ai-jsx/src/core/completion.tsx
farzadab: As per discussion, this was made to be more strict by enforcing a `z.ZodObject` instead of a general `z.Schema`.

packages/ai-jsx/src/lib/openai.tsx
farzadab: Yup exactly. Should be good now.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -25,8 +24,11 @@ export type ObjectCompletion = ModelPropsWithChildren & {
   /**
    * The object Schema that is required. This is a type of validator with the exception
    * that the schema description is also provided to the model.
+   *
+   * @note To match OpenAI function definition specs, the schema must be a Zod object.
+   * Arrays and other types should be wrapped in a top-level object in order to be used.
```
NickHeiner, line 29: Can you provide some examples with "correct" and "incorrect" cases here?

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -45,19 +45,17 @@ export interface FunctionDefinition {
   parameters: FunctionParameters;
 }
 
+/**
+ * This function creates a [JSON Schema](https://json-schema.org/) object to describe
+ * parameters for a {@link FunctionDefinition}.
+ * The parameters can be described either using a record of parameter names to
+ * {@link PlainFunctionParameter} objects, or using a {@link z.ZodObject} schema object.
+ *
+ * @note If using a Zod schema, the top-level schema must be an object as per OpenAI specifications.
```
NickHeiner, line 54: Any time we reference external specs like this, it's good to link to them.

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -45,19 +45,17 @@ export interface FunctionDefinition {
...
+ */
 export function getParametersSchema(parameters: FunctionParameters) {
-  if (parameters instanceof z.Schema) {
-    const jsonSchema = zodToJsonSchema(parameters);
-    if (!('type' in jsonSchema) || jsonSchema.type !== 'object') {
-      throw new AIJSXError(
-        `Function parameters param must be an object, but was: ${JSON.stringify(jsonSchema, null, 2)}`,
-        ErrorCode.InvalidParamSchemaType,
```
NickHeiner: Let's remove this from the `ErrorCode` enum if it's no longer used.

farzadab: Q: should we somehow disable that error code now?
I mean: do we care if the error codes from different versions possibly map to different things?

packages/ai-jsx/src/batteries/constrained-output.tsx
farzadab: Good idea. Done.

packages/ai-jsx/src/batteries/constrained-output.tsx
farzadab: Yeah good point. I had some fixes in another PR that I needed to pull in here.

packages/ai-jsx/src/batteries/constrained-output.tsx
farzadab: I assume I can resolve this thread?

packages/ai-jsx/src/batteries/constrained-output.tsx
farzadab: You're right, but since I haven't used `ErrorBoundary` before and I'm not actively playing with this right now (so I won't find the issues), I'd rather stick to this for now.

packages/ai-jsx/src/core/completion.tsx
NickHeiner: I felt good about where we ended up on this, but I ran into another related footgun to this.

I made a component:

```tsx
          <JsonChatCompletion>
            <UserMessage>Generate some information about hotels, including each hotel's name and how many stars it is</UserMessage>
          </JsonChatCompletion>
```

It generated this JSON:

```json
[{"hotel_name":"Hotel A","stars":4},{"hotel_name":"Hotel B","stars":3},{"hotel_name":"Hotel C","stars":5}]
```

This is valid for my usecase, but JCC threw a validation error:

```
    validationError: '[\n' +
      '  {\n' +
      '    "code": "invalid_type",\n' +
      '    "expected": "object",\n' +
      '    "received": "array",\n' +
      '    "path": [],\n' +
      '    "message": "Expected object, received array"\n' +
      '  }\n' +
      ']'
```

I believe this is because if `schema` is not passed, JCC defaults to `object`, so this array fails validation:

https://github.com/fixie-ai/ai-jsx/blob/55ddca6efa8999c8b0632274a811bba82ebc3986/packages/ai-jsx/src/batteries/constrained-output.tsx#L99.

Do you have thoughts as to what a good solution here is?

NickHeiner: As a workaround, I can manually specify a `schema` arg that's an object. But this would be confusing to people.

farzadab: Hmm yeah good point.

Q: do you want JCCFC specifically? If yes, then the top-level has to be an object (unless we do wrapping/unwrapping which I'm against and could add to confusion).

If not, then we can go back to not using JCCFC if the schema is not provided. I agree that this is a bit weird: presence or not of `schema` shouldn't change the underlying mechanism without any warnings.

Maybe the solution is to expose both and let people decide, but I think you were against that. We might also make it more explicit by having an option that says `useFunctionCall = ['true', 'false', 'auto']` (or something like that).

PR #183, opened 2023-07-11
Updated Guide (AI Newcomers)

benlower:
Cleaning up content a bit, typo fixes

### DIFF
--- a/packages/docs/docs/guides/brand-new.md
+++ b/packages/docs/docs/guides/brand-new.md
@@ -139,7 +139,7 @@ More detail: [`UseTools` API docs](../api/modules/batteries_use_tools#usetools).
 
 ### Accessing Knowledge (AKA "Docs QA")
 
-LLMs have "soft knowledge" of the world, but if you just ask a question without providing any context, they're prone to [hallucination](https://en.wikipedia.org/wiki/Hallucination_(artificial_intelligence)). Additionally, because LLMs are only trained on public data, they don't have context on private data your or your app may care about.
+LLMs have "soft knowledge" of the world, but if you just ask a question without providing any context, they're prone to [hallucination](<https://en.wikipedia.org/wiki/Hallucination_(artificial_intelligence)>). Additionally, because LLMs are only trained on public data, they don't have context on private data your or your app may care about.
 
 To address this, the community has developed a variety of techniques known collectively as "Document Q&A" (DocsQA). DocsQA enables your application to provide the LLM with a set of content/documents that serve as the source of truth. This avoids hallucination and enables the model to provide better responses.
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/guides/brand-new.md
```diff
@@ -26,44 +32,47 @@ LLMs, by themselves, aren't great at:
...
 
-The biggest constraint on the prompt is the context window, which is the combined length of the prompt and the model's response. (Because the budget is shared between input and output, the longer your prompt to the model, the shorter the model's response can be.)
+### Context Window = Prompt Length + Model's Response
 
 Each model API has a context window that it supports. For instance, [GPT-4 Standard](https://platform.openai.com/docs/models/gpt-4) is 8,192 tokens. (A [token](https://platform.openai.com/tokenizer) is a common grouping of characters.)
 
+The context window is the combined length of the prompt and the model's response. Effectively this means we have a budget (i.e. the size of thecontext window) that is shared between the input and the output. As we increase the length of the prompt to the model, we decrease the potential length of the response from the model.
```
NickHeiner, line 74:
```suggestion
The context window is the combined length of the prompt and the model's response. Effectively this means we have a budget (i.e. the size of the context window) that is shared between the input and the output. As we increase the length of the prompt to the model, we decrease the potential length of the response from the model.
```

packages/docs/docs/guides/brand-new.md
```diff
@@ -124,17 +137,19 @@ const tools: Record<string, Tool> = {
 
 More detail: [`UseTools` API docs](../api/modules/batteries_use_tools#usetools).
 
-## Accessing Knowledge ("Docs QA")
+### Accessing Knowledge (AKA "Docs QA")
+
+LLMs have "soft knowledge" of the world, but if you just ask a question without providing any context, they're prone to [hallucination](<https://en.wikipedia.org/wiki/Hallucination_(artificial_intelligence)>). Additionally, because LLMs are only trained on public data, they don't have context on private data your or your app may care about.
```
NickHeiner, line 142:
```suggestion
LLMs have "soft knowledge" of the world, but if you just ask a question without providing any context, they're prone to [hallucination](https://en.wikipedia.org/wiki/Hallucination_(artificial_intelligence)). Additionally, because LLMs are only trained on public data, they don't have context on private data your or your app may care about.
```

Are the `<` and `>` necessary?

NickHeiner (approved):

packages/docs/docs/guides/brand-new.md
benlower: apparently they are required by prettier. must be a way to have the parens in the URL not interfere with the markdown syntax for links.

PR #184, opened 2023-07-12
Update prompting guide

benlower:


packages/docs/docs/guides/prompting.md
```diff
@@ -24,11 +29,11 @@ function App() {
...
 To configure the output of `ChatCompletion`, use [`ModelProps`](../api/interfaces/core_completion.ModelProps.md). This allows you to do things like making the model more creative or precise, telling the model how long a response you want back, etc. Combined with the natural language of your [prompt](./brand-new.md#prompt-engineering), this is how you control the model's output.
 
-## What about non-chat models?
+## The Problem with Non-Chat Models
 
-Chat models have stronger "instructability", meaning they follow your instructions better.
+Chat models have stronger "instructability". This just means they follow your instructions better.
 
-Conversely, non-chat Completion models are just trained to predict the next token, so you might see output like this:
+Conversely, non-chat completion models are just trained to predict the next token. With a non-chat model, you might see something like this:
```
farzadab, line 36: It's possible they wanted to refer to the component. If that's the case, we can lean into it and add links (haven't verified the link works):
```suggestion
Conversely, non-chat [`Completion`](../api/modules/core_completion#completion) models are just trained to predict the next token. With a non-chat model, you might see something like this:
```

On the other hand, OpenAI is discontinuing completion models. As such part of this documentation (and our code) will become obsolete maybe?

farzadab (approved):
LGTM, added one comment.
Not sure if we should deprecate the completion models too or not. We can keep them in case other providers have them, but I think ATM we don't have any providers in AI.JSX that allows them (I gotta double check to see if OpenAI has stopped responding for them or not).

zkoch (approved):

packages/docs/docs/guides/prompting.md
benlower: thx for flagging this but i don't think we want to link to the component in this particular context as this is referring in general to chat vs. non-chat models.

PR #186, opened 2023-07-12
[WIP] UICompletion Improvements + Story Teller Demo

farzadab:
With this PR:
- UICompletion adds support to AI element generation (tested on `ImageGen`)
  - this includes making sure this feature works with UI streaming
- UICompletion interface is updated so that:
  - allow for docs instead of examples
  - 1-shot generation (e.g. instead of needing to 2 LLM calls, one to create `recipe`, one to create UI) 
- a story telling demo that uses these features

Note: this PR currently only works with the `create-react-app-demo`. As such it should not be merged. @petersalas since you're back tomorrow, let's talk about support for NextJS.


packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -81,13 +82,14 @@ export async function* JsonChatCompletion(
   props: Omit<TypedObjectCompletionWithRetry, 'typeName' | 'parser' | 'partialResultCleaner'>,
   { render }: AI.ComponentContext
 ) {
+  const partialResultCleaner = 'default' in untruncateJson ? untruncateJson.default : untruncateJson;
```
NickHeiner, line 85: Why is this `'default' in` check necessary?

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -173,11 +186,21 @@ async function* OneShotObjectCompletion(
...
       if (partial.done) {
-        logger.warn(
-          { output: partial.value, cleaned: partialResultCleaner ? str : undefined, errorMessage: e.message },
-          "ObjectCompletion failed. The final result either didn't parse or didn't validate."
+        // logger.warn(
+        //   { output: partial.value, cleaned: partialResultCleaner ? str : undefined, errorMessage: e.message },
+        //   "ObjectCompletion failed. The final result either didn't parse or didn't validate."
+        // );
+        console.log('bad output:', partial.value);
+        throw new AIJSXError(
```
NickHeiner, line 194: Did you mean to use `CompletionError` here?

packages/ai-jsx/src/experimental/next/index.ts
```diff
@@ -16,7 +16,11 @@ import _ from 'lodash';
  */
 const boundaryElements = [
   { tag: AI.React, unwrap: (e: AI.Element<any>) => e.props.children },
-  { tag: Image, unwrap: (e: AI.Element<any>) => ReactModule.createElement('img', { src: e.props.url }) },
+  {
+    tag: Image,
+    unwrap: (e: AI.Element<any>) =>
+      ReactModule.createElement('img', { src: e.props.url, alt: `Image generated from the prompt: ${e.props.prompt}` }),
```
NickHeiner, line 22: Nice use of the alt tag!

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -1,65 +1,73 @@
...
-  collectComponents(example, true);
+  const memoizedComponents = new Map<string, AI.Node>();
 
-  const validComponents = Object.fromEntries(Array.from(reactComponents).map((c) => [reactComponentName(c), c]));
-
-  interface SerializedComponent {
-    name: string;
-    children: (string | SerializedComponent)[];
+  function getOrCreateMemo(serializedComponent: SerializedComponent, nonMemoized: AI.Node): AI.Node {
+    const hash = CryptoJS.SHA256(JSON.stringify(serializedComponent)).toString();
```
NickHeiner, line 50: Why hash the `serializedComponent` JSON, instead of using the JSON as the key directly?

NickHeiner:
I'll let @petersalas take the lead on reviewing this.

packages/ai-jsx/src/react/completion.tsx
farzadab: Woah, I didn't know I could do that. Actually I am changing this part anyway, since it might `memoize` too much. If you have two images with the same prompt, it's gonna give you back the same image (which is probably not the default behaviour). I'm gonna give it a path-like string as its key instead.

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: Bummer. Ok, that's fine. It would be nice to add a comment in the code saying something like this. :smile:

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -77,40 +85,124 @@ export async function* UICompletion(
...
+        <Component>{children.map(toComponent)}</Component>
+      </AI.React>
+    );
   }
 
   const Element: z.Schema = z.object({
-    name: z.string().refine((c) => Boolean(validComponents[c]), {
-      message: `Unknown component "name". Supported components: ${Object.keys(validComponents)}`,
+    tag: z.string().refine((c) => Boolean(validComponentsMap[c]), {
+      message: `Unknown component "tag". Supported components: ${Object.keys(validComponentsMap)}`,
```
NickHeiner, line 105: Hmm, is there any way to get the invalid `tag` value in this error message? (I see that we didn't have it previously either. :smile:)

NickHeiner, line 105: Also, there's something wrong with the logic for getting the name of supported components. I see this in the error message.

```
Supported components: undefined,Header,Paragraph,Container,ImageGen
```

packages/ai-jsx/src/batteries/constrained-output.tsx
NickHeiner: Also, we'll need to upstream this change to #175.

packages/ai-jsx/src/react/completion.tsx
NickHeiner: The `undefined` comes from the fact that one of the elements in `allComponents` is `Symbol(react.fragment)`. I fixed the `undefined` by making this change:

```ts
function reactComponentName(component: AI.Component<any> | React.JSXElementConstructor<any> | string) {
  return typeof component === 'string' ? 
    component : 
    typeof component === 'symbol' ? 
      // @ts-expect-error
      component.description : 
      component.name;
}
```

However, I had to add the `ts-expect-error` because `AI.Component` doesn't include `Symbol`, so I think something deeper may be going wrong here.

packages/ai-jsx/src/react/completion.tsx
```diff
@@ -77,40 +85,124 @@ export async function* UICompletion(
...
     }),
-    children: z.array(z.union([z.string(), z.lazy(() => Element)])),
+    children: z.union([z.string(), z.array(z.union([z.string(), z.lazy(() => Element)]))]),
   });
+  const Elements = z.union([Element, z.array(Element)]);
```
NickHeiner, line 109: Variable names should be lower-cased unless they refer to TS types or JSX components.

PR #188, opened 2023-07-13
Fix append-only renders with non-append-only generators

### LINKED ISSUE #167: appendOnly renders might not be appendOnly
If a component yields two non-append-only frames, the append-only render will not be append-only. The fix is to disable yielding intermediate values from non-append-only streams when processing an AsyncIterator in `renderStream`.

petersalas:
Fixes #167


NickHeiner (approved):
Thanks for catching this + cleaning it up.

PR #190, opened 2023-07-13
Add Anthropic Polyfills

petersalas:
- Add  `<SystemMessage>` polyfill
- Add `<UserMessage name={...}>` polyfill
- Remove leading space from output


packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -105,33 +105,47 @@ export async function* AnthropicChatModel(
...
-          );
-      }
-    })
+    messageElements
+      .filter(AI.isElement)
+      .flatMap((message) => {
+        if (message.tag === SystemMessage) {
+          return [
+            <UserMessage>For subsequent replies you will adhere to the following instructions: {message}</UserMessage>,
+            <AssistantMessage>Okay, I will do that.</AssistantMessage>,
```
NickHeiner, line 114: heh, does this help? I've experimented with stuff like this in the past but haven't ever rigorously tested it.

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -105,33 +105,47 @@ export async function* AnthropicChatModel(
...
+          ];
+        }
+
+        return message;
+      })
+      .map(async (message) => {
+        switch (message.tag) {
+          case UserMessage:
+            return `${AnthropicSDK.HUMAN_PROMPT}:${message.props.name ? ` (${message.props.name})` : ''} ${await render(
```
NickHeiner, line 123: Can you add an example of using a `name` prop?

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -171,10 +185,18 @@ export async function* AnthropicChatModel(
...
+    if (isFirstResponse && text.length > 0) {
+      isFirstResponse = false;
+      if (text.startsWith(' ')) {
+        text = text.slice(1);
+      }
+    }
+    resultSoFar += text;
     logger.trace({ completion }, 'Got Anthropic stream event');
-    yield completion.completion;
+    yield text;
```
NickHeiner, line 199: heh, nice

NickHeiner:
If we had a better unit testing story, this would be a DELIGHTFUL AND WONDERFUL application of it.

NickHeiner (approved):

packages/ai-jsx/src/lib/anthropic.tsx
petersalas: It seemed to work well enough for the demos in `examples` that I tried 😂

NickHeiner: But like, you noticed a perf improvement with it?

petersalas: As compared to blowing up, yes :P

Langchain IIRC does something like `User: <admin>$SYSTEM_MESSAGE</admin>`, and this seemed to work better than that in the toy examples I tried 🤷

packages/ai-jsx/src/lib/anthropic.tsx
petersalas: The debate example uses it, which is what I was testing against

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: I didn't mean compared to blowing up – I meant specifically just including the `AssistantMessage`.

PR #198, opened 2023-07-19
UICompletion revamp

NickHeiner:
I rebased #186 on #175, then added some changes of my own. Most of them are just making those two branches work together.

I also switched the model to output Markdown. I think this will be more maintainable and more stable than having a bunch of individual `Header` `List` `Paragraph` components, etc. But as a todo, I still need to incorporate other more complicated components (cards, forms, etc).

This basically works and seems like a good approach. https://github.com/fixie-ai/ai-jsx/pull/175/files#r1268278018 is the biggest blocker right now.

I was initially seeing issues where the model would generate malformed output ~10% of the time – generally hallucinating components that don't exist. I think those issues are reduced or eliminated after switching to OpenAI function calling.

### DIFF
--- a/packages/ai-jsx/package.json
+++ b/packages/ai-jsx/package.json
@@ -303,9 +303,11 @@
   "dependencies": {
     "@anthropic-ai/sdk": "^0.5.0",
     "@nick.heiner/wandb-fork": "^0.5.2-5",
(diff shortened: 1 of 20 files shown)


NickHeiner:
I'm going to adapt this into a different PR.

PR #202, opened 2023-07-20
Evaluate output fidelity

NickHeiner:


NickHeiner:
I'm going to send this as several smaller PRs.

PR #204, opened 2023-07-20
Model provider update

farzadab:
This PR:
- Anthropic fix: throws an error when `functionDefinitions` are provided
- updates OpenAI chat model names
- improves error handling for JsonChatCompletion

### DIFF
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -244,8 +244,12 @@ async function* ObjectCompletionWithRetry(
     output = yield* render(childrenWithCompletion);
     return output;
   } catch (e: any) {
-    validationError = e.metadata.validationError;
-    output = e.metadata.output;
+    if (e instanceof CompletionError) {
+      validationError = e.metadata.validationError;
+      output = e.metadata.output;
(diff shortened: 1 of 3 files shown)


packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -244,8 +244,12 @@ async function* ObjectCompletionWithRetry(
     output = yield* render(childrenWithCompletion);
     return output;
   } catch (e: any) {
-    validationError = e.metadata.validationError;
-    output = e.metadata.output;
+    if (e instanceof CompletionError) {
+      validationError = e.metadata.validationError;
+      output = e.metadata.output;
+    } else {
+      throw e;
```
NickHeiner, line 251: Nice. It feels safer to only explicitly handle errors we know about and throw the rest.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -48,13 +48,16 @@ type ValidCompletionModel =
 
 type ValidChatModel =
```
NickHeiner, line 49: Thanks for updating these!

NickHeiner (approved):
Looks great!

PR #205, opened 2023-07-20
Use openai's function_call param to force the "print" function

mdepinet:
to be used in JsonChatCompletionFunctionCall

### DIFF
--- a/packages/ai-jsx/src/lib/openai.tsx
+++ b/packages/ai-jsx/src/lib/openai.tsx
@@ -3,40 +3,41 @@
  * @packageDocumentation
  */
 
+import GPT3Tokenizer from 'gpt3-tokenizer';
+import {
+  ChatCompletionFunctions,
+  ChatCompletionRequestMessage,
+  ChatCompletionResponseMessage,
+  Configuration,
+  CreateChatCompletionRequestFunctionCall,
+  CreateChatCompletionResponse,
+  CreateCompletionResponse,
+  CreateImageRequestResponseFormatEnum,
+  CreateImageRequestSizeEnum,
+  OpenAIApi,
+  ResponseTypes,
+} from 'openai-edge';
+import { Merge, MergeExclusive } from 'type-fest';
 import {
   AssistantMessage,
   ChatProvider,
   CompletionProvider,
+  FunctionCall,
+  FunctionDefinition,
+  FunctionResponse,
   ModelProps,
   ModelPropsWithChildren,
   SystemMessage,
   UserMessage,
-  FunctionDefinition,
-  FunctionCall,
-  FunctionResponse,
   getParametersSchema,
 } from '../core/completion.js';
+import { AIJSXError, ErrorCode, HttpError } from '../core/errors.js';
 import { Image, ImageGenPropsWithChildren } from '../core/image-gen.js';
-import {
-  Configuration,
-  CreateChatCompletionResponse,
-  CreateCompletionResponse,
-  OpenAIApi,
-  ChatCompletionFunctions,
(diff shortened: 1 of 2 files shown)


packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -3,40 +3,41 @@
...
+} from 'openai-edge';
+import { Merge, MergeExclusive } from 'type-fest';
 import {
   AssistantMessage,
   ChatProvider,
   CompletionProvider,
+  FunctionCall,
+  FunctionDefinition,
+  FunctionResponse,
   ModelProps,
```
farzadab, line 28: Nit: I'm slightly confused why there is so much change here. Is this just formatting going haywire?

farzadab (approved):
LGTM, good job noticing this Mike!
I'm slightly confused by the import changes though.

packages/ai-jsx/src/lib/openai.tsx
mdepinet: They're just being alphabetized. I probably need to make my formatter less aggressive :sweat_smile:

NickHeiner: Or we could update the project linter to alphabetize. :smile:

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -314,6 +317,20 @@ export async function* OpenAIChatModel(
     );
   }
 
+  if (props.forcedFunction) {
+    if (
+      !Object.entries(props.functionDefinitions)
+        .map(([functionName, _]) => functionName)
+        .find((f) => f == props.forcedFunction)
+    ) {
```
NickHeiner, line 325:
```tsx
if (!Object.keys(props.functionDefinitions).includes(props.forcedFunction)
```

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -392,6 +414,7 @@ export async function* OpenAIChatModel(
     temperature: props.temperature,
     messages,
     functions: openaiFunctions,
+    function_call: openaiFunctionCall,
```
NickHeiner, line 417: Damn, I didn't know about this! Nice.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -384,6 +401,11 @@ export async function* OpenAIChatModel(
         description: functionDefinition.description,
         parameters: getParametersSchema(functionDefinition.parameters),
       }));
+  const openaiFunctionCall: CreateChatCompletionRequestFunctionCall | undefined = !props.forcedFunction
+    ? undefined
+    : {
+        name: props.forcedFunction,
+      };
```
NickHeiner, line 408: I would flip this – this reads more naturally to me:

```ts
const openAIFunctionCall: CreateChatCompletionRequestFunctionCall = props.forcedFunction ? { name: props.forcedFunction } : undefined;
```

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -314,6 +317,20 @@ export async function* OpenAIChatModel(
     );
   }
 
+  if (props.forcedFunction) {
+    if (
```
NickHeiner, line 321: Seems like it would be cleaner to combine these two into one `if`.

packages/ai-jsx/src/lib/openai.tsx
mdepinet: Fixed in #207

PR #206, opened 2023-07-20
Bugfix: schema and validators being ignored in JCCFC

farzadab:


farzadab:
Merging to get rid of the bug since we probably won't get a review today.

packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -327,7 +330,26 @@ export async function* JsonChatCompletionFunctionCall(
 
   const frames = render(childrenWithCompletion);
   for await (const frame of frames) {
-    yield JSON.stringify(JSON.parse(frame).arguments);
+    const object = JSON.parse(frame).arguments;
```
NickHeiner, line 333: Let's call this `jsonResult` – `object` is super generic. Also, `json` doesn't have to be an object. :smile:

NickHeiner:
Hmm, so what was the bug? Good catch that `validators` weren't being respected by JCCFC, but it looks like `UICompletion` wasn't passing that anyway? Was it just that the schema was too narrow?

farzadab:
Yes, but it was passing in the schema and that wasn't being respected either. The issue was that it was sending back an object which didn't follow the schema that `UICompletion` expected. As a result it would throw an error even though it was an intermediate result that would have been fixed.

So I guess the contract is that JCC should never return a string that isn't an object or doesn't follow the schema (+ validators which are not used anywhere).

PR #207, opened 2023-07-21
Add DocsQAWithSources component

mdepinet:
This extends on DocsQA by encouraging the LLM to provide sources for its answer and allowing the developer to format the result as they like.

### DIFF
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -766,12 +766,15 @@ export interface DocsQAProps<ChunkMetadata extends Jsonifiable = Jsonifiable> {
    * ```
    */
   chunkFormatter?: (props: { doc: ScoredChunk<ChunkMetadata> }) => Node;
+}
 
+export interface DocsQAWithCitationsProps<ChunkMetadata extends Jsonifiable = Jsonifiable>
+  extends DocsQAProps<ChunkMetadata> {
   /**
-   * The component used to format results from a DocsQAWithSources query.
+   * The component used to format results from a DocsQAWithCitations query.
    *
    * ```tsx
-   *   function FormatQAResult(result: QAWithSourcesResult) {
+   *   function FormatQAResult(result: QAWithCitationsResult) {
    *     if (result.sources?.length) {
    *       return `${result.answer}\n\nSources:\n${result.sources.join('\n')}`;
    *     }
@@ -779,7 +782,7 @@ export interface DocsQAProps<ChunkMetadata extends Jsonifiable = Jsonifiable> {
    *   }
    * ```
    */
-  resultFormatter?: (result: QAWithSourcesResult) => Node;
+  resultFormatter?: (result: QAWithCitationsResult) => Node;
 }
 
 /**
@@ -807,17 +810,17 @@ export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(pr
   );
 }
 
-const ResultSchema: z.ZodObject<any> = z
+const ResultSchema = z
   .object({
     answer: z.string().describe("The answer to the user's question"),
     sources: z.array(z.string()).describe('The title or URL of each document used to answer the question'),
   })
   .required({ answer: true });
 
-export type QAWithSourcesResult = z.infer<typeof ResultSchema>;
+export type QAWithCitationsResult = Partial<z.infer<typeof ResultSchema>>;
 
-/** A default component for formatting DocsQAWithSources results. */
-function DefaultQAResultFormatter(result: QAWithSourcesResult) {
+/** A default component for formatting DocsQAWithCitations results. */
+function DefaultQAResultFormatter(result: QAWithCitationsResult) {
   if (result.sources?.length) {
     return `${result.answer}\n\nSources:\n${result.sources.join('\n')}`;
   }
@@ -827,13 +830,13 @@ function DefaultQAResultFormatter(result: QAWithSourcesResult) {
 /**
  * Similar to {@link DocsQA}, but encourages the LLM to return sources for its answer.
  */
-export async function* DocsQAWithSources<ChunkMetadata extends Jsonifiable = Jsonifiable>(
-  props: DocsQAProps<ChunkMetadata>,
-  { render }: AI.ComponentContext
+export async function* DocsQAWithCitations<ChunkMetadata extends Jsonifiable = Jsonifiable>(
+  props: DocsQAWithCitationsProps<ChunkMetadata>,
+  { render, logger }: AI.ComponentContext
 ) {
   const chunks = await props.corpus.search(props.question, { limit: props.chunkLimit });
   const chunkFormatter: (props: { doc: ScoredChunk<ChunkMetadata> }) => Node = props.chunkFormatter ?? DefaultFormatter;
-  const resultFormatter: (result: QAWithSourcesResult) => Node = props.resultFormatter ?? DefaultQAResultFormatter;
+  const resultFormatter: (result: QAWithCitationsResult) => Node = props.resultFormatter ?? DefaultQAResultFormatter;
 
   const stringifiedResult = (
     <JsonChatCompletion schema={ResultSchema}>
@@ -853,7 +856,7 @@ export async function* DocsQAWithSources<ChunkMetadata extends Jsonifiable = Jso
     try {
       yield resultFormatter(ResultSchema.parse(JSON.parse(frame)));
     } catch (e) {
-      // The intermediate result isn't valid yet. Return it as is instead.
+      logger.debug(`Failed to parse DocsQAWithCitations frame: ${e}`);
       yield frame;
     }
   }
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -828,7 +828,7 @@ function DefaultQAResultFormatter(result: QAWithCitationsResult) {
 }
 
 /**
- * Similar to {@link DocsQA}, but encourages the LLM to return sources for its answer.
+ * Similar to {@link DocsQA}, but encourages the LLM to return citations for its answer.
  */
 export async function* DocsQAWithCitations<ChunkMetadata extends Jsonifiable = Jsonifiable>(
   props: DocsQAWithCitationsProps<ChunkMetadata>,
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -14,6 +14,7 @@ import { Jsonifiable } from 'type-fest';
 import z from 'zod';
 import { ChatCompletion, SystemMessage, UserMessage } from '../core/completion.js';
 import { AIJSXError, ErrorCode } from '../core/errors.js';
+import * as AI from '../index.js';
 import { Node } from '../index.js';
 import { getEnvVar } from '../lib/util.js';
 import { JsonChatCompletion } from './constrained-output.js';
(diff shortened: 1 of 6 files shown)


packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -789,3 +806,56 @@ export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(pr
     </ChatCompletion>
   );
 }
+
```
NickHeiner, line 809: I like this approach! This is a neat way of getting the model to give a structured output so it can be fed into props of another component.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -789,3 +806,56 @@ export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(pr
...
+export async function* DocsQAWithSources<ChunkMetadata extends Jsonifiable = Jsonifiable>(
+  props: DocsQAProps<ChunkMetadata>,
+  { render }: AI.ComponentContext
+) {
+  const chunks = await props.corpus.search(props.question, { limit: props.chunkLimit });
+  const chunkFormatter: (props: { doc: ScoredChunk<ChunkMetadata> }) => Node = props.chunkFormatter ?? DefaultFormatter;
+  const resultFormatter: (result: QAWithSourcesResult) => Node = props.resultFormatter ?? DefaultQAResultFormatter;
+
+  const stringifiedResult = (
+    <JsonChatCompletion schema={ResultSchema}>
```
NickHeiner, line 839: @farzadab I'm glad you improved `JsonChatCompletion`, because I think it'll keep getting a lot of use.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -789,3 +806,56 @@ export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(pr
...
+      <UserMessage>{props.question}</UserMessage>
+    </JsonChatCompletion>
+  );
+
+  const frames = render(stringifiedResult);
+  for await (const frame of frames) {
+    try {
+      yield resultFormatter(ResultSchema.parse(JSON.parse(frame)));
+    } catch (e) {
+      // The intermediate result isn't valid yet. Return it as is instead.
```
NickHeiner, line 856: `JsonChatCompletion` should always emit valid JSON, because as the result is streaming in, it "untruncates" it. (Thanks @farzadab!)

That said, I did observe GPT-4 failing to emit valid JSON approximately 3% of the time. But in that case, it's fully borked and will never recover, so we wouldn't say "isn't valid yet".

Also, `resultFormatter` should tolerate a partial JSON object, because of how the JSON is streamed in – to encode that, I'd change this to:

```ts
QAWithSourcesResult = Partial<z.infer<typeof ResultSchema>>;
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -789,3 +806,56 @@ export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(pr
...
+      yield frame;
+    }
+  }
+  return resultFormatter(ResultSchema.parse(JSON.parse(await frames)));
+}
```
NickHeiner, line 861: I think this whole thing can be something like:

```tsx
return yield* render(stringifiedResult, {
  map: frame => {
    try { 
      return resultFormatter(ResultSchema.parse(JSON.parse(frame)));  
    } catch (e) { return frame }
  }
});
```

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: Also, this could be a good place for a `logger.trace` or `logger.debug`. (`logger` is on `AI.ComponentContext`.)

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -789,3 +806,56 @@ export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(pr
     </ChatCompletion>
   );
 }
+
+const ResultSchema: z.ZodObject<any> = z
```
NickHeiner, line 810:
```suggestion
const ResultSchema = z
```

I know this was copied from somewhere else where it was needed, but in this case, we don't need it. (It was only needed when we were doing a recursive type.)

The type annotation is more generic than reality, so our types end up sadly looser:

image
image

When we omit the type annotation, the types flow through nicely:

image
image

And of course these nicer types flow through to userland.

packages/tutorial/src/docsqa.tsx
```diff
@@ -17,22 +25,53 @@ const docs = [
...
+      return source;
+    }) || [];
+
+  if (linkedSources.length) {
+    return `${result.answer} (from ${linkedSources.join(', ')})`;
+  }
+  return result.answer;
+}
+
 function App() {
```
NickHeiner, line 58: Suggested changes here, just to make it a little clearer in the output:

```diff
diff --git a/packages/tutorial/src/docsqa.tsx b/packages/tutorial/src/docsqa.tsx
index b4dc3ca..8201fef 100644
--- a/packages/tutorial/src/docsqa.tsx
+++ b/packages/tutorial/src/docsqa.tsx
@@ -58,8 +58,7 @@ function OptionalCustomResultFormatter(result: QAWithSourcesResult) {
 function App() {
   return (
     <>
-      <DocsQAWithSources question="What was Hurricane Katrina?" corpus={corpus} chunkLimit={5} />
-      {'\n\n'}
+      DocsQA without source citations:{'\n'}
       <DocsQA
         question="Which dates did the storm occur?"
         corpus={corpus}
@@ -67,6 +66,11 @@ function App() {
         chunkFormatter={OptionalCustomChunkFormatter}
       />
       {'\n\n'}
+      {'\n\n'}
+      DocsQA with source citations:{'\n'}
+      <DocsQAWithSources question="What was Hurricane Katrina?" corpus={corpus} chunkLimit={5} />
+      {'\n\n'}
+      DocsQA with source citations with a custom results formatter:{'\n'}
       <DocsQAWithSources
         question="Where were the strongest winds reported?"
         corpus={corpus}
```

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -789,3 +806,56 @@ export async function DocsQA<ChunkMetadata extends Jsonifiable = Jsonifiable>(pr
...
+  if (result.sources?.length) {
+    return `${result.answer}\n\nSources:\n${result.sources.join('\n')}`;
+  }
+  return result.answer;
+}
+
+/**
+ * Similar to {@link DocsQA}, but encourages the LLM to return sources for its answer.
+ */
+export async function* DocsQAWithSources<ChunkMetadata extends Jsonifiable = Jsonifiable>(
```
NickHeiner, line 830:
```suggestion
export async function* DocsQAWithCitations<ChunkMetadata extends Jsonifiable = Jsonifiable>(
```

Minor, but I think this name could be clearer. All docs QA has sources, right? But the name `citations` to me is a little more evocative of citing those sources in the response. (not a blocker / do whatever you want.)

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -763,6 +766,20 @@ export interface DocsQAProps<ChunkMetadata extends Jsonifiable = Jsonifiable> {
...
+   * ```tsx
+   *   function FormatQAResult(result: QAWithSourcesResult) {
+   *     if (result.sources?.length) {
+   *       return `${result.answer}\n\nSources:\n${result.sources.join('\n')}`;
+   *     }
+   *     return result.answer;
+   *   }
+   * ```
+   */
+  resultFormatter?: (result: QAWithSourcesResult) => Node;
```
NickHeiner, line 782: `DocsQA` doesn't use `resultFormatter`, right? But it does take `DocsQAProps` as props? We shouldn't have an unused field in the props for `DocsQA`.

Instead, I'd do something like:

```tsx
interface DocsQAWithSourcesProps extends DocsQAProps {
  resultFormatter?: (result: QAWithSourcesResult) => Node;
}
```

NickHeiner (approved):

packages/docs/docs/guides/docsqa.md
```diff
@@ -86,7 +86,7 @@ flowchart LR
   que[Query] -->|string| embed2[Embed] -->|vector| vdb2[(Vector DB)] -->|similar chunks| LLM
 ```
 
-If you use the built-in DocsQA tag from AI.JSX, then you just need to decide how to present the chunk to your LLM:
+If you use the built-in `DocsQA` component from AI.JSX, then you just need to decide how to present the chunk to your LLM:
```
farzadab, line 89: Nit: you can link to the document page (not necessary if it's been done before).

Please check the format, I haven't checked it myself.
```suggestion
If you use the built-in [`DocsQA`](../api/modules/batteries_docs#docsqa) component from AI.JSX, then you just need to decide how to present the chunk to your LLM:
```

packages/docs/docs/guides/docsqa.md
```diff
@@ -109,6 +109,52 @@ function AskAndAnswer({ query }: { query: string }) {
...
+The `DocsQA` component provides an answer, like:
+
+```tsx
+<DocsQA question="What is the atomic number of nitrogen?" corpus={corpus} docComponent={ShowDoc} />
+/* Renders:
+    Nitogen's atomic number is 7
+*/
+```
+
+If you want an answer that cites sources, use `DocsQAWithSources`:
```
farzadab, line 121: Same here (and if you do please check the links work):
```suggestion
If you want an answer that cites sources, use [`DocsQAWithSources`](../api/modules/batteries_docs#docsqawithsources):
```

packages/ai-jsx/src/batteries/docs.tsx
farzadab: Yup, if you pass in the schema, the result should always match that too (sorry for a period there, this didn't work correctly, but it should now and if you see it working otherwise let me know to fix it).

Nick: What does `Partial` do?

Also I'm generally a bit confused about what type of logging should go where since I don't know what happens in terminal vs react vs next. So my logging is usually pretty sloppy.

farzadab (approved):
This makes a lot of sense.
I put in some minor comments, but otherwise LGTM.

mdwelsh (approved):
Nice!

packages/tutorial/src/docsqa.tsx
```diff
@@ -17,22 +25,56 @@ const docs = [
...
   return (
     <>
-      <DocsQA question="What was Hurricane Katrina?" corpus={corpus} chunkLimit={5} chunkFormatter={GetChunk} />
+      DocsQA without source citations:{'\n'}
+      <DocsQA
+        question="Which dates did the Huricane Katrina occur?"
```
mdwelsh, line 63:
```suggestion
        question="Which dates did the Hurricane Katrina occur?"
```

packages/ai-jsx/src/batteries/docs.tsx
mdepinet: Based on the docs (and a very quick read) for render, map only gets applied to intermediate results, so I don't think this quite works.

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: > Also I'm generally a bit confused about what type of logging should go where since I don't know what happens in terminal vs react vs next. So my logging is usually pretty sloppy.

Hmm, can you tell me more about what you're trying to figure out here?

NickHeiner: Re: partial https://www.perplexity.ai/search/5d43f745-8205-468a-b9fe-64664383328d?s=c

packages/ai-jsx/src/batteries/docs.tsx
NickHeiner: Hmm @petersalas we should probably do something more ergonomic/obvious here.

PR #208, opened 2023-07-21
Add llama2 support

NickHeiner:
I originally had a model with a base Llama2 implementation, which could be customized to run for a particular provider (e.g. Replicate). But then I realized that every provider would have a different API, including things like what types of messages (e.g. system, assistant) could be passed. So there's not that much to share.


packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -88,23 +88,21 @@ export function Anthropic({
...
 
-interface AnthropicChatModelProps extends ModelPropsWithChildren {
-  model: ValidChatModel;
-}
-export async function* AnthropicChatModel(
-  props: AnthropicChatModelProps,
-  { render, getContext, logger }: AI.ComponentContext
-): AI.RenderableStream {
-  const messageElements = await render(props.children, {
+export async function getPolyfilledMessages(
```
NickHeiner, line 91: I extracted this logic so I could use the same polyfilling for llama2.

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -88,23 +88,21 @@ export function Anthropic({
...
+) {
+  const messageElements = await render(children, {
     stop: (e) =>
       e.tag == SystemMessage ||
       e.tag == UserMessage ||
       e.tag == AssistantMessage ||
       e.tag == FunctionCall ||
       e.tag == FunctionResponse,
   });
-  yield AI.AppendOnlyStream;
```
NickHeiner, line 106: @petersalas we're changing the timing of when this is `yield`ed slightly. Does that matter?

packages/ai-jsx/src/lib/llama2.tsx
```diff
@@ -0,0 +1,174 @@
...
+    prompt,
+    max_length: props.maxTokens ?? defaultMaxTokens,
+    repetition_penalty: props.repetitionPenalty,
+    temperature: props.temperature,
+    top_p: props.topP,
+  };
+  logger.debug({ llama2Args }, 'Calling Llama2');
+  const response = await props.fetch(llama2Args);
+  yield* response;
+  return AI.AppendOnlyStream;
```
NickHeiner, line 95: @petersalas is this the right thing to be doing?

packages/ai-jsx/src/lib/llama2.tsx
```diff
@@ -0,0 +1,174 @@
...
+  ...defaults
+}: { children: Node; fetchLlama2: FetchLlama2 } & Llama2ModelProps) {
+  return (
+    <ChatProvider component={Llama2ChatModel} {...defaults} fetch={fetchLlama2}>
+      <CompletionProvider component={Llama2CompletionModel} {...defaults} fetch={fetchLlama2}>
+        {children}
+      </CompletionProvider>
+    </ChatProvider>
+  );
+}
```
NickHeiner, line 138: @petersalas this is much simpler than the Anthropic and OpenAI components, which makes me somewhat concerned. Am I missing something?

packages/ai-jsx/src/lib/llama2.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  async function* fetchLlama2(input: FetchLlama2Args) {
+    const output = (await replicate.run(
+      'replicate/llama70b-v2-chat:e951f18578850b652510200860fc4ea62b3b16fac280f83ff32282f87bbd2e48',
+      { input }
+    )) as string[];
+    const result = output.join('');
+    logger.debug({ output, result }, 'Replicate output');
+    yield result;
+    return result;
```
NickHeiner, line 166: Unbelievably, this doesn't support streaming. The closest it has is the ability to poll an endpoint. 🤮

packages/docs/docs/guides/models.md
NickHeiner, line 1: @benlower I see you have a PR against this file too – I'm only adding content to the end, so I don't think we'll see merge issues. But also as an fyi in case you want to modify this content too. Feel free to edit this branch directly if you'd like.

packages/ai-jsx/src/core/errors.ts
```diff
@@ -24,9 +24,12 @@ export enum ErrorCode {
   AnthropicDoesNotSupportSystemMessage = 1019,
   AnthropicDoesNotSupportFunctions = 1020,
   AnthropicAPIError = 1021,
+
   ChatModelDoesNotSupportFunctions = 1022,
   ChatCompletionBadInput = 1023,
 
+  Llama2DoesNotSupportFunctions = 1024,
```
juberti, line 31: what's the difference between this code and 1022? As the number of supported models grows, seems like having per-model codes is going to get unwieldy.

packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -120,16 +118,16 @@ export async function* AnthropicChatModel(
       .map(async (message) => {
         switch (message.tag) {
           case UserMessage:
-            return `${AnthropicSDK.HUMAN_PROMPT}:${message.props.name ? ` (${message.props.name})` : ''} ${await render(
+            return `${AnthropicSDK.HUMAN_PROMPT}${message.props.name ? ` (${message.props.name})` : ''} ${await render(
```
juberti, line 121: Do we still want the AnthropicSDK references here?

packages/ai-jsx/src/lib/anthropic.tsx
petersalas: Rendering blocks until the first `yield`, so you generally want to `yield AI.AppendOnlyStream` as early as possible (namely before any `await`s`).

packages/ai-jsx/src/lib/llama2.tsx
petersalas: Yeah, this is consistent with the convention in the other places.  (Alternatively you could `return null` or `return ""`, but IMO `return AI.AppendOnlyStream` also indicates _why_ the return value is empty.)

packages/ai-jsx/src/lib/llama2.tsx
petersalas: I don't think so? Main difference seems to be that with Llama all config is wrapped in the `fetchLlama2` (as opposed to e.g. separate chat/completion model configs/OpenAI client)

(It looks like `<Anthropic>`'s implementation was adapted from `<OpenAI>` and carried over some machinery from that.)

NickHeiner: Yes, that's definitely how Anthropic was made. Should I simplify Anthropic as well?

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: I checked, and they're just generic `Human: ` and `Assistant: ` strings. I think it's fine + better to just hardcode it than having this function take them as arguments.

packages/ai-jsx/src/core/errors.ts
NickHeiner: hehehe we had this same convo the last time I added a model provider: https://github.com/fixie-ai/ai-jsx/pull/165#discussion_r1245929073.

I hear your feedback and will continue to watch for the pain points noted in that original comment.

packages/ai-jsx/src/lib/anthropic.tsx
juberti: it seems a bit weird for Llama to use these references though. Maybe this goes away when we move to a Llama chat model?

NickHeiner: Yep – I just ripped this out.

packages/ai-jsx/src/lib/anthropic.tsx
NickHeiner: Cool, so moving it forward in this function seems good.

PR #210, opened 2023-07-21
Add chat demo using server actions

petersalas:
This adds a chat demo that uses server actions to the nextjs-experimental demo.

In `ai-jsx/experimental/next`, this:
- Fixes an issue with streaming where suffixes were not correctly being computed
- Adds an `onComplete` prop to `AI.JSX` so that consumers can access the final result (either as a `string` or as a `ReactNode`)
- Adds an experimental `stream` prop that can be used to work around the low streaming framerate caused by `<Suspense>` throttling (at the expense of latency)
- Adds a set of components under `ai-jsx/experimental/next/chat` (`<Chat>`, `<Conversation>`, `<Message>`) that try to put the chat data flow somewhat on rails. Notably, a chat UX can be constructed entirely with server components and server actions -- AI.JSX includes the client components that do all the data plumbing.


packages/ai-jsx/src/experimental/next/chat.client.tsx
NickHeiner, line 1: I like the `.client` convention.

packages/ai-jsx/src/experimental/next/chat.client.tsx
```diff
@@ -0,0 +1,143 @@
...
+ * UI that represents the contents of a message.
+ */
+export interface UIMessage {
+  type: 'assistant' | 'user';
+  message: ReactNode;
+}
+
+/**
+ * A Server Action that processes a message and returns the resulting UI and context for the next message.
+ * See the `sendMessage` helper in `ai-jsx/experimental/next/chat`.
```
NickHeiner, line 16: It would be nice to add a `@see` to a link for what server actions are.

packages/ai-jsx/src/experimental/next/chat.client.tsx
```diff
@@ -0,0 +1,143 @@
...
+ */
+export type SendMessage<T> = (formData: FormData, context: T | undefined) => Promise<[UIMessage[], Promise<T>]>;
+
+/**
+ * React Context containing the `UIMessage`s associated with a conversation. Typically consumed indirectly via
+ * the `<Conversation>` component.
```
NickHeiner, line 22:
```suggestion
 * React Context containing the `UIMessage`s associated with a conversation. You probably don't want to use this direction, and should use the `<Conversation>` component instead.
```

packages/ai-jsx/src/experimental/next/chat.client.tsx
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * React Context containing the UI of the contents of a single message. Typically consumed indirectly via the
+ * `<Message>` component.
+ */
+export const MessageContext = createContext(null as ReactNode);
+
+/**
+ * React Context containing the form status. Typically consumed indirectly via the `<ChatState>` component.
+ */
+export const StatusContext = createContext<'idle' | 'pending' | 'streaming'>('idle');
```
NickHeiner, line 35: Do we need an error state that's distinct from `idle`?

NickHeiner, line 35: Minor, but if we're using `'idle' | 'pending' | 'streaming'` in multiple places, it would be nice to extract into a helper type.

packages/ai-jsx/src/experimental/next/chat.client.tsx
```diff
@@ -0,0 +1,143 @@
...
+      </StatusContext.Provider>
+    </ConversationContext.Provider>
+  );
+}
+
+/**
+ * A component that when used within a `<Chat>` component can vary rendering according to the status.
+ * Accepts `children`, which is rendered when the form is idle; `pending`, which is rendered when
+ * a response has not yet been received from the server; and `streaming`, which is rendered when
+ * a server response is streaming in.
```
NickHeiner, line 91:
```suggestion
 * A component that when used within a `<Chat>` component can vary rendering according to the status.
 * It uses the following props:
 *     * `children`: rendered when the form is idle
 *     * `pending`: rendered when a response has not yet been received from the server
 *     * `streaming`: rendered when a server response is streaming in
```

packages/ai-jsx/src/experimental/next/chat.tsx
```diff
@@ -0,0 +1,110 @@
...
+ */
+export function Conversation(
+  {
+    children,
+  }: {
+    children?: MessageMap;
+  },
+  maybeContext?: AI.ComponentContext
+) {
+  if (maybeContext && 'render' in maybeContext) {
```
NickHeiner, line 52:
```suggestion
  const isInAIJSX = maybeContext && 'render' in maybeContext;
  if (isInAIJSX) {
```

minor, but would make the code more explicit for people will less familiarity

packages/ai-jsx/src/experimental/next/chat.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const serializedMessages = maybeContext.getContext(AIConversationContext);
+    return serializedMessages.map((msg, i) => {
```
NickHeiner, line 54: Hmm, we're so close to being able to use the `ConversationHistory` component here.

packages/ai-jsx/src/experimental/next/chat.tsx
```diff
@@ -0,0 +1,110 @@
...
+        type: 'assistant',
+        message: (
+          <AI.JSX onComplete={setFinalReply}>
+            <AIConversationContext.Provider value={conversationWithUser}>{reply}</AIConversationContext.Provider>
+          </AI.JSX>
+        ),
+      },
+    ],
+    finalReplyPromise.then((finalReply) => conversationWithUser.concat([{ type: 'assistant', message: finalReply }])),
+  ];
```
NickHeiner, line 109: Are these really three sequential values? Or would it make more sense to return an object here, so we can use the key names to communicate intent?

packages/ai-jsx/src/experimental/next/index.tsx
```diff
@@ -72,19 +71,47 @@ function computeSuffix(
...
+
+function StreamRoot({ stream, children }: { stream?: StreamType; children: ReactNode }) {
+  if (stream === 'smooth') {
+    return <Clock serverOrigin={new Date().valueOf()}>{children}</Clock>;
+  }
+
+  return children;
+}
+
+function Chunk({ stream, children }: { stream?: 'smooth' | 'fast'; children: ReactNode }) {
```
NickHeiner, line 84:
```suggestion
function Chunk({ stream, children }: { stream?: StreamType; children: ReactNode }) {
```

packages/ai-jsx/src/experimental/next/index.tsx
```diff
@@ -98,30 +125,34 @@ export const jsx = asJsxBoundary(function jsx(
...
-    replace: (value: ReactModule.ReactNode) => ReactModule.ReactNode;
-  }): Promise<ReactModule.ReactNode> {
+    replace: (value: ReactNode) => ReactNode;
+  }): Promise<ReactNode> {
     const nextResult = await asyncIterator.next();
 
     if (nextResult.done) {
-      return replace(nextResult.value.map(unwrapReact));
+      const finalResult = nextResult.value.map(unwrapReact);
+      const finalText = nextResult.value.filter((v) => typeof v === 'string').join('');
```
NickHeiner, line 139: What are the elements that would be filtered out here?

packages/nextjs-experimental/src/app/basic-chat/page.tsx
```diff
@@ -0,0 +1,92 @@
+/** @jsxImportSource ai-jsx/react */
+import { Chat, Conversation, Message, ChatState, sendMessage, SerializedMessage } from 'ai-jsx/experimental/next/chat';
+import { ChatCompletion, SystemMessage } from 'ai-jsx/core/completion';
+import ResultContainer from '@/components/ResultContainer';
+import { ReactNode, Suspense } from 'react';
+
+async function onSend(formData: FormData, conversation?: SerializedMessage[]) {
+  'use server';
+
+  return sendMessage(
```
NickHeiner, line 10:
```suggestion
  return sendChatMessage(
```

I think this name is a little more specific/clear; IMO `sendMessage` feels too close to the `postMessage` API, which is a totally different thing.

NickHeiner:
I think this looks great. It'll be a good place to start + iterate from.

NickHeiner (approved):

packages/ai-jsx/src/experimental/next/chat.client.tsx
petersalas: Probably for anything real/robust, yes. But deferring that for now.

packages/ai-jsx/src/experimental/next/index.tsx
petersalas: Any React elements produced by AI.JSX. (e.g. `<Recipe>` and friends, or whatever else replaces that in a JIT UI world)

packages/ai-jsx/src/experimental/next/chat.tsx
petersalas: It's true :)

However, on the React side it's ~necessary to make it user-extensible for the presentation and so making it symmetrical on the AI.JSX side seemed like the best choice. (The straw man scenario I came up with was annotating the `<UserMessage>`s with the user's name so that the model addresses the user by name.)

packages/ai-jsx/src/experimental/next/chat.tsx
petersalas: Indeed not :)

PR #211, opened 2023-07-21
ASR components and demo

juberti:
Probably want to migrate this to architecture 4 (ie no route.tsx), but wanted to get the basics landed first.


packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+    this.port.postMessage(data);
+
+    return true;
+  }
+}
+
+registerProcessor("input-processor", InputProcessor);
+`;
+
+class EventTargetImpl implements EventTarget {
```
NickHeiner, line 34: Can we just use EventEmitter for this? https://nodejs.org/api/events.html

packages/nextjs-demo/src/app/asr/page.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const inClean = inText
+      .replace(/[^\w\s]|_/g, '')
+      .replace(/\s+/g, ' ')
+      .trim();
+    return wordErrorRate(refClean, inClean);
+  };
+  const start = () => {
+    const recognizer = SpeechRecognitionFactory.create(id, manager);
+    setRecognizer(recognizer);
+    const el = document.getElementById(id) as HTMLTextAreaElement;
```
NickHeiner, line 51: Use `useRef` to get a handle to a specific element: https://react.dev/reference/react/useRef. (Perplexity can help you with this.)

NickHeiner:
I only glanced at this for a minute but will look more later.

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: note that this code is going to run in-browser

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+    const index = this.listeners[type].indexOf(listener);
+    if (index != -1) {
+      this.listeners[type].splice(index, 1);
+    }
+  }
+  dispatchEvent(e: CustomEvent) {
+    this.listeners[e.type]?.forEach((listener) => listener(e));
+    return true;
+  }
+}
```
NickHeiner, line 55: I don't think any of this is necessary – `EventTarget` implements these itself.

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+
+export class MicManager extends EventTargetImpl {
+  private outBuf?: Float32Array[];
+  private context?: AudioContext;  
+  private streamElement?: HTMLAudioElement;
+  private stream?: MediaStream;
+  private processorNode?: AudioWorkletNode;
+  private numSilentFrames = 0;
+  async startMic(timeslice: number) {
```
NickHeiner, line 64:
```suggestion
  async startMic(timeSlice: number) {
```

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+    }
+  }
+  dispatchEvent(e: CustomEvent) {
+    this.listeners[e.type]?.forEach((listener) => listener(e));
+    return true;
+  }
+}
+
+export class MicManager extends EventTargetImpl {
+  private outBuf?: Float32Array[];
```
NickHeiner, line 58:
```suggestion
  private outBuffer?: Float32Array[];
```

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+      timeslice,
+    );
+  }
+  async startFile(url: string, timeslice: number) {
+    const response = await fetch(url);
+    const blob = await response.blob();
+    this.streamElement = new Audio();
+    this.streamElement.src = URL.createObjectURL(blob);
+    await this.streamElement.play();
+    const stream = await (this.streamElement as any).captureStream();
```
NickHeiner, line 76: Why is the `as any` necessary?

packages/ai-jsx/src/lib/asr/asr.tsx
NickHeiner: Our preference is to fully spell out variable names.

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+    const source = this.context.createMediaStreamSource(stream);
+    source.connect(this.processorNode);
+    this.numSilentFrames = 0;
+  }
+  stop() {
+    this.processorNode!.disconnect();
+    this.stream!.getTracks().forEach((track) => track.stop());
+    if (this.streamElement) {
+      this.streamElement.pause();
+    }
```
NickHeiner, line 110:
```suggestion
this.streamElement?.pause();
```

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+    this.context!.close();
```
NickHeiner, line 111: Will this cause a crash if someone calls `stop` without calling `startGraph` first?

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+    this.stream = undefined;
+    this.context = undefined;
+    this.outBuf = [];
+  }
+  sampleRate() {
+    return this.context?.sampleRate;
+  }
+  private makePcmChunk(buffers: Float32Array[]) {
+    const byteLength = buffers.reduce((sum, buf) => sum + buf.length, 0) * 2;
+    const pcmBuffer = new ArrayBuffer(byteLength);
```
NickHeiner, line 122: Would you mind spelling out the acronyms?

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+      this.listeners[type].splice(index, 1);
+    }
+  }
+  dispatchEvent(e: CustomEvent) {
+    this.listeners[e.type]?.forEach((listener) => listener(e));
+    return true;
+  }
+}
+
+export class MicManager extends EventTargetImpl {
```
NickHeiner, line 57: For every `export`ed class, please include a `/** */` comment at the top that either explains how to use it (if it's a public API) or marks it as `@hidden`.

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+  protected language?: string;
+  private outBuf: ArrayBuffer[];
+  constructor(manager: MicManager, name: string, language?: string) {
+    super();
+    this.manager = manager;
+    this.name = name;
+    this.language = language;
+    this.lastUtteranceEndTime = 0;
+    this.outBuf = [];
+  }
```
NickHeiner, line 170:
```suggestion
export class SpeechRecognitionBase extends EventTargetImpl {
  private lastUtteranceEndTime: number = 0;
  private outBuf: ArrayBuffer[] = [];
  protected socket?: WebSocket;

  constructor(
    protected manager: MicManager,
    protected name: string,
    protected language?: string
  ) {
    super();
  }
}
```

This can be shorter via https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties.

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+  async start() {}
+  close() {
+    if (this.socket) {
+      this.sendClose();
+      this.socket.close();
+    }
+  }
+  protected async fetchToken() {
+    const response = await fetch('/asr/api', {
```
NickHeiner, line 179: Hmm, it doesn't seem like we want to hardcode an API endpoint. What's going on here?

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+      method: 'POST',
+      body: JSON.stringify({ provider: this.name }),
+    });
+    const json = await response.json();
+    return json.token;
+  }
+  protected startInternal(url: string, protocols?: string[]) {
+    const startTime = performance.now();
+    console.log(`${this.name} socket connecting...`);
```
NickHeiner, line 188: It would be better to use AI.JSX's `logger` instance but we don't have a great way to supply that here so we can ignore it for now.

packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -0,0 +1,428 @@
...
+        `${this.name} socket closed, code=${event.code} reason=${event.reason}`,
+      );
+    };
+    this.manager.addEventListener('chunk', (event: CustomEvent) => {
+      if (this.socket!.readyState == 1) {
+        this.sendChunk(event.detail);
+      } else if (this.socket!.readyState == 0) {
+        this.outBuf.push(event.detail);
+      } else {
+        console.warn(`${this.name} socket closed`);
```
NickHeiner, line 216: Maybe this should be an error instead of a warning?

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: The key needs to be retrieved dynamically since it's an ephemeral key, but probably better to pass this in the ctor. I was trying to understand how we might want to model this with a client + server component combo, but I imagine we'll want a looser coupling than this.

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: Counterpoint: https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: arcane W3Cisms, will fix

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: I can get behind that. Probably should auto-reconnect, although it will need a new token, so I need to think about that, but... next PR

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: I tried this but got errors. It seems like EventTarget is just the interface, not an implementation. Will take another look though.

juberti: nah I'm just dumb

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: left a TODO to come back and deal with this

juberti:
OK, I think this is ready for a final look.

packages/ai-jsx/src/lib/asr/asr.tsx
juberti: Changed this to allow an arbitrary `GetToken` function to be dropped in.

PR #212, opened 2023-07-24
MDX Chat Completion

NickHeiner:
This PR adds `MdxChatCompletion`, which produces MDX output as a string. It does not handle:
* Auto-healing partial MDX (related: https://github.com/micromark/micromark-extension-mdx-jsx/issues/10)
* Rendering to the UI

I agree with @zkoch that this also might not be the final API we want.

I added docs for this, but as a standalone, this feature isn't super interesting – it gets fun when we're actually able to render to the UI. That can be a follow-up.

To verify that I didn't break the previous `UICompletion` component, I manually tested the nextjs demo at http://localhost:3000/recipe.


packages/ai-jsx/src/core/completion.tsx
```diff
@@ -189,7 +189,10 @@ export function CompletionProvider<T extends ModelPropsWithChildren>(
 }
 
 /** The default context used by {@link ChatProvider}. */
-const chatContext = AI.createContext<[ModelComponent<ModelPropsWithChildren>, ModelProps]>([AutomaticChatModel, {}]);
+export const chatContext = AI.createContext<[ModelComponent<ModelPropsWithChildren>, ModelProps]>([
```
petersalas, line 192: Why does this need to be exported?

NickHeiner: Good catch. It's a holdover from a previous iteration of this PR that's no longer necessary.

petersalas (approved):

packages/ai-jsx/src/react/jit-ui/mdx.tsx
```diff
@@ -0,0 +1,115 @@
...
+      {/* Snipped from https://github.com/mdx-js/mdx/blob/main/docs/docs/what-is-mdx.server.mdx. */}
+      MDX allows you to use JSX in your markdown content.
+      You can import components, such as interactive charts or alerts, and embed them
+      within your content.
+      This makes writing long-form content with components a blast.
+
+      More practically MDX can be explained as a format that combines markdown with
+      JSX and looks as follows:
+
+      === Begin example
```
farzadab, line 27: I'm not exactly sure how the string is formatted in JSX, but you might need a few `\n`'s here and there otherwise what the model sees is gonna be very different from what you want.

packages/ai-jsx/src/react/jit-ui/mdx.tsx
```diff
@@ -0,0 +1,115 @@
+import * as AI from '../core.js';
+import { ChatCompletion, SystemMessage } from '../../core/completion.js';
+import React from 'react';
+import { collectComponents } from '../completion.js';
+
+/**
+ * Use GPT-4 with this.
+ */
+export function MdxChatCompletion({ children, usageExamples }: { children: AI.Node; usageExamples: React.ReactNode }) {
```
farzadab, line 9: Not sure if this and `UICompletion` should be separate things (I see you were already wary of this).

farzadab (commented):
I only skimmed the PR, but left two minor comments.

packages/ai-jsx/src/react/jit-ui/mdx.tsx
NickHeiner: Yep – the model seems to understand pretty well, despite the input being garbled.

packages/ai-jsx/src/react/jit-ui/mdx.tsx
NickHeiner: Yeah, I don't feel strongly, but at least for now, I think there's value in MDX as a standalone output format (e.g. an AI-driven static site generator).

PR #214, opened 2023-07-26
Dogfooding / Demo: Seattle Sidekick

NickHeiner:
Started from ~the bottom~ https://github.com/fixie-ai/template-vercel-ai-chatbot-ai-jsx now we're here.

It looks like we need to land this on `main` before Vercel will let us deploy it.


NickHeiner:
Closing this because there are too many build issues that don't exist in the isolated repo. I'll just keep developing there.

PR #215, opened 2023-07-26
Allow UseTools to augment a broader conversation

petersalas:
This allows `<UseTools>` (specifically when using a `<ChatCompletion>` that supports function calls) to drive the function-calling loop as part of existing conversation.  That is, it now accepts a full conversation and invokes functions/appends to the conversation as the model requests.

It also adds a `showSteps` parameter which allows for streaming by also streaming the intermediate steps. Without this, we don't know whether an assistant message is terminal until it's fully rendered. (I suspect we'll iterate on the presentation of these intermediate steps shortly.)

### DIFF
--- a/packages/ai-jsx/src/batteries/use-tools.tsx
+++ b/packages/ai-jsx/src/batteries/use-tools.tsx
@@ -5,6 +5,7 @@
  */
 
 import {
+  AssistantMessage,
   ChatCompletion,
   FunctionCall,
   FunctionParameters,
@@ -12,10 +13,11 @@ import {
   SystemMessage,
   UserMessage,
 } from '../core/completion.js';
-import { Node, RenderContext, isElement, Element } from '../index.js';
+import { Node, RenderContext, isElement, Element, AppendOnlyStream, RenderableStream } from '../index.js';
 import z from 'zod';
 import { zodToJsonSchema } from 'zod-to-json-schema';
 import { AIJSXError, ErrorCode } from '../core/errors.js';
+import { memo } from '../core/memoize.js';
 
 const toolChoiceSchema = z.object({
   nameOfTool: z.string(),
@@ -24,7 +26,7 @@ const toolChoiceSchema = z.object({
 });
 export type ToolChoice = z.infer<typeof toolChoiceSchema> | null;
 
-function ChooseTools(props: Pick<UseToolsProps, 'tools' | 'userData' | 'query'>): Node {
+function ChooseTools(props: Pick<UseToolsProps, 'tools' | 'userData' | 'children'>): Node {
   return (
     <ChatCompletion>
       <SystemMessage>
@@ -43,7 +45,7 @@ function ChooseTools(props: Pick<UseToolsProps, 'tools' | 'userData' | 'query'>)
         If none of the tools seem appropriate, or the user data doesn't have the necessary context to use the tool the
         user needs, respond with `null`.
       </SystemMessage>
-      <UserMessage>Generate a JSON response for this query: {props.query}</UserMessage>
+      <UserMessage>Generate a JSON response for this query: {props.children}</UserMessage>
     </ChatCompletion>
   );
 }
@@ -127,9 +129,14 @@ export interface UseToolsProps {
   tools: Record<string, Tool>;
 
   /**
-   * A query the AI will use to decide which tool to use, and what parameters to invoke it with.
+   * The conversation in which the AI can use a tool.
    */
-  query: string;
+  children: Node;
+
+  /**
+   * Whether the result should include intermediate steps, for example, the execution of the function and its response.
+   */
+  showSteps?: boolean;
 
   /**
    * A fallback response to use if the AI doesn't think any of the tools are relevant.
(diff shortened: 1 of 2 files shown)


docs/internal/aspirational-readme.md
NickHeiner, line 1: I appreciate the thoroughness here but I think we can fully let the aspirational README rot.

That said, as long as you're in here, would you mind removing the Notion link at the top?

Maybe we should just delete this file. It's always in git if we need it.

petersalas: I can probably delete all of `docs/internal`?

packages/ai-jsx/src/batteries/use-tools.tsx
NickHeiner, line 1: Unrelated typo fix

```diff
diff --git a/packages/ai-jsx/src/batteries/use-tools.tsx b/packages/ai-jsx/src/batteries/use-tools.tsx
index 197436f..64f69af 100644
--- a/packages/ai-jsx/src/batteries/use-tools.tsx
+++ b/packages/ai-jsx/src/batteries/use-tools.tsx
@@ -89,7 +89,7 @@ async function InvokeTool(
   return (
     <ChatCompletion>
       <SystemMessage>
-        You are a tool-using agent. You previously choose to use a tool, and generated this response to the user:
+        You are a tool-using agent. You previously chose to use a tool, and generated this response to the user:
         {toolChoiceResult.responseToUser}
         When you ran the tool, you got this result: {JSON.stringify(toolResult)}
         Using the above, provide a final response to the user.
```

docs/internal/aspirational-readme.md
petersalas: Went for it

packages/ai-jsx/src/batteries/use-tools.tsx
NickHeiner, line 1: Another typo fix

```diff
diff --git a/packages/ai-jsx/src/batteries/use-tools.tsx b/packages/ai-jsx/src/batteries/use-tools.tsx
index 197436f..f9d628d 100644
--- a/packages/ai-jsx/src/batteries/use-tools.tsx
+++ b/packages/ai-jsx/src/batteries/use-tools.tsx
@@ -262,7 +262,7 @@ export async function* UseToolsFunctionCall(props: UseToolsProps, { render }: Re
         const callable = props.tools[functionCallElement.props.name].func;
         response = await callable(functionCallElement.props.args);
       } catch (e: any) {
-        response = `Function called failed with error: ${e.message}.`;
+        response = `Function call to ${functionCallElement.props.name} failed with error: ${e.message}.`;
       } finally {
         const functionResponse = memo(
           <FunctionResponse name={functionCallElement.props.name}>{response}</FunctionResponse>
```

NickHeiner (approved):

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -43,7 +45,7 @@ function ChooseTools(props: Pick<UseToolsProps, 'tools' | 'userData' | 'query'>)
         If none of the tools seem appropriate, or the user data doesn't have the necessary context to use the tool the
         user needs, respond with `null`.
       </SystemMessage>
-      <UserMessage>Generate a JSON response for this query: {props.query}</UserMessage>
+      <UserMessage>Generate a JSON response for this query: {props.children}</UserMessage>
```
farzadab, line 48: I feel like the interface for the prompt-engineered version and the function-call version have diverged. For example, if you give it a `SystemMessage`, it will be ignored. I haven't used the following, but maybe something like this would work:
```suggestion
      <UserMessage>Generate a JSON response for the query below</UserMessage>
      {props.children}
```

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -203,30 +211,31 @@ export async function* UseTools(props: UseToolsProps, { render }: RenderContext)
...
-      <SystemMessage>Here's the fallback strategy/message if something failed: {props.fallback}</SystemMessage>
-    );
-  }
-  messages.push(<UserMessage>{props.query}</UserMessage>);
+export async function* UseToolsFunctionCall(props: UseToolsProps, { render }: RenderContext): RenderableStream {
+  yield AppendOnlyStream;
 
-  do {
-    const modelResponse = <ChatCompletion functionDefinitions={props.tools}>{messages}</ChatCompletion>;
+  const conversation = [memo(props.children)];
```
farzadab, line 217: Oh nice, I like that you are memoizing it. Never thought of that. This won't render the children, right?

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -246,25 +256,40 @@ export async function* UseToolsFunctionCall(props: UseToolsProps, { render }: Re
...
+        const functionResponse = memo(
+          <FunctionResponse name={functionCallElement.props.name}>{response}</FunctionResponse>
+        );
+        if (props.showSteps) {
+          yield (
+            <>
+              {'\n'}
+              {functionResponse}
+              {'\n'}
+            </>
```
farzadab, line 276: Hmm, seems like a "presentation" issue. Maybe we should have a better way of separating the "presentation" from content (I have the same issue in multiple places), but I understand why you put this here and it's probably better to have it. Still, you can't remove the `\n`'s if you don't want them which is not good.

packages/docs/docs/guides/brand-new.md
```diff
@@ -130,8 +130,11 @@ const tools: Record<string, Tool> = {
 
 // Provide the tools to the agent
 <UseTools tools={tools} fallback="Politely explain you aren't able to help with that request.">
-  You control a home automation system. The user has requested you take some action in their home: "{userRequest}". Take
-  an action, then generate a response telling the user what you're doing.
+  <SystemMessage>
```
farzadab, line 133: Same: this `SystemMessage` will be ignored in the prompt engineered version I think. Also, the interface is weird since the full message will be:
```
Generate a JSON response for this query: You control a home automation system. The user will request an action in their home. You should take an action and then generate a response telling the user what you've done.
```

farzadab (commented):
Man you folks merge fast!

I left some comments. Love that you were already doing what I needed :)

packages/ai-jsx/src/batteries/use-tools.tsx
farzadab: You might need to memoize here too?

packages/ai-jsx/src/batteries/use-tools.tsx
farzadab: Generally I have this "presentation" issue in multiple places (e.g. Image and Audio tags). `map` is nice for modifying results, but I can't recall why I don't like it as much. Does `map` only work on the intermediate results?

PR #217, opened 2023-07-27
Move memo to RenderContext

petersalas:
This moves the `memo` function to `RenderContext`, where it will bind a `Renderable` to a single `RenderContext`. This ensures that it will only be rendered once and has symmetry with `render`.

This is a breaking change and bumps the version to `0.7.0`.

(To resolve circular import issues, it also splits `core.ts` into `node.ts` and `render.ts`.)

### DIFF
--- a/packages/ai-jsx/src/core/memoize.tsx
+++ b/packages/ai-jsx/src/core/memoize.tsx
@@ -1,60 +1,28 @@
-import * as AI from '../index.js';
-import { RenderContext, Node, Renderable } from '../index.js';
+import { Renderable, RenderContext, AppendOnlyStream } from './render.js';
+import { Node, getReferencedNode, isIndirectNode, makeIndirectNode, isElement } from './node.js';
 import { Logger } from './log.js';
 
 let memoizedId = 0;
+/** @hidden */
 export const isMemoizedSymbol = Symbol('isMemoized');
 
 /**
- * Memoize a {@link Renderable} so it always returns the same thing.
- *
- * For example, imagine you have the following:
- * ```tsx
- *    const catName = (
- *      <ChatCompletion>
- *        <UserMessage>Give me a cat name</UserMessage>
- *      </ChatCompletion>
- *    );
- *
- *    <ChatCompletion>
- *      <UserMessage>
- *        I have a cat named {catName}. Tell me a story about {catName}.
- *      </UserMessage>
- *     </ChatCompletion>
- * ```
- *
- * In this case, `catName` will result in two separate model calls, so you'll get two different cat names.
- *
- * If this is not desired, you can wrap the component in `memo`:
- * ```tsx
- *    const catName = memo(<ChatCompletion>
- *      <UserMessage>Give me a cat name</UserMessage>
- *    </ChatCompletion>);
- *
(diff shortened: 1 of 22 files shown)


NickHeiner:
Would you mind updating the rest of the docs to reflect this? For instance, https://docs.ai-jsx.com/guides/rules-of-jsx?_highlight=memo#memoization.

packages/ai-jsx/src/core/memoize.tsx
```diff
@@ -1,60 +1,28 @@
...
- *        I have a cat named {catName}. Tell me a story about {catName}.
- *      </UserMessage>
- *     </ChatCompletion>
- * ```
- * Now, `catName` will result in a single model call, and its value will be reused everywhere that component appears
- * in the tree.
- *
- * The memoization is fully recursive.
+ * "Partially" memoizes a renderable such that it will only be rendered once in any
+ * single `RenderContext`.
```
NickHeiner, line 11: Hmm, why did we make this change to the docstring?

NickHeiner: Ah, I see it was moved elsewhere – I think we should put a `@hidden` on this function.

packages/ai-jsx/src/core/memoize.tsx
NickHeiner, line 1: This comment in this file is no longer accurate, right?

```
    // N.B. The memoization applies per-RenderContext -- if the same component is rendered under
    // two different RenderContexts, it won't be memoized.
```

NickHeiner, line 1: We're not actually changing any of the logic in this file, right? Just renaming / moving things around?

packages/ai-jsx/src/core/node.ts
```diff
@@ -0,0 +1,137 @@
...
+  }) as T & IndirectNode;
+}
+
+/** @hidden */
+export function withContext(renderable: Renderable, context: RenderContext): Element<any> {
+  function SwitchContext() {
+    return renderable;
+  }
+
+  const withContext = {
```
NickHeiner, line 83: ugh, I should enable the lint rule to prevent var names from shadowing function names. :smile: 

I know this wasn't introduced in this PR.

packages/docs/docs/guides/performance.md
```diff
@@ -188,11 +188,9 @@ Use AI.JSX's built-in support for memoization instead.
 Instead, use [`memo`](/api/modules/core_memoize#memo):
 
 ```tsx
-import { memo } from 'ai-jsx/core/memoize';
-
-function StoryGenerator(props, { render }) {
+function StoryGenerator(props, { memo }) {
   // highlight-next-line
-  const heroName = await memo(<CharacterName role="hero" />);
+  const heroName = memo(<CharacterName role="hero" />);
```
NickHeiner, line 193: This `await` was always unnecessary, right?

NickHeiner (approved):
Glad to remove this footgun!

packages/docs/docs/guides/performance.md
petersalas: Yup

PR #218, opened 2023-07-28
Change ChatCompletion to emit AssistantMessage

petersalas:
This changes the various `ChatCompletion` implementations to emit `<AssistantMessage>`s rather than loose text. This allows callers to easily separate `<AssistantMessage>` and `<FunctionCall>` outputs.

Preserving the append-only nature of the stream requires non-trivial gymnastics to split the iterator consumption across multiple components, but this technique also allows for the removal of the `experimental_streamFunctionCallOnly` flag.

### DIFF
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -338,9 +338,9 @@ export async function* JsonChatCompletionFunctionCall(
 
   const frames = render(childrenWithCompletion, { stop: (e) => e.tag === FunctionCall, map: (e) => e });
   for await (const frame of frames) {
-    const functionCall = frame.find((e) => AI.isElement(e) && e.tag === FunctionCall) as AI.Element<
-      AI.PropsOfComponent<typeof FunctionCall>
-    >;
+    const functionCall = frame.find((e) => AI.isElement(e) && e.tag === FunctionCall) as
+      | AI.Element<AI.PropsOfComponent<typeof FunctionCall>>
+      | undefined;
     if (!functionCall) {
       continue;
     }
@@ -356,11 +356,15 @@ export async function* JsonChatCompletionFunctionCall(
     yield JSON.stringify(object);
   }
 
-  const functionCall = (await frames).find((e) => AI.isElement(e) && e.tag === FunctionCall) as AI.Element<
-    AI.PropsOfComponent<typeof FunctionCall>
-  >;
+  const functionCall = (await frames).find((e) => AI.isElement(e) && e.tag === FunctionCall) as
+    | AI.Element<AI.PropsOfComponent<typeof FunctionCall>>
+    | undefined;
 
-  const object = functionCall?.props.args;
+  if (functionCall === undefined) {
+    return null;
+  }
+
+  const object = functionCall.props.args;
   try {
     for (const validator of validatorsAndSchema) {
       validator(object);
--- a/packages/ai-jsx/src/batteries/constrained-output.tsx
+++ b/packages/ai-jsx/src/batteries/constrained-output.tsx
@@ -11,6 +11,7 @@ import { zodToJsonSchema } from 'zod-to-json-schema';
 import {
   AssistantMessage,
   ChatCompletion,
+  FunctionCall,
   ModelPropsWithChildren,
   SystemMessage,
   UserMessage,
@@ -318,7 +319,6 @@ export async function* JsonChatCompletionFunctionCall(
 
   const childrenWithCompletion = (
     <ChatCompletion
-      experimental_streamFunctionCallOnly
       {...props}
       functionDefinitions={{
         print: {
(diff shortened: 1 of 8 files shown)


packages/ai-jsx/src/batteries/constrained-output.tsx
```diff
@@ -336,9 +336,16 @@ export async function* JsonChatCompletionFunctionCall(
...
   for await (const frame of frames) {
-    const object = JSON.parse(frame).arguments;
+    const functionCall = frame.find((e) => AI.isElement(e) && e.tag === FunctionCall) as
+      | AI.Element<AI.PropsOfComponent<typeof FunctionCall>>
+      | undefined;
+    if (!functionCall) {
+      continue;
+    }
+
+    const object = functionCall.props.args;
```
farzadab, line 348: I forgot to apply this comment from Nick on previous PR, can you rename `object` to `jsonResult` too please?
```suggestion
    const jsonResult = functionCall.props.args;
```

packages/ai-jsx/src/lib/replicate-llama2.tsx
```diff
@@ -30,7 +30,6 @@ async function* fetchLlama2<ModelArgs extends Llama2ModelArgs>(
   const output = (await replicate.run(modelId, { input })) as string[];
   const result = output.join('');
   logger.debug({ result }, 'Replicate llama2 output');
-  yield result;
```
NickHeiner, line 33: Was this always unnecessary?

petersalas: yup

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -438,59 +422,97 @@ export async function* OpenAIChatModel(
...
+            delta = await advance();
+          }
+
+          return AI.AppendOnlyStream;
+        })()
+      );
+      yield <AssistantMessage>{assistantStream}</AssistantMessage>;
+
+      // Ensure the assistantStream() is flushed by rendering it.
+      await render(assistantStream);
```
farzadab, line 471: Q: since this is a `memo` object, rendering it (without checking the output) is enough?
Otherwise this piece of code is confusing.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -438,59 +422,97 @@ export async function* OpenAIChatModel(
...
+          yield AI.AppendOnlyStream;
+
+          while (delta !== null) {
+            if (delta.content) {
+              yield delta.content;
+            }
+            if (delta.function_call) {
+              break;
+            }
+            delta = await advance();
```
farzadab, line 462: I'm confused. So the `advance` function is shared between the two streams, plus the scope above? It's the same data stream, right? So how do we now which one calls it when? How is there no race-condition? Like if you `await` it and it turns out to be a `function_call`, doesn't that mean the other streams are "not" going to get this `delta` and we just threw out the data that was inside that `delta`?

NickHeiner (approved):

farzadab (commented):
Unsurprisingly I understand only 30% of the code at best, so I'll just comment instead of approve/deny.

So the idea is that since all the layers are `appendOnly` then we can render it as `appendOnly`? And this won't cause issues with NextJS stream now allowing multiple streams?

I think this also solves my issue with the `Audio` tag, since you can stream the byte data from inside the tag! Amazing!

packages/ai-jsx/src/lib/openai.tsx
petersalas: Yeah it's more than a little sketchy:

- `memo` ensures that side-effects of the generator will be consumed _at most_ once
- `yield`ing the generator inside a component does not guarantee that the generator will be executed (it will only be executed if the yielded value is actually rendered)
- `render`ing the memoized generator guarantees that the generator will be consumed _at least_ once

packages/ai-jsx/src/lib/openai.tsx
petersalas: The (dark) magic is the `memo`/`yield`/`await render` sequence which acts like a memory barrier in that the code below it will not execute until the generators above it have been flushed once. (This ensures that the execution-order of the code is top-down.)

There are two side-effects these closure-generators have:
- Calling `advance` mutates the shared iterator
- Assigning to `delta`, which is a shared `let` binding. (Note that each closure-generator starts by looking at the existing value of `delta`, rather than starting with an `advance`.)

packages/ai-jsx/src/lib/replicate-llama2.tsx
NickHeiner: SSSSSSSHHHHHHHHHHIIIIIIIIIIIIIIIIIII

petersalas:
> So the idea is that since all the layers are appendOnly then we can render it as appendOnly?

Yup. More specifically the `<ChatCompletion>` is append-only, and the emitted `<AssistantMessage>` is append-only, but the stream that emits `<FunctionCall>`s is _not_ append-only (it will only appear once the JSON is fully resolved).

> And this won't cause issues with NextJS stream now allowing multiple streams?

The multiple stream limitation is specific to having nested `React -> AI.JSX -> React -> AI.JSX` layering.

> I think this also solves my issue with the Audio tag, since you can stream the byte data from inside the tag! Amazing!

I think an approach like this would work. (Though there are likely other challenges that will emerge as well :))

PR #221, opened 2023-07-28
Add `Converse` and `ShowConversation` components

petersalas:
image

- Move conversational components (`UserMessage` et al) to a separate file
- Add `isAppendOnlyRender` to `ComponentContext` to allow components to optimize their stream based on whether they're part of an append-only stream.
- Add `renderToConversation`, `Converse`, and `ShowConversation` helpers/components to facilitate conversational UI

The `UseTools` example demonstrates using `ShowConversation` to manage the presentation.

### DIFF
--- a/packages/ai-jsx/src/batteries/use-tools.tsx
+++ b/packages/ai-jsx/src/batteries/use-tools.tsx
@@ -5,26 +5,17 @@
  */
 
 import {
-  AssistantMessage,
   ChatCompletion,
-  FunctionCall,
   FunctionParameters,
   FunctionResponse,
   SystemMessage,
   UserMessage,
 } from '../core/completion.js';
-import {
-  Node,
-  RenderContext,
-  isElement,
-  Element,
-  AppendOnlyStream,
-  ComponentContext,
-  RenderableStream,
-} from '../index.js';
+import { Node, RenderContext, ComponentContext } from '../index.js';
 import z from 'zod';
 import { zodToJsonSchema } from 'zod-to-json-schema';
 import { AIJSXError, ErrorCode } from '../core/errors.js';
+import { Converse, isConversationalComponent, renderToConversation } from '../core/conversation.js';
 
 const toolChoiceSchema = z.object({
   nameOfTool: z.string(),
@@ -208,7 +199,11 @@ export interface UseToolsProps {
  */
 export async function* UseTools(props: UseToolsProps, { render }: RenderContext) {
   try {
-    const rendered = yield* render(<UseToolsFunctionCall {...props} />);
+    const rendered = yield* render(<UseToolsFunctionCall {...props} />, {
+      // TODO: ErrorBoundaries should be able to preserve the conversational structure, but they can't currently.
+      stop: isConversationalComponent,
+      map: (x) => x,
+    });
     return rendered;
   } catch (e: any) {
     if (e.code === ErrorCode.ChatModelDoesNotSupportFunctions) {
@@ -219,83 +214,46 @@ export async function* UseTools(props: UseToolsProps, { render }: RenderContext)
 }
 
 /** @hidden */
-export async function* UseToolsFunctionCall(
-  props: UseToolsProps,
-  { render, memo, logger }: ComponentContext
-): RenderableStream {
-  yield AppendOnlyStream;
-
-  const conversation = [memo(props.children)];
-
-  do {
-    const modelResponse = memo(<ChatCompletion functionDefinitions={props.tools}>{conversation}</ChatCompletion>);
-    if (props.showSteps) {
-      yield modelResponse;
-    }
-
-    const renderResult = await render(modelResponse, {
-      stop: (el) => el.tag === AssistantMessage || el.tag == FunctionCall,
-    });
-    let functionCallElement: Element<any> | null = null;
-
-    for (const element of renderResult) {
-      if (isElement(element)) {
-        conversation.push(memo(element));
-
-        if (element.tag === FunctionCall) {
-          // Model has generated a function call.
-          if (functionCallElement) {
-            throw new AIJSXError(
-              `ChatCompletion returned 2 function calls at the same time ${renderResult.join(', ')}`,
-              ErrorCode.ModelOutputCouldNotBeParsedForTool,
-              'runtime'
-            );
+export async function UseToolsFunctionCall(props: UseToolsProps, { render }: ComponentContext) {
+  const converse = (
+    <Converse
+      reply={async (messages, fullConversation) => {
+        const lastMessage = messages[messages.length - 1];
+        switch (lastMessage.type) {
+          case 'functionCall': {
+            const { name, args } = lastMessage.element.props;
+            try {
+              return <FunctionResponse name={name}>{await props.tools[name].func(args)}</FunctionResponse>;
+            } catch (e: any) {
+              return (
+                <FunctionResponse name={name}>
+                  Function call to {name} failed with error: {e.message}.
+                </FunctionResponse>
+              );
+            }
           }
-          functionCallElement = element;
+          case 'functionResponse':
+          case 'user':
+            return (
+              <ChatCompletion functionDefinitions={props.tools}>
+                {fullConversation.map((m) => m.element)}
+              </ChatCompletion>
+            );
+          default:
+            return null;
         }
-      } else {
-        logger.debug(
-          { text: element },
-          '<ChatCompletion> emitted something other than <AssistantMessage> or <FunctionCall>, which is unexpected.'
-        );
-      }
-    }
+      }}
+    >
+      {props.children}
+    </Converse>
+  );
 
-    if (functionCallElement) {
-      // Call the selected function and append the result to the messages.
-      let response;
-      try {
-        const callable = props.tools[functionCallElement.props.name].func;
-        response = await callable(functionCallElement.props.args);
-      } catch (e: any) {
-        response = `Function call to ${functionCallElement.props.name} failed with error: ${e.message}.`;
-      } finally {
-        const functionResponse = memo(
-          <FunctionResponse name={functionCallElement.props.name}>{response}</FunctionResponse>
-        );
-        if (props.showSteps) {
-          yield (
-            <>
-              {'\n'}
-              {functionResponse}
-              {'\n'}
-            </>
-          );
-        }
-        conversation.push(functionResponse);
-      }
+  if (props.showSteps) {
+    return converse;
+  }
 
-      continue;
-    }
-
-    // Terminate the loop when the model produces a response without a function call.
-    if (!props.showSteps) {
-      yield modelResponse;
-    }
-    break;
-  } while (true);
-
-  return AppendOnlyStream;
+  const messages = await renderToConversation(converse, render);
+  return messages.length && messages[messages.length - 1].element;
 }
 
 /** @hidden */
--- a/packages/ai-jsx/src/batteries/use-tools.tsx
+++ b/packages/ai-jsx/src/batteries/use-tools.tsx
@@ -11,7 +11,7 @@ import {
   SystemMessage,
   UserMessage,
 } from '../core/completion.js';
-import { Node, RenderContext, ComponentContext } from '../index.js';
+import { Node, RenderContext, ComponentContext, isElement } from '../index.js';
 import z from 'zod';
 import { zodToJsonSchema } from 'zod-to-json-schema';
 import { AIJSXError, ErrorCode } from '../core/errors.js';
@@ -197,14 +197,21 @@ export interface UseToolsProps {
  * ```
  *
  */
-export async function* UseTools(props: UseToolsProps, { render }: RenderContext) {
+export async function UseTools(props: UseToolsProps, { render, memo }: RenderContext) {
   try {
-    const rendered = yield* render(<UseToolsFunctionCall {...props} />, {
-      // TODO: ErrorBoundaries should be able to preserve the conversational structure, but they can't currently.
+    // TODO: ErrorBoundaries should be able to preserve the conversational structure, but they can't currently.
+    // So instead we start rendering the function call until it yields any conversational message. If we see one,
+    // we know that the <ChatCompletion> didn't immediately fail.
+    const memoizedFunctionCall = memo(<UseToolsFunctionCall {...props} />);
+    for await (const containsElement of render(memoizedFunctionCall, {
       stop: isConversationalComponent,
-      map: (x) => x,
-    });
-    return rendered;
+      map: (frame) => frame.find(isElement) !== undefined,
+    })) {
+      if (containsElement) {
+        break;
+      }
+    }
+    return memoizedFunctionCall;
   } catch (e: any) {
     if (e.code === ErrorCode.ChatModelDoesNotSupportFunctions) {
       return <UseToolsPromptEngineered {...props} />;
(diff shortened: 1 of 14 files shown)


packages/ai-jsx/src/core/conversation.tsx
NickHeiner, line 1: Looks like large parts of this are extracted unchanged from `completion`, right?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+  | ConversationMessageType<'user', typeof UserMessage>
+  | ConversationMessageType<'assistant', typeof AssistantMessage>
+  | ConversationMessageType<'system', typeof SystemMessage>
+  | ConversationMessageType<'functionCall', typeof FunctionCall>
+  | ConversationMessageType<'functionResponse', typeof FunctionResponse>;
+
+/** @hidden */
+export function isConversationalComponent(element: AI.Element<any>): boolean {
+  return (
+    element.tag === UserMessage ||
```
NickHeiner, line 148: This can also be done with something like `[UserMessage, /* ... */].includes(element.tag)`,

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+          );
+      }
+    });
+}
+
+/** @hidden */
+export async function renderToConversation(
+  conversation: AI.Node,
+  render: AI.ComponentContext['render']
+): Promise<ConversationMessage[]> {
```
NickHeiner, line 185:
```suggestion
) {
```

Is this type annotation necessary?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+ *
+ * The `reply` function is invoked with the new messages produced from its _own replies_ until
+ * it fails to return a conversational message (e.g. UserMessage or AssistantMessage). The first
+ * invocation includes the messages from the `children` prop.
+ *
+ * @example
+ * ```tsx
+ *    <Converse reply={(messages) => messages[0].type === "user" ? <AssistantMessage>Hello there!</AssistantMessage> : null}>
+ *      <UserMessage>Hello!</UserMessage>
+ *    </ChatCompletion>
```
NickHeiner, line 200: I appreciate the simplicity, but I think this might be a bit of a confusing example, because it seems pretty unlikely that a user would directly render an `AssistantMessage` they generate themselves, right?

Maybe a better example would be driving an automatic user response based on looking at what the previous message was? Like, the pattern where the AI emits a response, and the user responds "are you sure you followed all the instructions?" or "please critique your own response and amend it as necessary".

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+    children: AI.Node;
+  },
+  { render, memo }: AI.ComponentContext
+): AI.RenderableStream {
+  yield AI.AppendOnlyStream;
+
+  let fullConversation = [] as ConversationMessage[];
+  let next = memo(children);
+  while (true) {
+    const messages = await renderToConversation(next, render);
```
NickHeiner, line 221:
```suggestion
    const newMessages = await renderToConversation(next, render);
```

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+}
+
+/**
+ * Allows the presentation of conversational components ({@link UserMessage} et al) to be altered.
+ *
+ * Also accepts an `onComplete` prop, which will be invoked once per render with the entire conversation.
+ *
+ * @example
+ * ```tsx
+ *     <ShowConversation present={(msg) => msg.type === "assistant" && <>Assistant: {msg.element}</>}>
```
NickHeiner, line 241: Hmm, I'm not fully convinced that `present` belongs at this layer, as opposed to just giving a full JSON to the caller, and letting the UI decide what to show.

I guess there's a case where we might not want to return function call/responses to the client, and this is an easier place to enforce that than modifying the final output stream itself?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -444,9 +428,14 @@ export async function* OpenAIChatModel(
     return next.value.choices[0].delta;
   }
 
+  let isAssistant = false;
```
NickHeiner, line 431: Why is this now necessary?

NickHeiner:
This basically LGTM.

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -219,83 +214,46 @@ export async function* UseTools(props: UseToolsProps, { render }: RenderContext)
...
+export async function UseToolsFunctionCall(props: UseToolsProps, { render }: ComponentContext) {
+  const converse = (
+    <Converse
+      reply={async (messages, fullConversation) => {
+        const lastMessage = messages[messages.length - 1];
+        switch (lastMessage.type) {
+          case 'functionCall': {
+            const { name, args } = lastMessage.element.props;
+            try {
+              return <FunctionResponse name={name}>{await props.tools[name].func(args)}</FunctionResponse>;
```
farzadab, line 226: Not related to PR, but shouldn't `FunctionResponse` have a `failed` flag so that you can tell if it has failed without checking the children?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+      }
+    });
+}
+
+/** @hidden */
+export async function renderToConversation(
+  conversation: AI.Node,
+  render: AI.ComponentContext['render']
+): Promise<ConversationMessage[]> {
+  return toConversationMessages(await render(conversation, { stop: isConversationalComponent }));
```
farzadab, line 186: It's possible that this returns something that's not a `ConversationMessage` (e.g. a plain string), right?

packages/ai-jsx/src/core/conversation.tsx
petersalas: Yeah the tricky part (as usual) is streaming (and/or potentially embedded UI), which is why these transformations are in tree-space rather than JSON- or text- space.

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+  // Memoize before rendering so that the all the conversational components get memoized as well.
+  const finalFrame = yield* render(memo(children), {
+    map: handleFrame,
+    stop: isConversationalComponent,
+    appendOnly: isAppendOnlyRender,
+  });
+
+  // Prioritize rendering by yielding the final frame before running the `onComplete` handler.
+  yield handleFrame(finalFrame);
```
farzadab, line 289: I feel like we have this issue of wanting `map` to apply to every frame, not just the intermediate ones in multiple places.

packages/ai-jsx/src/core/conversation.tsx
petersalas: Yeah I was trying to come up with the simplest example that demonstrates the semantics of `<Converse>` which might be more confusing than just showing an example with `<ChatCompletion>`. Will rework this

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+    conversationMessages = toConversationMessages(frame);
+    return toConversationMessages(frame).map(present ?? ((m) => m.element));
+  }
+
+  // Memoize before rendering so that the all the conversational components get memoized as well.
+  const finalFrame = yield* render(memo(children), {
+    map: handleFrame,
+    stop: isConversationalComponent,
+    appendOnly: isAppendOnlyRender,
```
farzadab, line 285: Oh shoot, does that mean from now on we have to manually pass `appendOnly` to render? If not, then we're altering the expected behaviour which might not be okay (and previous code wasn't built with this in mind). Is there a way to pass it anyway if not provided?

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -208,7 +199,11 @@ export interface UseToolsProps {
  */
 export async function* UseTools(props: UseToolsProps, { render }: RenderContext) {
   try {
-    const rendered = yield* render(<UseToolsFunctionCall {...props} />);
+    const rendered = yield* render(<UseToolsFunctionCall {...props} />, {
```
petersalas, line 202: I'm pretty sure this can lead to flickering, need to rework this (and/or punt until better error boundaries)

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -0,0 +1,295 @@
...
+  });
+
+  // Prioritize rendering by yielding the final frame before running the `onComplete` handler.
+  yield handleFrame(finalFrame);
+  await onComplete?.(conversationMessages, render);
+
+  // N.B. This may not have been an append-only stream until now, but by returning this
+  // we can indicate that we've already yielded the final frame.
+  return AI.AppendOnlyStream;
```
farzadab, line 294: Nit: This seems weird, but I don't know the details, so it's fine I guess.

packages/examples/src/use-tools.tsx
```diff
@@ -22,9 +22,39 @@ function App({ query }: { query: string }) {
...
+              <>
+                Function Result: {msg.element}
+                {'\n'}
+              </>
+            );
+          default:
+            return null;
+        }
+      }}
+    >
```
farzadab, line 53: I like the flexibility, but the example does seem way more scary now, plus, the main logic is buried deep which might be confusing.

farzadab (approved):
Seems like a huge improvement, thanks!
I left some comments. Mainly I'm concerned with `isAppendOnly` not getting passed by default which makes it the burden of the developer.

packages/examples/src/use-tools.tsx
farzadab: Also, I haven't run this but IIRC FunctionCall and FunctionResponse already print some commentary, so you might get a result like: `Function Call: Calling function X with args Y` which is redundant. One way to make this less scary is to get a `mapping` instead of a function. Assuming people always want to do a switch, this makes it easier to handle. Plus, a map can give you some defaults (e.g. if I only want to change `FunctionResponse`) instead of having to either not use `ShowConversation` vs handling all cases.

packages/ai-jsx/src/core/conversation.tsx
petersalas: Not at all -- this component is just trying to optimize for its behavior for either scenario. If it's being rendered in append-only then it presents itself as append-only, but if not, it doesn't (e.g. allow `<FunctionCall partial>` values to surface).

packages/ai-jsx/src/core/conversation.tsx
petersalas: Yeah, these `map` scenarios are very awkward. I've delayed changing the API because my intuition is roughly "we shouldn't need to be doing this" but the better option hasn't really emerged yet so I should probably just fix it.

packages/ai-jsx/src/core/conversation.tsx
petersalas: `render` can but `toConversationMessages` cannot because it explicitly filters out loose text.

farzadab: You're right, sorry. Why didn't I see that.

packages/ai-jsx/src/core/conversation.tsx
petersalas: If a component presents _itself_ as append-only by yielding `AI.AppendOnlyStream` then it must only `yield` values _to be appended_.

And if a component presents itself as append-only but wants to yield intermediate values from an explicit `render` then it can only maintain the invariant by also requesting that the explicit `render` _also_ be append-only.

So, because this component conditionally presents itself as append-only, it must also conditionally make the child render append-only. (It's also currently the only component that does this wacky conditional append-only. If we find ourselves doing this in several places we should almost certainly change the abstraction.)

packages/ai-jsx/src/batteries/use-tools.tsx
petersalas: Yeah that seems like a good idea.

packages/examples/src/use-tools.tsx
petersalas: I started with the mapping approach but the object-of-functions seemed awkward and I didn't think the existing `<FunctionCall>`/`<FunctionResponse>` values were reasonable defaults. But maybe if we're good with those being silent by default that would make sense? WDYT?

Re: the demo:
- Showing the intermediate steps is IMO pretty informative
- I wasn't sure how to make the demo legible _without_ `<ShowConversation>` (recall that I used to have the "\n" hacks to work around this)

packages/ai-jsx/src/lib/openai.tsx
petersalas: I discovered that previously we were emitting:

```tsx
<AssistantMessage></AssistantMessage>
<FunctionCall name="foo" />
```

So I added this check to wait for any content tagged with the assistant role.

packages/ai-jsx/src/core/conversation.tsx
petersalas: I ended up doing something along the lines of the original motivation, which is user-managed control over the function calling loop. (Which is on its own a little lame because it's roughly what `<UseToolsFunctionCall>` already does. But right now my intuition is that we'll ultimately want to delegate more control of the loop to user code rather than having `<UseTools>` as a one-size-fits-all.)

packages/ai-jsx/src/core/conversation.tsx
petersalas: Per our conversation going to leave this for now. As our approach to UI/RSC/embedded React/etc. crystalizes I'm totally open to reworking this.

packages/ai-jsx/src/core/conversation.tsx
petersalas: Yeah, in general this component is weird, and in particular the conditional `AppendOnlyStream` is _really_ weird. (My **hope** is that the ugliness/badness/weirdness of this component's implementation can prevent other components from needing to do similarly weird things.)

packages/examples/src/use-tools.tsx
petersalas: I'm going to resolve/merge to be able to start consuming this elsewhere but I'm not particularly wedded to the precise API and/or what's in the example.

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -145,12 +153,8 @@ export type ConversationMessage =
 /** @hidden */
 export function isConversationalComponent(element: AI.Element<any>): boolean {
   return (
-    element.tag === UserMessage ||
-    element.tag === AssistantMessage ||
-    element.tag === SystemMessage ||
-    element.tag === FunctionCall ||
-    element.tag === FunctionResponse
-  );
+    [UserMessage, AssistantMessage, SystemMessage, FunctionCall, FunctionResponse] as AI.Component<any>[]
```
NickHeiner, line 156: `as const` may also have worked here.

packages/ai-jsx/src/core/completion.tsx
```diff
@@ -12,6 +11,14 @@ import { getEnvVar } from '../lib/util.js';
...
 import z from 'zod';
 import { zodToJsonSchema } from 'zod-to-json-schema';
+export {
+  UserMessage,
+  SystemMessage,
+  AssistantMessage,
+  FunctionCall,
+  FunctionResponse,
+  ConversationHistory,
+} from './conversation.js';
```
NickHeiner, line 21: Presumably you're doing this to avoid breaking backwards compat, but also think that `conversation.js` is the more natural home for these? In that case, we should mark these as `@deprecated` and give a note telling people to use `conversation` instead.

PR #223, opened 2023-07-31
Throw an error when the user puts bare strings in a chat completion

NickHeiner:


packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -333,6 +333,22 @@ export async function* OpenAIChatModel(
       e.tag == FunctionResponse,
   });
 
+  logger.warn({ messageElements });
+
+  const invalidChildren = messageElements.filter((el) => typeof el === 'string' && el.trim()) as string[];
```
petersalas, line 338: Per the merge conflict this will need to move to `renderToConversation`

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -160,28 +160,40 @@ export function isConversationalComponent(element: AI.Element<any>): boolean {
...
-          return { type: 'functionCall', element: e };
-        case FunctionResponse:
-          return { type: 'functionResponse', element: e };
-        default:
-          throw new AIJSXError(
-            `Unexpected tag (${e.tag.name}) in conversation`,
-            ErrorCode.UnexpectedPartialRenderResult,
-            'internal'
-          );
+  const invalidChildren = partialRendering.filter((el) => typeof el === 'string' && el.trim()) as string[];
```
NickHeiner, line 163: @petersalas I moved this check to `toConversationMessages` rather than `openai` – does that seem right?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -394,6 +406,20 @@ export async function ShrinkConversation(
       stop: (e) => isConversationalComponent(e) || e.tag === InternalShrinkable,
     });
 
+    const invalidChildren = rendered.filter((el) => typeof el === 'string' && el.trim()) as string[];
```
NickHeiner, line 409: @petersalas I found that I needed to repeat the check here – can you think of anywhere else I should do it?

packages/examples/src/conversation-shrinking.tsx
```diff
@@ -13,7 +13,7 @@ function App() {
     }
 
     messages.push(
-      <Shrinkable importance={0}>
+      <Shrinkable importance={0} replacement={counting[counting.length - 1]}>
```
NickHeiner, line 16: This is just for testing and will be removed.

packages/ai-jsx/src/core/conversation.tsx
petersalas: I think so, yeah

packages/ai-jsx/src/core/conversation.tsx
petersalas: Sigh, yeah. I don't think so -- this mixed conversation + shrinkable thing felt icky when I did it because I don't think I did it previously.

petersalas (approved):
I suspect this breaks the debate demo, but I think it can be killed and/or re-written with `Converse`/`ShowConversation`.

NickHeiner:
> I suspect this breaks the debate demo, but I think it can be killed and/or re-written with `Converse`/`ShowConversation`.

It still works for me locally.

PR #224, opened 2023-07-31
New Render API

NickHeiner:
Our current Render API has a few challenges.

1. We need to list [some constraints](https://docs.ai-jsx.com/guides/rendering#constraints) that may not be immediately obvious to people.
2. Some people are less familiar with generator semantics.
3. Is it weird to have a value that's both a promise and an async generator?
4. `AI.createRenderContext().render` is odd because wtf is a render context?
5. Other output methods (e.g. `toTextStream`, `toReactStream`, etc) are scattered.
6. If you want deltas, you need to explicitly do the `frame.slice(lastValue.length)` pattern.

To address this, I started sketching a new API that wraps the old one.

Downsides of the new approach:
1. With `AI.createRenderContext().render`, if you want to render within a component, it's more natural to use `componentContext.render`, rather than reimporting the top-level API. If I simplify that API, that natural barrier erodes.

I think this may be valuable as the top-level render API. For the "within components" rendering – maybe? If we move forward with this, I suspect we'd wrap the current top-level render API, rather than trying to use this for all render internals.

### DIFF
--- a/packages/ai-jsx/src/lib/openai.tsx
+++ b/packages/ai-jsx/src/lib/openai.tsx
@@ -243,6 +243,8 @@ async function checkOpenAIResponse<M extends OpenAIMethod>(response: Response, l
   }
 }
 
+let iteration = 0;
+
 /**
  * Represents an OpenAI text completion model (e.g., `text-davinci-003`).
  */
@@ -410,6 +412,13 @@ export async function* OpenAIChatModel(
     stream: true,
   };
 
+  iteration++;
+  console.log('yielding');
+  yield `first ${iteration} `
+  yield `second ${iteration} `
+  yield `third ${iteration}`
+  return AI.AppendOnlyStream;
+
   logger.debug({ chatCompletionRequest }, 'Calling createChatCompletion');
   const chatResponse = await openai.createChatCompletion(chatCompletionRequest);
 
@@ -446,6 +455,7 @@ export async function* OpenAIChatModel(
 
   let delta = await advance();
   while (delta !== null) {
+    // @ts-ignore
     if (delta.role === 'assistant') {
       // Memoize the stream to ensure it renders only once.
       const assistantStream = memo(
--- a/packages/ai-jsx/src/lib/openai.tsx
+++ b/packages/ai-jsx/src/lib/openai.tsx
@@ -243,8 +243,6 @@ async function checkOpenAIResponse<M extends OpenAIMethod>(response: Response, l
   }
 }
 
-let iteration = 0;
-
 /**
  * Represents an OpenAI text completion model (e.g., `text-davinci-003`).
  */
@@ -412,28 +410,6 @@ export async function* OpenAIChatModel(
     stream: true,
   };
 
-  function sleep() {
-    // return new Promise(resolve => setTimeout(resolve, 500));
-
-    // return new Promise<void>((resolve) => {
-    //   console.log('timeout start');
-    //   setTimeout(() => {
-    //     console.log('timeout done');
-    //     resolve();
-    //   }, 1000);
-    // })
-
-    return Promise.resolve();
-  }
-  // iteration++;
-  // await sleep();
-  // yield `first ${iteration} `
-  // await sleep();
-  // yield `second ${iteration} `
-  // await sleep();
-  // yield `third ${iteration}`
-  // return AI.AppendOnlyStream;
-
   logger.debug({ chatCompletionRequest }, 'Calling createChatCompletion');
   const chatResponse = await openai.createChatCompletion(chatCompletionRequest);
 
@@ -470,7 +446,6 @@ export async function* OpenAIChatModel(
 
   let delta = await advance();
   while (delta !== null) {
-    // @ts-ignore
     if (delta.role === 'assistant') {
       // Memoize the stream to ensure it renders only once.
       const assistantStream = memo(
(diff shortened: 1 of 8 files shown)


packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -410,6 +412,28 @@ export async function* OpenAIChatModel(
...
+  function sleep() {
+    // return new Promise(resolve => setTimeout(resolve, 500));
+
+    // return new Promise<void>((resolve) => {
+    //   console.log('timeout start');
+    //   setTimeout(() => {
+    //     console.log('timeout done');
+    //     resolve();
+    //   }, 1000);
+    // })
```
NickHeiner, line 424: @petersalas when I did this, the whole app hung. This surprised me. Maybe we can sync and work through together why that may be happening.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -410,6 +412,28 @@ export async function* OpenAIChatModel(
...
+    return Promise.resolve();
+  }
+  iteration++;
+  await sleep();
+  yield `first ${iteration} `
+  await sleep();
+  yield `second ${iteration} `
+  await sleep();
+  yield `third ${iteration}`
+  return AI.AppendOnlyStream;
```
NickHeiner, line 435: This is because I was working on this on the plane with no wifi.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -446,6 +470,7 @@ export async function* OpenAIChatModel(
 
   let delta = await advance();
   while (delta !== null) {
+    // @ts-ignore
```
NickHeiner, line 473: This will be removed.

packages/examples/src/simple-chat.tsx
NickHeiner, line 1: This is just a hack to show the idea. If we move forward, I'll clean it up / move it somewhere more sensible.

packages/ai-jsx/src/lib/openai.tsx
NickHeiner: nm I figured it out

packages/examples/src/simple-chat.tsx
```diff
@@ -10,4 +13,157 @@ function App() {
...
+const rendered = render(renderable);
+
+console.log('=== First Render ===');
+// You can consume a stream multiple times.
+console.log(
+  await Promise.all([
+    streamToValues(rendered.treeStream()),
+    streamToValues(rendered.treeStream()),
+    streamToValues(rendered.deltaStream()),
+  ])
```
petersalas, line 142: What's the scenario where one wants to consume the same stream in multiple formats/multiple times? That feels like it should inform what the right semantics are, i.e. should they operate independently or in sync?

packages/examples/src/simple-chat.tsx
```diff
@@ -10,4 +13,157 @@ function App() {
...
+);
+// If you consume a stream after the render is complete, you just get one chunk with the final result.
+console.log(await streamToValues(rendered.treeStream()));
+// console.log('Deltas', await streamToValues(rendered.deltaStream()));
+// console.log('Final result:', await rendered.result);
+
+console.log();
+console.log('=== Second Render ===');
+// Pass a single map function, and it's applied to intermediate and the final results.
```
petersalas, line 151: This could (should?) just be a change to the existing `render`?

packages/examples/src/simple-chat.tsx
```diff
@@ -10,4 +13,157 @@ function App() {
...
+console.log('=== First Render ===');
+// You can consume a stream multiple times.
+console.log(
+  await Promise.all([
+    streamToValues(rendered.treeStream()),
+    streamToValues(rendered.treeStream()),
+    streamToValues(rendered.deltaStream()),
+  ])
+);
+// If you consume a stream after the render is complete, you just get one chunk with the final result.
```
petersalas, line 144: This could also just be a change to the existing `render`?

petersalas (commented):
This seems like a mix of:
1. Grouping what are effectively existing "entry-point" type APIs
2. Tweaking of semantics for the existing `render`
3. Facilitating consumption of the same render stream in multiple formats

But 1 and 2 feel like they can be done directly (e.g. refactoring and/or organizing imports, changing the existing implementation), and I'm not sure what the motivation for 3 is.

At a high level, I agree there are pain points here, but I'd rather solve them more directly than by adding another layer on top of the existing stuff, if that makes sense.

NickHeiner:
> This seems like a mix of:
> 
> 1. Grouping what are effectively existing "entry-point" type APIs
> 2. Tweaking of semantics for the existing `render`
> 3. Facilitating consumption of the same render stream in multiple formats
> 
> But 1 and 2 feel like they can be done directly (e.g. refactoring and/or organizing imports, changing the existing implementation), and I'm not sure what the motivation for 3 is.
> 
> At a high level, I agree there are pain points here, but I'd rather solve them more directly than by adding another layer on top of the existing stuff, if that makes sense.

I'd be happy to address some of the issues by changing the existing render semantics. I did this separately in part to make it easier to demo, but also because I wasn't sure if changing the existing semantics would break some other cases. (Like, is there a reason that `map` only applies to intermediate results?)

For (3), I don't expect it to be a common case, but I also felt like it needed to be called out in the docs as footguns, so I would just as soon avoid it.

PR #226, opened 2023-08-01
ai.jsx Hello World Fixie Agent and fastify shim

juberti:
Right now the shim is hard-coded to invoke the agent as I was having troubles getting dynamic imports to work. Sending this out now as a draft just to get feedback on the overall approach and location.


packages/fixie/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "fixie",
+  "version": "1.0.0",
```
petersalas, line 3: Given that this is all external to the AI.JSX package I think this should probably all live in the fixie-sdk repo?

packages/fixie/src/serve-bin.ts
```diff
@@ -0,0 +1,83 @@
...
+  silentStartup: boolean;
+}): Promise<void> {
+  //const handler = await import(packagePath);
+  const handler = index;
+  const app: FastifyInstance = fastify();
+  app.post('/', async (req: FastifyRequest, res: FastifyReply) => {
+    const body = req.body as { message: { text: string } };
+    try {
+      const responseStream = handler(body.message.text);
+      await sendReadableStreamToFastifyReply(res, responseStream);
```
petersalas, line 38: From some poking around I did previously I think you can just write `res.send(responseStream)` -- did that not work?

packages/fixie/package.json
petersalas: I guess unless the eventual plan is for the the serving CLI to be in ai-jsx itself? But it's weird to see the `agent.yaml` here.

packages/fixie/src/serve-bin.ts
juberti: cribbed this from an example - didn't try simplifying yet, will give it a shot

packages/fixie/src/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <ChatCompletion>
+      <SystemMessage>Respond to the user using some variant of the phrase "Hello World!".</SystemMessage>
+      <UserMessage>{message}</UserMessage>
+    </ChatCompletion>
+  );
+}
+
+export default function handleMessage(message: string) {
+  return toTextStream(<HelloWorld message={message} />);
```
NickHeiner, line 14: If we're trying to make the platform hosting as seamless as possible, should we not even do `toTextStream` – instead, have this file just export the component, and the platform wraps with `toTextStream`?

packages/fixie/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "scripts": {
+    "prod": "yarn build && node dist/serve-bin.js --port 8080",
+    "lint": "eslint . --max-warnings 0",
+    "lint:fix": "eslint . --fix",
+    "typecheck": "tsc -p tsconfig.json",
+    "build": "yarn run typecheck",
+    "bin:dev": "npx ts-node dist/serve-bin.js --port 8080 --package-path fixie" 
+  },
+  "dependencies": { 
+    "ai-jsx": "^0.7.3",
```
NickHeiner, line 31:
```suggestion
"ai-jsx": "workspace:*",
```

packages/fixie/agent.yaml
```diff
@@ -0,0 +1,4 @@
+handle: ai-jsx-hello-world
+description: Hello World for AI.JSX
+more_info_url: ''
+entry_point: main:agent
```
NickHeiner, line 4: Do we still need this?

NickHeiner:
Be sure to run `yarn` locally to fix the CI problem.

NickHeiner (approved):
I haven't run this locally but it seems reasonable to me.

packages/fixie/package.json
juberti: Since this is going to deploy to GCR independently of ai-jsx, I think this needs to have a specific revision that it can pull? Or am I overlooking something?

packages/fixie/src/index.tsx
juberti: let me give that a shot

packages/fixie/package.json
juberti: I think it's OK to have agent.yaml outside the core ai.jsx package, it's a config file just like the next.js config files in the nextjs-demo package. Right?

packages/fixie/agent.yaml
juberti: I think this would probably be the way we identify the package to call to get the main component, but clearly it needs to be different than `main:agent`.

packages/fixie/package.json
juberti: maybe we postproc this file when building the container to replace "workspace" with the local version?

petersalas: If you do `npm pack` and it has a workspace reference will that automatically "link in" (i.e. copy) the files from the workspace into the tarball? Might be a question for @NickHeiner

petersalas: The answer seems to be "no". IMO sticking with a version that resolves through NPM is sensible.

juberti: thanks for looking into that. going to just hack this to insert "latest" for the time being

PR #230, opened 2023-08-07
Make `memo` less spammy for logs/instrumentation

petersalas:
This changes `memo` and `withContext` to introduce wrapper elements less aggressively so as not to bloat instrumentation with `<Memoized>`/`<MemoizedGenerator>` et al.

Additionally, change the `memoizedId` semantics so that memoized _subtrees_ are tagged with the same memoization ID, rather than assigning new IDs at each layer. This will be used subsequently to allow instrumentation to associate each memoized subtree to a single asynchronous render (drastically improving instrumentation legibility).


NickHeiner (approved):
I didn't review this super closely tbh.

Is there anything in the docs we should update, given the changed semantics?

petersalas:
> Is there anything in the docs we should update, given the changed semantics?

I don't think so, no -- these are all internal implementation details.

PR #232, opened 2023-08-08
Add ability to customize debug representation

petersalas:
Allow elements to have custom `debug` representations and use it so that the debug representation of streaming `<AssistantMessage>`s includes the streamed value thus far.


packages/ai-jsx/src/lib/anthropic.tsx
```diff
@@ -175,32 +176,34 @@ export async function* AnthropicChatModel(
...
-          }
-
-          logger.debug({ completion: resultSoFar }, 'Anthropic completion finished');
-          return AI.AppendOnlyStream;
-        })()
-      )}
-    </AssistantMessage>
+  let accumulatedContent = '';
+  let complete = false;
+  const Stream = async function* (): AI.RenderableStream {
```
NickHeiner, line 181: Any reason not to do the traditional `async function* Stream(): AI.RenderableStream {` here?

PR #233, opened 2023-08-09
Update rendering.md

benlower:
refactoring this page to have streaming concepts first

### DIFF
--- a/packages/docs/docs/guides/rendering.md
+++ b/packages/docs/docs/guides/rendering.md
@@ -1,39 +1,221 @@
 # Rendering
 
-Once you assemble your AI.JSX component tree, you'll want to render it into text or UI components. The way you do it depends on how you're using AI.JSX.
+Once you assemble your AI.JSX component tree, you'll want to render it into text or UI components. The way in which components get assembled depends on your scenario and how you are using AI.JSX. By default (and to provide the best end-user experience) AI.JSX streams results from all calls you make to an LLM.
 
-## From clientside React
+## Streaming in AI.JSX
+
+Rendering an AI.JSX program yields a stream. By default, AI.JSX will give you partial results for the output. Alternatively, you can choose to use append-only streams which outputs sequential chunks that can then be concatenated together.
+
+### Strings vs. Streams
+
+When you `await` the result of `render`, you get a string:
+
+```tsx
+const str = await AI.createRenderContext().render(<App />);
+```
+
+This might be fine for a short response. However, for performance, you'll often prefer to get a stream of results. This enables your app to display content to your end-user as it comes in versus having a long period of time where nothing seems to be happening.
+
+To stream like this, treat the result of `render` like a generator:
+
+```tsx
+const result = AI.createRenderContext().render(<App />);
+let frameCount = 0;
+for await (const frame of result) {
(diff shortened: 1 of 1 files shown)


packages/docs/docs/guides/rendering.md
```diff
@@ -1,74 +1,179 @@
...
 
-Once you assemble your AI.JSX component tree, you'll want to render it into text or UI components. The way you do it depends on how you're using AI.JSX.
+Once you assemble your AI.JSX component tree, you'll want to render it into text or UI components. The way in which components get assembled depends on your scenario and how you are using AI.JSX. By default (and to provide the best end-user experience) AI.JSX streams results from all calls you make to an LLM.
 
-## From clientside React
+## Streaming in AI.JSX
 
-:::note Architectures
-This applies to the following architectures:
+Rendering an AI.JSX program yields a stream. By default, AI.JSX will give you partial results for the output. Alternatively, you can choose to use append-only streams which outputs sequential chunks that can then be concatenated together.
```
NickHeiner, line 7: I don't think this sentence will make sense to low-context people. Maybe what I wrote here is better, but I'm not sure. :smile:

```suggestion
Rendering an AI.JSX program yields a stream. You can choose to consume it one of two ways:
1. A stream of deltas (that you have to `+=` together to get the final result)
1. A stream of in-progress versions of the final full result.

We'll explore these two options in more detail below.
```

packages/docs/docs/guides/rendering.md
```diff
@@ -102,49 +207,79 @@ export async function POST(request: NextRequest) {
...
 This applies to the following architectures:
 
-- [Headless AI (no UI)](./architecture.mdx#headless-ai)
-- Any other case where you just want a string.
+- [Server-Side AI.JSX + Client UI](./architecture.mdx#architecture-3-server-side-aijsx--client-ui)
 
 :::
 
-To render your component to a string:
+Use the `useAIStream` hook to fetch content from your server endpoint:
```
NickHeiner, line 222: These docs show how to fetch from the endpoint, but they don't actually talk about what the endpoint itself should be, which is really where the rendering is happening.

NickHeiner (approved):

packages/docs/docs/guides/rendering.md
benlower: cool thanks. took your suggestion here and reworked it a bit.

packages/docs/docs/guides/rendering.md
```diff
@@ -1,74 +1,179 @@
...
 
-:::
+When you `await` the result of `render`, you get a string:
 
-Use the `AI.JSX` component to directly render AI-generated content into your UI:
+```tsx
+const str = await AI.createRenderContext().render(<App />);
+```
+
+This might be fine for a short response. However, for performance, you'll often prefer to get a stream of results. This enables your app to display content to your end-user as it comes in versus having a long period of time where nothing seems to be happening.
```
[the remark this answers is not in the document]
benlower: updating based on this suggestion

packages/docs/docs/guides/rendering.md
benlower: resolving this for now and will address this in the next update

PR #234, opened 2023-08-09
Fix ai-jsx ref for fixie package

juberti:


packages/fixie/src/index.tsx
```diff
@@ -1,3 +1,4 @@
+import 'dotenv/config';
```
petersalas, line 1: Can probably remove this one now :)

petersalas (approved):

PR #237, opened 2023-08-11
Model updates

NickHeiner:
Fixing some issues I ran in to when trying to use new models.

See the changelog for more detail about what's changing.


packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -4,97 +4,9 @@
  * @packageDocumentation
  */
 
-import {
```
NickHeiner, line 7: Removing the prompt-engineered polyfill re @zkoch's suggestion, which I agree with.

PR #238, opened 2023-08-14
Add token counts to OpenTel traces

NickHeiner:
This makes it easier to know how different types of content (e.g. system messages, docsQA, etc) are contributing to the final token count.

This will not exhaustively cover all usecases.

This also adds a unit test.


packages/ai-jsx/src/core/opentelemetry.ts
```diff
@@ -59,8 +63,12 @@ export function openTelemetryStreamRenderer(streamRenderer: StreamRenderer): Str
           throw ex;
         } finally {
           if (result) {
+            const resultIsPartial = result.find(isElement);
             // Record the rendered value.
-            span.setAttribute('ai.jsx.result', result.find(isElement) ? debug(result, true) : result.join(''));
+            span.setAttribute('ai.jsx.result', resultIsPartial ? debug(result, true) : result.join(''));
+            if (!resultIsPartial) {
+              span.setAttribute('ai.jsx.result.tokenCount', getEncoder().encode(result.join('')).length);
+            }
```
NickHeiner, line 71: It seems like it would be nice if this and the standard logging were unified, such that we only had one place to attach metadata to render results.

@petersalas, I'm guessing your preferred means of that unification would be to kill the other logging. :smile:

petersalas: Yup exactly

packages/ai-jsx/src/core/opentelemetry.ts
```diff
@@ -59,8 +63,12 @@ export function openTelemetryStreamRenderer(streamRenderer: StreamRenderer): Str
           throw ex;
         } finally {
           if (result) {
+            const resultIsPartial = result.find(isElement);
             // Record the rendered value.
-            span.setAttribute('ai.jsx.result', result.find(isElement) ? debug(result, true) : result.join(''));
+            span.setAttribute('ai.jsx.result', resultIsPartial ? debug(result, true) : result.join(''));
+            if (!resultIsPartial) {
+              span.setAttribute('ai.jsx.result.tokenCount', getEncoder().encode(result.join('')).length);
```
petersalas, line 70: Is there no sensible place to log this from OpenAI or ShrinkConversation? We already log the entire prompt going into the model (I think twice even), and attaching the token counts seems like it would be straightforward.

Seeing a single event with the entire prompt broken down by token count seems easier to grok than needing to inspect N different components.

From a trace perspective if we wanted a span to attach aggregated token counts to, I think the OpenAIChatModel one would make sense to use for both prompt and completion token counts/estimates. (It's straightforward to attach arbitrary attributes to the active span because there's a global way to get the current span.)

NickHeiner: I agree that we need better visibility into the final prompt. But I also think there's value in getting a token count for each individual component; by the time we get to `OpenAI`, those components may have been rendered away. And `ShrinkConversation` may not even be in use.

So, my suggestion is that we do both this and what you're suggestion.

PR #240, opened 2023-08-16
Add explicit prompt/completion logging/attributes, skip tracing for synchronous components

petersalas:
1. Add explicit (conversational) prompt/completion logging and span attributes, including token counts
2. Skip OpenTelemetry tracing for non-root synchronous components, which drastically simplifies the traces
3. Remove universal `ai.jsx.result.tokenCount` span attribute, which doesn't have the opportunity to handle conversational messages correctly

### DIFF
--- a/packages/ai-jsx/src/core/conversation.tsx
+++ b/packages/ai-jsx/src/core/conversation.tsx
@@ -204,6 +204,8 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
 export async function renderToConversation(
   conversation: AI.Node,
   render: AI.ComponentContext['render'],
+  logger?: AI.ComponentContext['logger'],
+  logType?: 'prompt' | 'completion',
   cost?: (message: ConversationMessage, render: AI.ComponentContext['render']) => Promise<number>,
   budget?: number
 ) {
@@ -215,7 +217,21 @@ export async function renderToConversation(
     ) : (
       conversation
     );
-  return toConversationMessages(await render(conversationToUse, { stop: isConversationalComponent }));
+  const messages = toConversationMessages(await render(conversationToUse, { stop: isConversationalComponent }));
+
+  if (logger && logType) {
+    const loggableMessages = await Promise.all(
+      messages.map(async (m) => ({
+        element: debug(m.element, true),
+        ...(cost && { cost: await cost(m, render) }),
+      }))
+    );
+
+    logger.setAttribute(`ai.jsx.${logType}`, JSON.stringify(loggableMessages));
+    logger.info({ [logType]: { messages: loggableMessages } }, `Got ${logType} conversation`);
+  }
+
+  return messages;
 }
 
 /**
@@ -258,14 +274,14 @@ export async function* Converse(
     reply: (messages: ConversationMessage[], fullConversation: ConversationMessage[]) => AI.Renderable;
     children: AI.Node;
   },
-  { render, memo }: AI.ComponentContext
+  { render, memo, logger }: AI.ComponentContext
 ): AI.RenderableStream {
   yield AI.AppendOnlyStream;
 
   const fullConversation = [] as ConversationMessage[];
   let next = memo(children);
   while (true) {
-    const newMessages = await renderToConversation(next, render);
+    const newMessages = await renderToConversation(next, render, logger);
     if (newMessages.length === 0) {
       break;
     }
--- a/packages/ai-jsx/src/core/log.ts
+++ b/packages/ai-jsx/src/core/log.ts
@@ -155,22 +155,16 @@ export class CombinedLogger extends LogImplementation {
     super();
   }
 
-  log(
-    level: LogLevel,
-    element: Element<any>,
-    renderId: string,
-    metadataOrMessage: string | object,
-    message?: string | undefined
-  ): void {
-    this.loggers.forEach((l) => l.log(level, element, renderId, metadataOrMessage, message));
+  log(...args: Parameters<LogImplementation['log']>): void {
+    this.loggers.forEach((l) => l.log(...args));
   }
 
-  logException(element: Element<object>, renderId: string, exception: unknown): void {
-    this.loggers.forEach((l) => l.logException(element, renderId, exception));
+  logException(...args: Parameters<LogImplementation['logException']>): void {
+    this.loggers.forEach((l) => l.logException(...args));
   }
 
-  setAttribute(element: Element<any>, renderId: string, key: string, value: string): void {
-    this.loggers.forEach((l) => l.setAttribute(element, renderId, key, value));
+  setAttribute(...args: Parameters<LogImplementation['setAttribute']>): void {
+    this.loggers.forEach((l) => l.setAttribute(...args));
   }
 }
 
--- a/packages/ai-jsx/src/core/log.ts
+++ b/packages/ai-jsx/src/core/log.ts
@@ -6,6 +6,7 @@
 import _ from 'lodash';
 import pino from 'pino';
 import { Element } from './node.js';
+import opentelemetry from '@opentelemetry/api';
 import { logs, type Logger as OpenTelemetryLoggerInterface } from '@opentelemetry/api-logs';
 import { getEnvVar } from '../lib/util.js';
 
@@ -14,7 +15,9 @@ export type LogLevel = 'fatal' | 'error' | 'warn' | 'info' | 'debug' | 'trace';
 /**
  * A Logger represents an object that can log messages.
  */
-export type Logger = Record<LogLevel, (obj: object | string, msg?: string) => void>;
+export type Logger = Record<LogLevel, (obj: object | string, msg?: string) => void> & {
+  setAttribute: (key: string, value: string) => void;
+};
 
 /**
  * An abstract class that represents an implementation of {@link Logger}.
@@ -64,6 +67,11 @@ export abstract class LogImplementation {
       `Rendering element ${elementTag} failed with exception: ${exception}`
     );
   }
+
+  /**
+   * Sets an attribute to be associated with the rendering of a particular element.
+   */
+  setAttribute(_element: Element<any>, _renderId: string, _key: string, _value: string) {}
 }
 
 /**
@@ -129,6 +137,14 @@ export class OpenTelemetryLogger extends LogImplementation {
       },
     });
   }
+
+  setAttribute(element: Element<any>, renderId: string, key: string, value: string): void {
+    // Set the attribute on the current span.
+    const span = opentelemetry.trace.getActiveSpan();
+    if (span) {
+      span.setAttribute(key, value);
+    }
+  }
 }
 
 /**
@@ -152,6 +168,10 @@ export class CombinedLogger extends LogImplementation {
   logException(element: Element<object>, renderId: string, exception: unknown): void {
     this.loggers.forEach((l) => l.logException(element, renderId, exception));
   }
+
+  setAttribute(element: Element<any>, renderId: string, key: string, value: string): void {
+    this.loggers.forEach((l) => l.setAttribute(element, renderId, key, value));
+  }
 }
 
 /**
@@ -170,4 +190,5 @@ export class BoundLogger implements Logger {
   info = (obj: object | string, msg?: string) => this.impl.log('info', this.element, this.renderId, obj, msg);
   debug = (obj: object | string, msg?: string) => this.impl.log('debug', this.element, this.renderId, obj, msg);
   trace = (obj: object | string, msg?: string) => this.impl.log('trace', this.element, this.renderId, obj, msg);
+  setAttribute = (key: string, value: string) => this.impl.setAttribute(this.element, this.renderId, key, value);
 }
--- a/packages/ai-jsx/src/core/opentelemetry.ts
+++ b/packages/ai-jsx/src/core/opentelemetry.ts
@@ -3,10 +3,9 @@ import { PartiallyRendered, StreamRenderer } from './render.js';
 import { Element, isElement } from './node.js';
 import { memoizedIdSymbol } from './memoize.js';
 import { debug } from './debug.js';
-import { getEncoding } from 'js-tiktoken';
-import _ from 'lodash';
+import { getEnvVar } from '../lib/util.js';
 
-function bindAsyncGeneratorToActiveContext<T = unknown, TReturn = any, TNext = unknown>(
+export function bindAsyncGeneratorToActiveContext<T = unknown, TReturn = any, TNext = unknown>(
   generator: AsyncGenerator<T, TReturn, TNext>
 ): AsyncGenerator<T, TReturn, TNext> {
   const activeContext = context.active();
@@ -31,6 +30,10 @@ interface MemoizedIdAndOwner {
   isOwner: boolean;
 }
 
+// Indicates whether we're currently tracing _any_ component so that we can skip
+// tracing non-root synchronous components.
+const isTracingComponent = createContextKey('Whether the current async context is tracing a component');
+
 // Indicates whether the current async context is responsible for tracing the memoized subtree.
 const activeMemoizedIdStorage = createContextKey('The active MemoizedIdAndOwner');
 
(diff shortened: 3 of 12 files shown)


petersalas:
@NickHeiner do you think this tracing/logging is good/comprehensive enough to remove the full render logging/token count attribute? Here's an [example trace](https://app.datadoghq.com/apm/trace/14113410336604669720) from the use-tools example

packages/ai-jsx/src/core/log.ts
```diff
@@ -129,6 +137,14 @@ export class OpenTelemetryLogger extends LogImplementation {
...
     });
   }
+
+  setAttribute(element: Element<any>, renderId: string, key: string, value: string): void {
+    // Set the attribute on the current span.
+    const span = opentelemetry.trace.getActiveSpan();
+    if (span) {
+      span.setAttribute(key, value);
+    }
+  }
```
NickHeiner, line 147: Heh, it's a bit funny that `element` isn't actually used in this method – do we actually want to commit ourselves to there only ever being one active span? And for that span only being linked to an element? Does that conflict with using a span to track the API call to the model provider? Maybe I'm missing something here.

Does this even work reliably with async?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -204,6 +204,8 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
 export async function renderToConversation(
   conversation: AI.Node,
   render: AI.ComponentContext['render'],
+  logger?: AI.ComponentContext['logger'],
+  logType?: 'prompt' | 'completion',
```
NickHeiner, line 208: What's the case where these two values are undefined?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -215,7 +217,21 @@ export async function renderToConversation(
...
+  if (logger && logType) {
+    const loggableMessages = await Promise.all(
+      messages.map(async (m) => ({
+        element: debug(m.element, true),
+        ...(cost && { cost: await cost(m, render) }),
+      }))
+    );
+
+    logger.setAttribute(`ai.jsx.${logType}`, JSON.stringify(loggableMessages));
+    logger.info({ [logType]: { messages: loggableMessages } }, `Got ${logType} conversation`);
```
NickHeiner, line 231: Are there links you can share of how this looks?

packages/ai-jsx/src/core/opentelemetry.ts
```diff
@@ -66,9 +78,6 @@ export function openTelemetryStreamRenderer(streamRenderer: StreamRenderer): Str
             const resultIsPartial = result.find(isElement);
             // Record the rendered value.
             span.setAttribute('ai.jsx.result', resultIsPartial ? debug(result, true) : result.join(''));
-            if (!resultIsPartial) {
-              span.setAttribute('ai.jsx.result.tokenCount', getEncoder().encode(result.join('')).length);
-            }
```
NickHeiner, line 71: lol I'm annoyed with myself for doing something that was basically useless... was there a good way for me to have known that this wasn't right?

packages/ai-jsx/src/core/opentelemetry.ts
```diff
@@ -31,24 +30,37 @@ interface MemoizedIdAndOwner {
...
 const activeMemoizedIdStorage = createContextKey('The active MemoizedIdAndOwner');
 
 function spanAttributesForElement(element: Element<any>): opentelemetry.Attributes {
   return { 'ai.jsx.tag': element.tag.name, 'ai.jsx.tree': debug(element, true) };
 }
 
-const getEncoder = _.once(() => getEncoding('cl100k_base'));
-
 export function openTelemetryStreamRenderer(streamRenderer: StreamRenderer): StreamRenderer {
+  const traceEveryComponent = Boolean(getEnvVar('AIJSX_OPENTELEMETRY_TRACE_ALL', false));
```
NickHeiner, line 45: Hmm... this is not a strong objection, but I feel like the more robust thing to do here is use loglevels. But I guess OpenTel and/or DataDog doesn't have a loglevel concept? So even if we were to annotate spans with a loglevel, the consumers we have today wouldn't be able to use that?

packages/ai-jsx/src/core/opentelemetry.ts
```diff
@@ -31,24 +30,37 @@ interface MemoizedIdAndOwner {
...
 
-  return (renderContext, renderable, shouldStop, appendOnly) => {
+  const wrappedRender: StreamRenderer = (renderContext, renderable, shouldStop, appendOnly) => {
     if (!isElement(renderable)) {
       return streamRenderer(renderContext, renderable, shouldStop, appendOnly);
     }
 
-    const activeContext = context.active();
+    let activeContext = context.active();
+    if (!traceEveryComponent && renderable.tag.constructor === Function && activeContext.getValue(isTracingComponent)) {
```
NickHeiner, line 54: Hmm. This seems like a clever heuristic that happens to eliminate most of the AI.JSX internal components, but I think the false positives of "userland components that also get eliminated" will be pretty frustrating ("where are my components?"). In the Sidekick Fantasy Football, for instance, I had [a synchronous user-land component](https://github.com/fixie-ai/sidekick-fantasy-football/blob/e74b07cda8ac953abb1c5fa872da9bc026b960b1/app/api/chat/stats.tsx). 

And the "non-async functions which return promises" thing also seems like a footgun.

What I would suggest instead, and I recognize this is annoyingly boilerplatey, is to explicitly mark every component with some sort of flag:

```ts
function SwitchContext() { }
SwitchContext.aiJSXInternal = true;

// or
SwitchContext.logInTelemetry = true;

// or
SwitchContext.telemetryLogLevel = 'trace';
```

This way, by default, every userland component shows up in the trace, which I think is what we want. And the latter two options I gave would let the user opt out of that as well.

packages/ai-jsx/src/core/opentelemetry.ts
```diff
@@ -91,13 +100,11 @@ export function openTelemetryStreamRenderer(streamRenderer: StreamRenderer): Str
...
-          tracer.startActiveSpan(
-            `<${renderable.tag.name}>`,
-            { attributes: spanAttributesForElement(renderable) },
-            renderInSpan
-          )
+        return tracer.startActiveSpan(
+          `<${renderable.tag.name}>`,
+          { attributes: spanAttributesForElement(renderable) },
+          activeContext.setValue(activeMemoizedIdStorage, { id: memoizedId, isOwner: true }),
+          renderInSpan
```
NickHeiner, line 107: What's the significance of passing `newContext` as a param to `startActiveSpan` instead of doing `context.with`?

packages/ai-jsx/src/core/render.ts
```diff
@@ -311,19 +311,10 @@ async function* renderStream(
...
-       * locate + call the model provider, etc.
-       *
-       * To get around this, maybe we want components to be able to choose the loglevel used for their rendering.
-       */
-      logImpl.log('debug', renderable, renderId, 'Start rendering element');
       const finalResult = yield* renderingContext.render(
         renderable.render(renderingContext, new BoundLogger(logImpl, renderId, renderable), appendOnly),
         recursiveRenderOpts
       );
-      logImpl.log('debug', renderable, renderId, { finalResult }, 'Finished rendering element');
```
NickHeiner, line 326: If we take this out, is there any other way to get this signal that doesn't involve OpenTel?

NickHeiner: I would be fine removing this if we replaced it with some sort of an "OpenTel console logger". That would be nicer, because we would be unifying the logging. And it would preserve visibility for people who can't/don't want to run Docker and have an external viewer like Jaeger or DD set up.

packages/ai-jsx/src/core/conversation.tsx
NickHeiner, line 208: What's the semantic difference between `prompt` and `completion`?

packages/examples/test/core/completion.tsx
```diff
@@ -53,9 +52,14 @@ afterEach(() => {
...
-  tracerProvider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
-  trace.setGlobalTracerProvider(tracerProvider);
+  const sdk = new NodeSDK({
+    spanProcessor: new SimpleSpanProcessor(memoryExporter) as any,
+  });
+  sdk.start();
+
+  afterAll(async () => {
+    await sdk.shutdown();
+  });
```
NickHeiner, line 62: This can just be:

```suggestion
  afterAll(() =>  sdk.shutdown());
```

packages/ai-jsx/src/core/conversation.tsx
NickHeiner: Ah, it's the input and output?

packages/ai-jsx/src/core/conversation.tsx
NickHeiner: ah nm I see you shared https://app.datadoghq.com/apm/trace/14113410336604669720?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=15969530707541520513 in a separate comment

NickHeiner:
> @NickHeiner do you think this tracing/logging is good/comprehensive enough to remove the full render logging/token count attribute? Here's an [example trace](https://app.datadoghq.com/apm/trace/14113410336604669720) from the use-tools example

image

Maybe I'm missing something, but where do we see how many tokens `UseTools` resolved to?

packages/ai-jsx/src/core/log.ts
```diff
@@ -64,6 +67,11 @@ export abstract class LogImplementation {
       `Rendering element ${elementTag} failed with exception: ${exception}`
     );
   }
+
+  /**
+   * Sets an attribute to be associated with the rendering of a particular element.
+   */
+  setAttribute(_element: Element<any>, _renderId: string, _key: string, _value: string) {}
```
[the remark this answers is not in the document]
petersalas: Lint rule -- the alternative is to omit the parameters, is that preferable?

packages/ai-jsx/src/core/opentelemetry.ts
petersalas: Yeah I agree -- I looked around for something like that but I don't think that's a thing. I suspect the reason is that spans have a parent-child link, and so if you were to link an INFO span to a DEBUG span, any level based filter would have to buffer non-trivial state in order to patch up the relationships.

So I think any level-ish logic is on our side where we decide to emit them -- if we find ourselves wanting more granularity here we can model this after log levels but right now my philosophy is ~YAGNI.

packages/ai-jsx/src/core/log.ts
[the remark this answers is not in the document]
NickHeiner: Hmm, I suppose that `LogImplementation` is also serving as an interface in this case, so if you remove these params, then that could break other code. Leaving it is preferable then.

packages/ai-jsx/src/core/opentelemetry.ts
NickHeiner: Yep, that's fine with me.

packages/ai-jsx/src/core/opentelemetry.ts
petersalas: Yeah I started looking at this from exactly that angle but convinced myself that the juice wasn't going to be worth the squeeze vs. the async heuristic.

One long-running related thought I've had is whether "structural components" should be made explicitly inert with `string` or `Symbol` tags (e.g. `<AssistantMessage>` might become something like `<assistant>`). This might allow for:
 - natural distinction between no-op components and meaningful components for things like OTel
 - replace partial rendering with something more constrained/easier to reason about
 - avoid the "oops the tree was flattened" footgun
 - native tree serialization

(Don't worry, that's still just an idea at this point.)

packages/ai-jsx/src/core/conversation.tsx
petersalas: Right, specifically it indicates that the conversation is the input to/output from a chat model.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -536,16 +539,19 @@ export async function* OpenAIChatModel(
...
         return AI.AppendOnlyStream;
       };
-      const assistantStream = memo(
-        <Stream {...debugRepresentation(() => `${accumulatedContent}${complete ? '' : '▮'}`)} />
+      const assistantMessage = memo(
+        <AssistantMessage>
+          <Stream {...debugRepresentation(() => `${accumulatedContent}${complete ? '' : '▮'}`)} />
+        </AssistantMessage>
       );
-      yield <AssistantMessage>{assistantStream}</AssistantMessage>;
```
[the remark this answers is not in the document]
petersalas: Nah, I just needed a reference to the assistant message instead of the stream and felt this was easier to read than

```tsx
<AssistantMessage>
   {memo(<Stream />)}
</AssistantMessage>
```

and in particular, I wanted to have the `render` and the `memo` call operate on the same value.

petersalas:
> Maybe I'm missing something, but where do we see how many tokens UseTools resolved to?

You can't -- my claim is that token counts are only meaningful _in the context of a single model_, so the token counts are only available on the `<OpenAIChatModel>` spans. In particular with things like context-window-trimming, the relationship between what `<UseTools>` renders to and the underlying model calls is non-trivial. (Also, which tokenizer should be used? What if different models are used?)

That said, a more well-defined thing we could do is _aggregate_ token usage as reported by the descendant model calls, but that's a different beast. (My inclination is also to do that aggregation downstream rather than in the code itself.)

packages/ai-jsx/src/core/log.ts
petersalas: It does work with async -- OpenTelemetry uses https://nodejs.org/api/async_context.html to propagate context. (Which works nicely modulo https://github.com/open-telemetry/opentelemetry-js/issues/2951, hence the temporary workarounds for generators)

If we were worried about competing spans we could add our own property on the context for something like "active AI.JSX render span" to distinguish between "active span" but I think it'd be unusual for a component to go through the trouble of entering a custom span and use `logger.setAttribute` (with the specific expectation of it landing on the render span). So I'm going to defer that :)

packages/ai-jsx/src/core/opentelemetry.ts
petersalas: Meh, I wouldn't go so far to say it's useless:
- It's not obvious from this code that there's extensive usage of partial rendering for conversations
- Maybe one hint of trouble is that there's no way to use the correct tokenizer here? (If we did token counting for Anthropic/Llama, that is.)
- That this code is far away from anything else that does token counting is also a bit dicey

packages/ai-jsx/src/core/render.ts
petersalas: There's a straightforward-ish way to set up a collector so that it emits things like:
```
Span #2
    Trace ID       : 48b03803a7fd4c6beea750037e9a846f
    Parent ID      : 939c625fdad147a9
    ID             : fd2a38dae778b6c9
    Name           : <OpenAICompletionModel>
    Kind           : Internal
    Start time     : 2023-08-18 16:09:22.015000064 +0000 UTC
    End time       : 2023-08-18 16:09:25.32702592 +0000 UTC
    Status code    : Ok
    Status message : 
Attributes:
     -> ai.jsx.tag: Str(OpenAICompletionModel)
     -> ai.jsx.tree: Str(<OpenAICompletionModel @memoizedId=1 model="text-davinci-003" maxTokens={100}>
  {"Q: Why is the sky blue?"}
  {"\n"}
  {"A:"}
</OpenAICompletionModel>)
     -> ai.jsx.result: Str( The sky is blue because molecules in the Earth's atmosphere scatter sunlight in all directions. Blue light is scattered more than other colors because it travels as shorter, smaller waves. This is why we see a blue sky most of the time.)
```
to the console.

Another alternative is to configure the OTel SDK to write spans to the console, though obviously that doesn't work well if the console is also being used for the inspector.

I guess my high level take is that I think these logs are so noisy that sifting through them _without_ a UI is of limited utility (or at least significantly _less_ useful) but that's certainly a judgment call.

Mainly I wanted to see if you were (at all) reluctant to see this go, which it sounds like you are, so I can bring them back. :)

packages/ai-jsx/src/core/opentelemetry.ts
petersalas: Oh, I meant to also add, re: the fantasy football example:
- I think the contents of such components are mainly useful when shown as part of the prompt (which is logged both as JSX and OpenAI request object)
- Such components are free to log whatever they want :)
- There's a cheesy but trivial workaround to opt into tracing, which is simply to make the function `async` (even if it doesn't use `await`)

NickHeiner:
> > Maybe I'm missing something, but where do we see how many tokens UseTools resolved to?
> 
> You can't -- my claim is that token counts are only meaningful _in the context of a single model_, so the token counts are only available on the `<OpenAIChatModel>` spans. In particular with things like context-window-trimming, the relationship between what `<UseTools>` renders to and the underlying model calls is non-trivial. (Also, which tokenizer should be used? What if different models are used?)
> 
> That said, a more well-defined thing we could do is _aggregate_ token usage as reported by the descendant model calls, but that's a different beast. (My inclination is also to do that aggregation downstream rather than in the code itself.)

Ok, great point that the tokenizer means the count is only meaningful within the context of a single model.

Broadly, the thing I'm trying to support is the ability to ask "how many tokens did I spend on docs vs API responses". My original solution was to give an output size for every component, but I agree with your points that there are some problems with that. Could we address that with your downstream aggregation suggestion?

packages/ai-jsx/src/core/render.ts
NickHeiner: Yes, in my judgment, the logs are noisy, but sifting through them is actually not that hard, and without them, setting up OpenTel is a meaningful burden. 

`ai-jsx-program | grep MyComponentName | npx pino-pretty` drops a lot of the noise for me.

NickHeiner (approved):
I disagree with some aspects of how this is being done but don't want to block on it.

Can you update https://docs.ai-jsx.com/guides/observability to reflect these changes?

And perhaps add one or two more unit tests?

petersalas:
> Broadly, the thing I'm trying to support is the ability to ask "how many tokens did I spend on docs vs API responses". My original solution was to give an output size for every component, but I agree with your points that there are some problems with that. Could we address that with your downstream aggregation suggestion?

IIUC I think that scenario is already addressed with the newly added `ai.jsx.prompt` attribute, which shows each conversation message given to the LLM and its weight.

PR #242, opened 2023-08-18
Add ability to reserve tokens without (statically) limiting response length

petersalas:
Add a `reservedTokens` prop to allow consumers to reserve a certain number of tokens for the response without unconditionally limiting the response length. Assuming token estimation is perfectly accurate (and that `reservedTokens < maxTokens`) this means that:
- The response will never be cut off before `reservedTokens`
- The response _may_ be cut off between `reservedTokens` and `maxTokens`
- The response will never exceed `maxTokens`

Consumers may also set `maxTokens` _less_ than `reservedTokens` to set an upper bound on the total number of input/output tokens (e.g. to constrain costs). The upper bound is `context size - reserved tokens + max tokens`.

### DIFF
--- a/packages/ai-jsx/src/core/completion.tsx
+++ b/packages/ai-jsx/src/core/completion.tsx
@@ -30,6 +30,8 @@ export interface ModelProps {
   temperature?: number;
   /** The maximum number of tokens to generate. */
   maxTokens?: number;
+  /** The number of tokens to reserve for the generation. */
+  reservedTokens?: number;
   /** A list of stop tokens. */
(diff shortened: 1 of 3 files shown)


NickHeiner (approved):
LGTM. Can you add docs and unit tests?

petersalas:
> Can you add docs and unit tests?

Added unit tests, but I don't think there's a natural place to add docs (beyond the API docs, which are already there) short of committing hard to `<Shrinkable>` et al by writing "guide"-type docs, which I don't think we're ready to do.

PR #243, opened 2023-08-22
Change memo to no longer replay streams

petersalas:
- Change the behavior of `memo` so that it only yields the most recently rendered frame, rather than replaying the entire stream. When combined with non-append-only streams + partial rendering the previous behavior could cause an exponential blowup in the number of frames.
- Change `AI.AppendOnlyStream` to be a function that takes a `Node` so that converting a stream to append-only does not require an extra yield (i.e. an extra frame).
- Change partial rendering to memoize returned elements, instead of simply binding the context, and remove the now-unneeded `memo`s in `conversation.tsx` accordingly.
- Source the memoized ID from the render context so that it (can be) deterministic within a single `RenderContext`
- Add unit tests for `memoize.tsx`.

### DIFF
--- a/packages/ai-jsx/src/core/memoize.tsx
+++ b/packages/ai-jsx/src/core/memoize.tsx
@@ -1,9 +1,16 @@
-import { Renderable, RenderContext, AppendOnlyStream, RenderableStream } from './render.js';
-import { Node, getReferencedNode, isIndirectNode, makeIndirectNode, isElement } from './node.js';
+import {
+  Renderable,
+  RenderContext,
+  AppendOnlyStream,
+  RenderableStream,
+  AppendOnlyStreamValue,
+  isAppendOnlyStreamValue,
+  valueToAppend,
+} from './render.js';
+import { Node, Element, getReferencedNode, isIndirectNode, makeIndirectNode, isElement } from './node.js';
 import { Logger } from './log.js';
 import { bindAsyncGeneratorToActiveContext } from './opentelemetry.js';
 
-let lastMemoizedId = 0;
 /** @hidden */
 export const memoizedIdSymbol = Symbol('memoizedId');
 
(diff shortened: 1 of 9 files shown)


packages/ai-jsx/src/core/memoize.tsx
```diff
@@ -76,32 +83,53 @@ export function partialMemo(renderable: Renderable, existingId?: number): Node |
...
+              }
+
+              if (!isAppendOnly) {
+                nodes = [value];
+              } else {
+                nodes.push(value);
+              }
+            }
+
+            const valueToYield = isAppendOnly ? AppendOnlyStream(nodes) : nodes;
```
NickHeiner, line 115: I'm sure I'm missing something, but why yield `nodes` as opposed to `_.last(sink)`?

packages/examples/test/core/completion.tsx
```diff
@@ -75,8 +75,6 @@ describe('OpenTelemetry', () => {
 
     const spans = memoryExporter.getFinishedSpans();
     const minimalSpans = _.map(spans, 'attributes');
-    // Unfortunately, the @memoizedId will be sensitive how many tests in this file ran before it.
-    // To avoid issues with that, we put this test first.
```
NickHeiner, line 79: NICE

packages/examples/test/core/memoize.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const ctx = AI.createRenderContext();
+  const element = ctx.memo(<Component />);
+
+  const frames = [] as string[];
+  const renderResult = ctx.render(element);
+  for await (const frame of renderResult) {
+    frames.push(frame);
+  }
+  expect(frames).toEqual(['', '3', '32', '321']);
+  expect(await renderResult).toEqual('321LIFTOFF');
```
NickHeiner, line 109: lol ugh re: frames not having the final result. (No need to change it now.)

packages/examples/test/core/memoize.tsx
NickHeiner, line 1: Wow this is much nicer than "🎵 it works in the examples 🎵 "

packages/ai-jsx/src/core/memoize.tsx
```diff
@@ -76,32 +83,53 @@ export function partialMemo(renderable: Renderable, existingId?: number): Node |
...
+              if (isAppendOnlyStreamValue(value)) {
+                isAppendOnly = true;
+                value = valueToAppend(value);
+              }
+
+              if (!isAppendOnly) {
+                nodes = [value];
+              } else {
+                nodes.push(value);
+              }
```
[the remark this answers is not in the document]
petersalas: Fair!

packages/ai-jsx/src/core/memoize.tsx
petersalas: Good suggestion -- refactored a bit to make the loop more explicitly only in service to the `isAppendOnly` scenario.

packages/examples/test/core/memoize.tsx
petersalas: the best

PR #244, opened 2023-08-23
Replace `openai-edge` dependency with `openai` v4

petersalas:
- Replace dependency
- Lazily instantiate OpenAI (and Anthropic, for symmetry) clients (OpenAI client initialization fails with an exception if the environment variable is not present)
- Remove OpenAI SSE parsing logic for streams
- (Unrelated) add `outputs` to `turbo.json` so that it can properly restore outputs from the cache


packages/ai-jsx/src/lib/util.ts
```diff
@@ -8,12 +8,22 @@ export function getEnvVar(name: string, shouldThrow: boolean = true) {
...
+  let env = globalThis.process?.env ?? undefined;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (env === undefined) {
+    // Webpack for Create React App requires `process.env`, but in other environments
+    // process might not be defined, so we need to try/catch it.
+    try {
+      env = process.env;
+    } catch {
+      // Do nothing, leave it undefined.
+    }
```
NickHeiner, line 20: Is there a reason we can't just do a `'env' in process` check? The `try/catch` feels a bit awkward.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -115,42 +100,6 @@ export function OpenAI({
   return result;
 }
 
-export const SSE_PREFIX = 'data: ';
```
NickHeiner, line 118: Damn it's nice if we can yank all this.

turbo.json
```diff
@@ -5,7 +5,8 @@
       "dependsOn": ["^build"]
     },
     "build": {
-      "dependsOn": ["^build"]
+      "dependsOn": ["^build"],
+      "outputs": ["dist/**", ".next/**", "!.next/cache/**"]
```
NickHeiner, line 9: heh I knew we weren't actually using Turbo very well – neat if this is the thing that makes it work.

NickHeiner (approved):

packages/ai-jsx/src/lib/util.ts
petersalas: The tricky part is:
- webpack apparently needs to see the literal `process.env` to replace it with the build-time value
- some environments don't have `process` defined at all

...so this was the only way I could figure out how to do it.

PR #246, opened 2023-08-26
Migrate FixieCorpus to the new (beta) corpus API.

mdwelsh:
The new corpus service is currently only deployed on beta.fixie.ai, however, it should be more reliable and useful than the existing app.fixie.ai API. Hence this PR updates AI.JSX's `FixieCorpus` component to use the new API surface. I've also added a little demo program for the component.

If we want to retain support for the old API, I'm happy to rename this component to something like `FixieCorpusNew` or add a flag or something to use the new API instead - feedback welcome!


NickHeiner:
> If we want to retain support for the old API, I'm happy to rename this component to something like FixieCorpusNew or add a flag or something to use the new API instead - feedback welcome!

I vote for simplicity and just cutting over to the new API.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -635,27 +635,31 @@ export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implem
...
         'runtime'
       );
     }
-    const apiResults = await response.json();
-    return apiResults.chunks.map((result: any) => ({
+    const apiResult = await response.json();
+    return apiResult.results.map((result: any) => ({
       chunk: {
-        content: result.content,
-        metadata: result.metadata,
```
NickHeiner, line 657: Will we ever get this back?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -635,27 +635,31 @@ export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implem
...
-        documentName: result.document_name,
+        content: result.snippet,
+        documentName: result.document_id,
```
mdepinet, line 662: This is not a document name.  I think we should probably only surface the chunk content for now and add more fields once we have some translation between DB models and the API we're actually serving.

packages/examples/src/fixie-corpus.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { ScoredChunk, FixieCorpus } from 'ai-jsx/batteries/docs';
+import { showInspector } from 'ai-jsx/core/inspector';
+
+// This is the corpus ID for the docs.ai-jsx.com corpus.
+const FIXIE_CORPUS_ID = '07d42038-0c0c-4108-8b9e-163e0a4ce683';
+
+function ChunkFormatter({ doc }: { doc: ScoredChunk<any> }) {
+  return (
+    <>
+      {'\n\n'}Chunk from source: {doc.chunk.metadata?.source}
```
mdepinet, line 15: Won't this always be null right now? Might make sense to omit this line until we're providing something useful here.

mdepinet (approved):

packages/ai-jsx/src/batteries/docs.tsx
mdwelsh: Yes, but it probably requires some further changes to the new corpus API.

mdwelsh:
Thanks for the comments - PTAL

mdepinet (approved):
LGTM!

packages/examples/src/fixie-corpus.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const query = 'How do I write a chatbot in AI.JSX?';
+  const results = await corpus.search(query, { limit: 4 });
+
+  return results.map((chunk) => <ChunkFormatter doc={chunk} />);
+}
+
+(async () => {
+  const corpusId = await prompt('Fixie Corpus ID: ');
+  showInspector(<App corpusId={corpusId} />);
+})();
```
NickHeiner, line 34: It's 2023 we can just do top-level `await`
```suggestion
  const corpusId = await prompt('Fixie Corpus ID: ');
  showInspector(<App corpusId={corpusId} />);
```

NickHeiner (approved):

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -616,7 +616,7 @@ export class LocalCorpus<
 
 /** A fully mananged {@link Corpus} served by Fixie. */
 export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implements Corpus<ChunkMetadata> {
-  private static readonly DEFAULT_FIXIE_API_URL = 'https://app.fixie.ai/api';
+  private static readonly DEFAULT_FIXIE_API_URL = 'https://beta.fixie.ai/api/v1';
```
NickHeiner, line 619: Should we update the error message below to also point to `beta.fixie.ai`? 

```
'You must provide a Fixie API key to access Fixie corpora. Find yours at https://app.fixie.ai/profile.',
```

PR #247, opened 2023-08-27
First stab at JS-based CLI for Fixie.

mdwelsh:
The idea here is to replace the Python-based "fixie" CLI with an npm-based "fixie" CLI, which will allow one to create and manage Sidekicks and Corpora. This way an AI.JSX developer need not have a Python environment set up. It also provides a JS-based FixieClient class for interacting with the Fixie platform APIs programmatically.

This version only has the Corpus-related commands, we will need to port over the logic from the Python-based "fixie deploy" command for AI.JSX Sidekick deployment to JS as well.


packages/fixie/src/client.ts
```diff
@@ -0,0 +1,156 @@
...
+   * Create a new Octopus Farmer client.
+   * @param url The URL of the Fixie API server.
+   */
+  public static async Create(url: string): Promise<FixieClient> {
+    const client = new FixieClient();
+    client.url = url;
+    const apiKey = process.env.FIXIE_API_KEY;
+    if (!apiKey) {
+      throw new Error(
+        'Must set the FIXIE_API_KEY environment variable. This can be found at: https://app.fixie.ai/profile'
```
NickHeiner, line 21:
```suggestion
        'You must set the FIXIE_API_KEY environment variable. This can be found at: https://app.fixie.ai/profile'
```

packages/fixie/src/client.ts
```diff
@@ -0,0 +1,156 @@
+/**
+ * A client to the Fixie AI platform.
+ */
+export class FixieClient {
+  url!: string;
+  apiKey!: string;
+
+  /** Use Create() instead. */
```
NickHeiner, line 8: If this is the intent, we should mark the constructor as `private`.

packages/fixie/src/client.ts
```diff
@@ -0,0 +1,156 @@
...
+        },
+      });
+    }
+    if (!res.ok) {
+      throw new Error(`Failed to access Fixie API ${this.url}${path}: ${res.statusText}`);
+    }
+    return res.json();
+  }
+
+  // Read the state of the current game.
```
NickHeiner, line 54: Remove this

packages/fixie/src/client.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  async listCorpora(): Promise<any> {
+    return this.request('/api/v1/corpora');
+  }
+
+  async getCorpus(corpusId: string): Promise<any> {
+    return this.request(`/api/v1/corpora/${corpusId}`);
+  }
+
+  async createCorpus(name?: string): Promise<any> {
```
NickHeiner, line 67: Instead of `Promise<any>`, it would be ideal if we could type out the API's return interface.

If that's too onerous, we should use `unknown` or `Jsonifiable` instead of `any`.

NickHeiner:
For reference, here's the Codeshot fixie-sdk TS feature branch: https://github.com/fixie-ai/fixie-sdk/tree/feature-ts.

packages/fixie/tsconfig.json
```diff
@@ -1,13 +1,15 @@
 {
-  "extends": "@tsconfig/node18/tsconfig.json",
```
NickHeiner, line 2: Why did we remove this?

packages/fixie/package.json
```diff
@@ -1,56 +1,32 @@
...
-        "default": "./dist/index.js"
-      },
-      "require": {
-        "default": "./dist/index.cjs"
-      }
-    }
-  }
+	"name": "fixie",
+	"version": "1.0.0",
+	"private": true,
```
NickHeiner, line 4: How are we thinking of distributing this binary? Is it a standalone npm package? Or is it bundled with ai-jsx?

If it's the former, we need to remove `private: true`.
If it's the latter, then we shouldn't make a new monorepo package for the CLI, but should put it instead in `packages/ai-jsx`.

I'm not totally sure which we want. I'm curious to chat through it with @psalas.

packages/fixie/package.json
```diff
@@ -1,56 +1,32 @@
...
+	"license": "MIT",
+	"type": "module",
+	"scripts": {
+		"build": "tsc",
+		"start": "node dist/main.js",
+		"build-start": "yarn run build && yarn run start",
+		"format": "prettier --write .",
+		"lint": "eslint ."
```
NickHeiner, line 12: We're missing the `.eslintrc.cjs` file in this directory.

NickHeiner:
I used this to make a Help Scout corpus: `8d3fff4b-09e0-4ff3-901a-a4c517bee019`. We'll see how it does. :smile:

packages/fixie/src/client.ts
```diff
@@ -0,0 +1,156 @@
...
+ */
+export class FixieClient {
+  url!: string;
+  apiKey!: string;
+
+  /** Use Create() instead. */
+  constructor() {}
+
+  /**
+   * Create a new Octopus Farmer client.
```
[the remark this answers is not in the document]
mdwelsh: No, this was TOTALLY INTENTIONAL

packages/fixie/src/client.ts
mdwelsh: Eventually Mike and I will have some kind of OpenAPI spec generated from the Fixie REST APIs that could be used for this. I'll use Jsonifiable for now.

packages/fixie/tsconfig.json
mdwelsh: Not intentional.

mdwelsh:
Thanks for the comments - PTAL

mdwelsh:
> I think that's just because the service had to spin up, and thus this isn't worth worrying about.

Yep, I'm fixing this.

mdwelsh:
> Is there a single command I can run to get corpus status? (e.g. is it ready to query)

Not yet but I will add that to the backend.

NickHeiner:
image

image

packages/fixie/src/main.ts
```diff
@@ -0,0 +1,141 @@
...
+    term('Created corpus: \n').green(JSON.stringify(result, null, 2));
+  });
+
+corpus
+  .command('query <corpusId> <query>')
+  .description('Query a given corpus.')
+  .action(async (corpusId: string, query: string) => {
+    const client = await FixieClient.Create(program.opts().url);
+    const result = await client.queryCorpus(corpusId, query);
+    term('Query result: \n').green(JSON.stringify(result, null, 2));
```
NickHeiner, line 60: Hmm. When I get a JSON output, I like to be able to pipe it to `jq` for further processing. However, with the current form, I have to do things like `tail` and `sed -E '1d; s/\x1B\[[0-9;]*[mG]//'` to make this `jq`-able.

It would be nice to just dump JSON.

NickHeiner: If you're attached to the nicer output, maybe we could have a `--format json` flag?

mdwelsh:
> image
> image

Hey, the error message is correct! "Missing argument `query`" on your first command line is exactly right: an `argument` is not the same thing as an `option`!

packages/fixie/src/main.ts
mdwelsh: OK, I removed the extra lines at the top, so everything can be piped to jq, and added a `--raw` option which disables pretty-printing. I'd rather have pretty-printing on by default because I *think* most people running this tool will want that rather than hunt for an option...

packages/fixie/package.json
NickHeiner: If we're moving the whole CLI SDK to node, then a standalone `fixie` package seems right to me.

NickHeiner (approved):

PR #251, opened 2023-08-30
Separate fixie-hello-world into fixie-hello-world and @fixieai/sdk

petersalas:
- Split out the fastify server from the hello-world sample
- Use new conversation contracts with Fixie service
- Add JWT parsing to the fastify shim to extract agent ID
- Change AI.JSX `<ConversationHistory>` component to be a simple extensibility point that the Fixie request wrapper injects the `<FixieConversation>` component into

Subsequent changes will:
- Plumb the auth token so that Fixie corpus queries can use it


packages/fixie-hello-world/package.json
```diff
@@ -2,55 +2,19 @@
...
-    "eslint": "^8.40.0",
-    "eslint-config-nth": "^2.0.1",
     "typescript": "^5.1.3"
   },
   "private": true,
   "type": "module",
   "scripts": {
-    "dev": "yarn build && npx ts-node dist/serve-bin.js --port 8080 --package-path fixie",
-    "lint": "eslint . --max-warnings 0",
-    "lint:fix": "eslint . --fix",
```
NickHeiner, line 22: Why not lint this package?

packages/fixie-sdk/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "devDependencies": {
+    "@tsconfig/node18": "^2.0.1",
+    "@types/lodash": "^4.14.195",
+    "@types/node": "^20.3.1",
+    "@types/prompt-sync": "^4.2.0",
+    "@types/uuid": "^9.0.2",
+    "@types/yargs": "^17.0.24",
+    "@typescript-eslint/eslint-plugin": "^5.60.0",
+    "@typescript-eslint/parser": "^5.60.0",
+    "ai-jsx": "*",
```
NickHeiner, line 49: Don't we want the `workspace:` option here?

packages/fixie-sdk/src/conversation.tsx
```diff
@@ -0,0 +1,52 @@
+/// <reference lib="dom" />
+import * as AI from 'ai-jsx';
+import { GetConversationResponse } from './types.js';
+import { AssistantMessage, FunctionCall, FunctionResponse, UserMessage } from 'ai-jsx/core/conversation';
+import { fixieContext } from './request-wrapper.js';
+
+export async function FixieConversation(_: {}, { getContext }: AI.ComponentContext) {
+  const fixieContextValue = getContext(fixieContext);
+  if (!fixieContextValue) {
+    throw new Error('FixieConversation components may only be used in the context of requests from Fixie.');
```
NickHeiner, line 10: If we were in AI.JSX, it would be nice to use the AI.JSX more structured errors here. (Which I know you also don't see a ton of value in. 🤪 )

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,125 @@
...
+async function serve({
+  packagePath,
+  port,
+  silentStartup,
+}: {
+  packagePath: string;
+  port: number;
+  silentStartup: boolean;
+}): Promise<void> {
+  const Handler = (await import(path.join(process.cwd(), packagePath))).default;
```
NickHeiner, line 25: Seems like we should have a robust error message if this lookup fails.

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const app: FastifyInstance = fastify();
```
NickHeiner, line 26: Boo hiss that this type annotation is necessary.

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const fixieApiHost = process.env.FIXIE_API_HOST ?? 'https://app.fixie.ai';
```
NickHeiner, line 28: uber-nit: `host` doesn't include the `protocol`. What if we name this `BASE_URL` instead?

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const getJwks = createRemoteJWKSet(new URL(`${fixieApiHost}/.well-known/jwks.json`));
+
+  app.addHook('onRequest', async (request: FastifyRequest, reply: FastifyReply) => {
```
NickHeiner, line 32: Wow I guess the Fastify types suck?

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,125 @@
...
+    try {
+      const token = request.headers.authorization?.split(' ')[1];
+      if (typeof token !== 'string') {
+        reply.send(new Error('No authorization header'));
```
NickHeiner, line 36: Shouldn't this be a `4xx` response? Will this code do that?

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,125 @@
...
+      return res
+        .status(200)
+        .type('application/jsonl')
+        .send(
+          Readable.from(
+            (async function* () {
+              while (true) {
+                try {
+                  const next = await generator.next();
+                  const messages = next.value.split('\n').slice(0, -1);
```
NickHeiner, line 74:
```suggestion
                  const messages = _.last(next.value.split('\n'));
```

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,125 @@
...
+      required: true,
+    },
+    silentStartup: {
+      describe: 'Do not log on startup',
+      type: 'boolean',
+      default: false,
+    },
+    packagePath: {
+      describe:
+        'Path to the package to serve functions from. If this is a relative path, it will be interpreted relative to the current working directory.',
```
NickHeiner, line 114:
```suggestion
        'Path to the package the AI.JSX program. If this is a relative path, it will be interpreted relative to the current working directory.',
```

packages/fixie-sdk/src/serve-bin.tsx
NickHeiner, line 25:
```suggestion
  const Handler = (await import(path.relative(process.cwd(), packagePath))).default;
```

To match the stated behavior of this option below.

packages/fixie-hello-world/package.json
petersalas: The linter config/machinery/dependencies were significantly larger than the meaningful contents of the package so I removed them to try to maximize the signal to noise ratio. Was that a mistake?

NickHeiner (approved):

packages/fixie-hello-world/package.json
NickHeiner: I like to lint all day erry day but it's not a big deal. For having 10 lines of real code it's probably fine to skip; TS gives us basic protection anyway.

packages/fixie-sdk/src/serve-bin.tsx
petersalas: Oops, I was just trying to be consistent -- will change to `FIXIE_API_URL`.

packages/fixie-sdk/package.json
petersalas: I thought so as well, but vercel build failed when I had that...so I dropped it.

packages/fixie-sdk/src/serve-bin.tsx
petersalas: Uhh, or these were all just unnecessary

packages/fixie-sdk/src/serve-bin.tsx
petersalas: It will now :)

packages/fixie-sdk/src/serve-bin.tsx
petersalas: `path.relative` computes a relative path (which is not what we want to pass to `import`). But `path.resolve` is what I should've used in the first place :)

packages/fixie-hello-world/src/index.tsx
```diff
@@ -1,10 +1,10 @@
-import { ChatCompletion, SystemMessage, UserMessage } from 'ai-jsx/core/completion';
+import { ChatCompletion, SystemMessage, ConversationHistory } from 'ai-jsx/core/completion';
 
-export default function HelloWorld({ message }: { message: string }) {
+export default function HelloWorld() {
   return (
     <ChatCompletion temperature={1}>
-      <SystemMessage>Respond to the user using some variant of the phrase "Hello World!". Be creative!</SystemMessage>
-      <UserMessage>{message}</UserMessage>
+      <SystemMessage>Respond to the user as if you were Clippy from Microsoft Office.</SystemMessage>
```
juberti, line 6: Maybe say "You are Clippy from Microsoft Office. Respond to the user accordingly."

packages/fixie-sdk/src/request-wrapper.tsx
```diff
@@ -0,0 +1,79 @@
+import * as AI from 'ai-jsx';
+import { InvokeAgentRequest } from './types.js';
+import { ShowConversation, ConversationHistoryContext } from 'ai-jsx/core/conversation';
+import { Json } from './json.js';
+import { FixieConversation } from './conversation.js';
+
+export interface FixieRequestContext {
+  request: InvokeAgentRequest;
+  agentId: string;
+  fixieApiHost: string;
```
juberti, line 10: why is this prefaced with `fixie` but not the others (e.g., `authToken`)

packages/fixie-sdk/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export interface TextMessage extends MessageBase {
+  kind: 'text';
+  content: string;
+}
+
+export type Message = FunctionCallMessage | FunctionResponseMessage | TextMessage;
+
+export interface ConversationTurn {
```
juberti, line 25: Did we decide not to pursue turns being a full user->assistant exchange?

packages/fixie-sdk/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  id: string;
+  timestamp: string;
+  role: 'user' | 'assistant';
+  messages: Message[];
+  state: 'in-progress' | 'done' | 'stopped' | 'error';
+  metadata?: Record<string, string | number | boolean | object | null> | null;
```
juberti, line 31: is this where things like tz offset will get stuffed?

packages/fixie-sdk/src/serve-bin.tsx
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  const address = await app.listen({ host: '0.0.0.0', port });
+  if (!silentStartup) {
+    console.log(`AI.JSX agent listening on ${address}.`);
+  }
+}
+
+const { argv } = yargs(hideBin(process.argv))
+  .scriptName('cli-tool')
```
juberti, line 99: is this the right name?

packages/fixie-sdk/src/types.ts
petersalas: I ended up modeling this by pointing the assistant turn back to the user turn via `reply_to_turn_id` -- the main thing that was gained by that was that regenerations could have new turn IDs while pointing to the original/unchanged user turn.

packages/fixie-sdk/package.json
NickHeiner: ugh ok

NickHeiner:
When I used this with a Sidekick and had a bug where a `Tool`'s `func` returned `undefined`, I got this error:

```
sidekick-github:dev: Error during generation: SyntaxError: Unexpected token } in JSON at position 72
```

When I logged out the value of `next.value`, I saw it was:

```
{"kind": "functionResponse","name": "runGithubGraphqlQuery","response": }
```

packages/fixie-sdk/src/fixie-serve-bin.tsx
```diff
@@ -0,0 +1,133 @@
...
+              while (true) {
+                try {
+                  const next = await generator.next();
+                  const messages = next.value.split('\n').slice(0, -1);
+                  yield `${JSON.stringify({ messages: messages.map((msg) => JSON.parse(msg)) })}\n`;
+                  if (next.done) {
+                    break;
+                  }
+                } catch (ex) {
+                  console.error(`Error during generation: ${ex}`);
```
NickHeiner, line 88:
```suggestion
                  console.error(`Error during generation: ${ex} ${ex.stack}}`);
```

juberti (approved):

packages/fixie-sdk/src/types.ts
NickHeiner: This was covered by our in-person convo, right?

petersalas: Yes, though I didn't add the time zone offset context thing.

I did wire up the OpenAI model selection and proxy though.

PR #254, opened 2023-08-31
Ensure the CLI reports the package version correctly.

mdwelsh:


packages/fixie/package.json
```diff
@@ -1,12 +1,13 @@
...
   "name": "@fixieai/fixie",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "license": "MIT",
   "repository": "fixie-ai/ai-jsx",
   "bugs": "https://github.com/fixie-ai/ai-jsx/issues",
   "homepage": "https://fixie.ai",
   "type": "module",
   "scripts": {
+    "prebuild": "node -p \"'export const LIB_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
```
NickHeiner, line 10: Why not just read `package.json` directly?

Either:

```ts
require('../package').version
```

Or, if that CJS complains, then:

```ts
const currentPath = path.dirname(fileURLToPath(import.meta.url));
const packageJsonPath = path.resolve(currentPath, '../package.json');
const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
packageJson.version
```

PR #255, opened 2023-09-01
Add support for agents and agent deployment to the Node-based Fixie CLI.

mdwelsh:
This PR allows us to deploy Fixie Agents from the node-based CLI. It also adds additional commands for viewing and manipulating agents.

There are also a few cleanups and fixes.


packages/fixie/src/agent.ts
```diff
@@ -0,0 +1,288 @@
...
+            agentId
+          }
+        }
+      `,
+    });
+    return Promise.all(
+      result.data.allAgents.map(async (agent: any) => {
+        const retAgent = await this.GetAgent(client, agent.agentId);
+        return retAgent;
+      })
```
NickHeiner, line 82: Minor, but this can just be:

```
result.data.allAgents.map(agent => this.GetAgent(client, agent.agentId))
```

PR #259, opened 2023-09-06
Fix FIXIE_API_URL bug in FixieCorpus. Bump package version.

mdwelsh:
Fixing both of the bugs I created. Will YOLO this in unless you squawk.


NickHeiner (approved):
LGTM on a quick glance

PR #260, opened 2023-09-06
Add support for deployment environment variables

petersalas:
- In `@fixieai/fixie` add a `-e` to option to set deployment environment variables, and set `FIXIE_API_URL` to the instance being invoked.
- In `@fixieai/sdk` use `new URL` to construct the correct JWKS URL.

This obviates the need to create a `.env` file to point at non-prod instances.

### DIFF
--- a/packages/fixie/src/main.ts
+++ b/packages/fixie/src/main.ts
@@ -4,7 +4,7 @@
  * This is a command-line tool to interact with the Fixie platform.
  */
 
-import { program } from 'commander';
+import { Command, program } from 'commander';
 import terminal from 'terminal-kit';
 import { FixieClient } from './client.js';
 import { FixieAgent } from './agent.js';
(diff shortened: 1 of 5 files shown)


mdwelsh (approved):
Noice

NickHeiner (approved):

packages/fixie/src/main.ts
```diff
@@ -24,9 +24,24 @@ function showResult(result: any, raw: boolean) {
...
 
 /** Deploy an agent from the current directory. */
-async function deployAgent(path?: string) {
-  const client = await FixieClient.Create(program.opts().url);
-  await FixieAgent.DeployAgent(client, path ?? process.cwd());
+function registerDeployCommand(command: Command) {
+  command
+    .command('deploy [path]')
+    .description('Deploy an agent')
+    .option('-e, --env <key=value>', 'Environment variables to set', (v, m: Record<string, string> | undefined) => {
```
NickHeiner, line 31: How do `-e` params interact with a `.env` file? Which will take precedence? It would be nice to be explicit about that in this description to the user.

petersalas: Added a longer description. Eventually I think we should move away from `.env` but these currently need to be specified with every `fixie deploy` so I won't push for that just yet.

PR #262, opened 2023-09-07
patching corpus query

benlower:
Changing the body to use 'max_chunks'.

### DIFF
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -645,7 +645,7 @@ export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implem
       body: JSON.stringify({
         corpus_id: this.corpusId,
         query,
-        page_size: params?.limit ?? 10,
+        max_chunks: params?.limit ?? 10,
       }),
     });
     if (response.status !== 200) {
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -658,7 +658,7 @@ export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implem
     const apiResult = await response.json();
     return apiResult.results.map((result: any) => ({
       chunk: {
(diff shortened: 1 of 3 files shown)


packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -645,7 +645,7 @@ export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implem
       body: JSON.stringify({
         corpus_id: this.corpusId,
         query,
-        page_size: params?.limit ?? 10,
+        max_chunks: params?.limit ?? 10,
```
mdepinet, line 648: We should really be using camelCase I think, but if this works then I suppose I don't really care either way. :)

mdepinet (approved):
Thanks!

packages/ai-jsx/src/batteries/docs.tsx
benlower: i don't know what the right way is. i just copied the format that was used here https://github.com/fixie-ai/fixie/blob/4d7f23adaa585a01e408f8f68a05574800a9db08/src/proto_api/fixie_api/corpus/v1/service.proto#L152C22-L152C22

mdepinet: Fair enough. The OpenAPI spec we generate uses camelCase for field names, but our server will parse either one.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -645,7 +645,7 @@ export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implem
...
       }),
     });
     if (response.status !== 200) {
```
mdepinet, line 651: Could you update line 661 to use result.chunkContent while you're here please?

mdepinet (approved):

mdwelsh (approved):

packages/ai-jsx/src/batteries/docs.tsx
benlower: done

PR #263, opened 2023-09-07
Add Text-to-speech classes and demo page

juberti:
Currently has streaming versions of Eleven Labs and Azure TTS, and infrastructure to easily allow adding more.

Screenshot 2023-09-07 at 3 04 30 PM

PR #264, opened 2023-09-08
Audio Sidekick demo

juberti:
Stacks on top of #263


package.json
```diff
@@ -17,6 +17,7 @@
     "yarn": "3.6.0"
   },
   "scripts": {
+    "nd": "cd packages/ai-jsx && yarn build && cd ../../packages/nextjs-demo && yarn dev",
```
NickHeiner, line 20: I think this is redundant with:

```
$ yarn turbo run dev --scope nextjs-demo
```

packages/nextjs-demo/src/app/agent/page.tsx
```diff
@@ -0,0 +1,242 @@
...
+              <li>VANILLA SPECIALTY LATTE $3.49</li>
+              <li>ORIGINAL GLAZED® LATTE $3.49</li>
+              <li>CARAMEL SPECIALTY LATTE $3.49</li>
+              <li>CARAMEL MOCHA SPECIALTY LATTE $3.49</li>
+              <li>MOCHA SPECIALTY LATTE $3.49</li>
+            </ul>
+          </div>
+        </div>
+        <div>
+          <textarea className="m-2 w-full" readOnly id="output" rows={10} cols={80} value={output}></textarea>
```
NickHeiner, line 224: Minor: this is what the agent says, and not actual input, right? In that case, I'd use a `div` instead of `textarea`. https://tailwindui.com/components has examples you can copy/paste from. (Ask Zach for access if you don't already have it.)

This is only worth changing if it takes <5 minutes.

NickHeiner (approved):

PR #266, opened 2023-09-08
API Token Limiting

farzadab:
This PR adds automatic handling of large API responses by chunking them and exposing a `loadBySimilarity` function to LLM based on Cohere reranker. The reranker is proxied through the Fixie server (PR https://github.com/fixie-ai/fixie/pull/1573).

There are still some TODOs that I will get rid of, but I would love to have reviewers' thoughts on it as well.

TODO:
- [x] remove and finalize TODOs
- [x] fix test errors
- [x] before merge: add changelog and update versions (`fixie-sdk` and `ai-jsx`)

image


packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,192 @@
...
+    lengthFunction = openAITokenCount,
+    ...props
+  }: AI.PropsOfComponent<typeof FunctionResponse> & {
+    maxLength: number;
+    numChunks?: number;
+    lengthFunction?: LengthFunction;
+  },
+  { render, logger, getContext }: AI.ComponentContext
+) {
+  let stringified = await render(children);
```
farzadab, line 55: For some reason I didn't see any issues with rendering tool responses that might be an object (i.e. parsed JSON vs stringified).

@petersalas since this might be relevant to your existing PR, just wanted to get you to double check this part.

https://fixie-ai.slack.com/archives/C05B6EE9JE5/p1694191388802549

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -77,27 +78,47 @@ export async function getNextConversationStep(
...
       const { name, args } = lastMessage.element.props;
+      let response;
       try {
-        return <FunctionResponse name={name}>{await tools[name].func(args)}</FunctionResponse>;
+        response = await updatedTools[name].func(args);
       } catch (e: any) {
         return (
           <FunctionResponse failed name={name}>
-            {e.message}
+            <TruncateByChars maxLength={2000}>{e.message}</TruncateByChars>
```
farzadab, line 100: This might be confusing, but here `maxLength` is based on characters, but in `LargeFunctionResponseHandler` it's based on tokens by default (i.e. `lengthFunction` is set to `tokenCounter`). I wanted to allow both chars and tokens here but the effort didn't seem worth it at the time. I can take another pass at it though if need be.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,192 @@
...
+import { UseToolsProps } from '../../use-tools.js';
+import { cohereContext, reranker } from '../../../lib/cohere.js';
+
+export type LengthFunction = ((text: string) => number) | ((text: string) => Promise<number>);
+export interface RedactedFuncionResponseMetadata {
+  isRedacted: true;
+  chunks: string[];
+}
+
+const getEncoder = _.once(() => getEncoding('cl100k_base'));
```
farzadab, line 17: I'm arbitrarily using OpenAI's tokenizer here. Trying to tie this back to the actual model being used didn't seem worth it.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,192 @@
...
+  }
+  return `${stringified.slice(0, maxLength - 3)}...`;
+}
+
+export async function LargeFunctionResponseHandler(
+  {
+    children,
+    maxLength,
+    numChunks = 4,
+    // use Cohere token counter?
```
farzadab, line 45: If we want to optimize for Cohere, we can use their tokenizer and set the `chunkLimit` to 500 (that way the capacity of reranker becomes 500K tokens (less the overlap).

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,192 @@
...
+  });
+  const chunks = await splitter.splitText(stringified);
+
+  return (
+    <FunctionResponse {...props} metadata={{ isRedacted: true, chunks }}>
+      ... The response is too big and hence redacted. The response can be queried using semantic similarity search by
+      calling the `loadBySimilarity` function.
+    </FunctionResponse>
+  );
+  // TODO: return a chunk based on query (gotta get it from conversation)
```
farzadab, line 94: I'd rather not to do this, but @juberti if you still want it I can take another stab at it.

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -63,6 +64,18 @@ export function present(conversationElement: ConversationMessage) {
...
+/**
+ * Will filter the conversation to only include the current round of messages
+ * @returns A subset of the full conversation, starting from the last user message.
+ */
+function getCurrentRound(fullConversation: ConversationMessage[]) {
+  const lastUserMessage = fullConversation.findLastIndex((m) => m.type === 'user');
+  if (lastUserMessage === -1) {
+    return fullConversation;
+  }
+  return fullConversation.slice(lastUserMessage);
```
NickHeiner, line 76: minor
```suggestion
  return _.takeRightWhile(fullConversation, ({type}) => type !== 'user');
```

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -81,15 +94,39 @@ export function getNextConversationStep(
...
+  // will need to call the function again. This is to prevent the LLM from accessing stale data.
+  const updatedTools = { ...tools, ...redactedFunctionTools(getCurrentRound(fullConversation)) };
+
   switch (lastMessage.type) {
     case 'functionCall': {
       const { name, args } = lastMessage.element.props;
-      return <ExecuteFunction func={tools[name].func} name={name} args={args} />;
+      const executedFunction = (
+        <ExecuteFunction
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
NickHeiner, line 108: Are the types wrong here?

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { ConversationMessage, FunctionResponse, FunctionResponseProps } from '../../../core/conversation.js';
+
+import { RecursiveCharacterTextSplitter } from 'langchain/text_splitter';
+import { getEncoding } from 'js-tiktoken';
+import yaml from 'js-yaml';
+import _ from 'lodash';
+import { UseToolsProps } from '../../use-tools.js';
+import { cohereContext, MarkdownChunkFormatter, RerankerFormatted } from '../../../lib/cohere.js';
+
+export type LengthFunction = ((text: string) => number) | ((text: string) => Promise<number>);
```
NickHeiner, line 11:
```suggestion
export type LengthFunction = (text: string) => Promisable<number>
```

https://github.com/sindresorhus/type-fest/blob/main/source/promisable.d.ts

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,228 @@
...
+  let stringified = await render(children);
+
+  // Option 1: do nothing if it's already small enough
+  if ((await lengthFunction(stringified)) <= maxLength) {
+    return <FunctionResponse {...props}>{stringified}</FunctionResponse>;
+  }
+
+  stringified = yamlOptimizeIfPossible(stringified);
+
+  // Option 2: try dumping as YAML, if it's small enough then we are done
```
NickHeiner, line 77:
```suggestion
  // Option 2: try dumping as YAML. If it's small enough, then we are done.
```

uber-nit

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,228 @@
...
+  if ((await lengthFunction(stringified)) <= maxLength) {
+    return <FunctionResponse {...props}>{stringified}</FunctionResponse>;
+  }
+
+  // Option 3 (last reosrt): split into chunks and allow LLM to query by similarity
+  // Requires Cohere API key for doing similarity search
```
NickHeiner, line 83: Hmm. I feel like, if the Cohere API key isn't set, it would be better to just truncate the response.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,228 @@
...
+      Array.isArray(msg.element.props.metadata.chunks) &&
+      msg.element.props.metadata.chunks.every((chunk) => typeof chunk === 'string')
+    ) {
+      lastFnResData = msg.element.props.metadata as unknown as RedactedFuncionResponseMetadata;
+    }
+  }
+  if (lastFnResData === undefined) {
+    return undefined;
+  }
+  return lastFnResData;
```
NickHeiner, line 166:
```suggestion
  return lastFnResData;
```

Maybe I'm missing something, but that first condition seems unnecessary.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,228 @@
...
+    if (
+      msg.type == 'functionResponse' &&
+      typeof msg.element.props.metadata !== 'undefined' &&
+      'isRedacted' in msg.element.props.metadata &&
+      Boolean(msg.element.props.metadata.isRedacted) &&
+      'chunks' in msg.element.props.metadata &&
+      Array.isArray(msg.element.props.metadata.chunks) &&
+      msg.element.props.metadata.chunks.every((chunk) => typeof chunk === 'string')
+    ) {
+      lastFnResData = msg.element.props.metadata as unknown as RedactedFuncionResponseMetadata;
```
NickHeiner, line 160: Hmm, it's a little odd that the function is named `lastFnResData`, but this loop will continue searching backwards until it finds the first element meeting the criteria. Did you want a `findLast` call instead?

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+  if (lastFnResData === undefined) {
+    return undefined;
+  }
+  return lastFnResData;
+}
+
+export function redactedFunctionTools(messages: ConversationMessage[]): UseToolsProps['tools'] {
```
NickHeiner, line 169: Do we ever anticipate having more than one tool? Having machinery for multiple tools seems like overkill if we only ever actually use a single tool?

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,228 @@
...
+ * Also see the following for more comparisons:
+ *     [Internal] https://www.notion.so/fixieai/API-Response-Token-Limiting-2ba2a63b047044599370c2f26fcf2bfa
+ *     [Public] https://nikas.praninskas.com/ai/2023/04/05/efficient-gpt-data-formats/
+ *     [Public] https://betterprogramming.pub/yaml-vs-json-which-is-more-efficient-for-language-models-5bc11dd0f6df
+ *
+ * @param possiblyObjectText The text to optimize (in JSON/YAML format)
+ * @returns Optimized text if possible, otherwise the original text
+ */
+function yamlOptimizeIfPossible(possiblyObjectText: string) {
+  let content: any;
```
NickHeiner, line 217:
```suggestion
  let content: Jsonifiable;
```

from `type-fest`

packages/ai-jsx/src/lib/cohere.tsx
```diff
@@ -0,0 +1,145 @@
...
+  { getContext, logger }: Pick<AI.ComponentContext, 'getContext' | 'logger'>
+): Promise<{ index: number; relevance_score: number }[]> {
+  const cohereConfig = getContext(cohereContext);
+  if (!cohereConfig.api_key) {
+    logger.error(
+      { CohereContext: cohereConfig },
+      'Cohere API key is not set, but it is needed for doing reranking (e.g. loadBySimilarity).' +
+        ' Please set it in the context.'
+    );
+    // fallback to original order
```
NickHeiner, line 62:
```suggestion
    // fall back to original order
```

"fallback" is a noun; "fall back" is a verb

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: We used to have `loadByExpression` and `loadRemaining` (e.g. the functions can be conditioned based on the nature of the results), so the idea was to allow adding that back in if needed.

Also, handling 0 vs 1 tool seemed easier this way which also allows for more tools later on.

packages/ai-jsx/src/lib/cohere.tsx
```diff
@@ -0,0 +1,145 @@
...
+    // TODO: maybe try OpenAI too?
+    return [...Array(Math.min(documents.length, top_n)).keys()].map((index) => ({ index, relevance_score: 0 }));
```
NickHeiner, line 64:
```suggestion
    const countResultDocs = Math.min(documents.length, top_n);
    return _.range(countResultDocs).map(index => ({ index, relevance_score: 0 }))
```

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
farzadab: I'm not sure why TypeScript doesn't throw an error here. There is no reason why `name` should be available in `tools`.
I also noticed that @petersalas had handled the case where `func` was `undefined`, but from what I understand, this line will error out before getting to that point since `tool[name]` can be the thing that is `undefined` (usually if the tool exists, the `func` also exists, but no guarantee that the `tool` exists).

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: > I feel like, if the Cohere API key isn't set

We do set it using a context inside `request-wrapper.tsx`, so Fixie sidekicks will have it set. Also we do have a default context that uses `COHERE_API_KEY` env var. Is there anywhere I should document this?

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: Huh you're right.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: Yes. At first this was implemented with a `.findLastIndex`, but type checking issues and I didn't want to have duplicate code. I'll take another look since I see I'm already doing `as unknown as RedactedFuncionResponseMetadata` anyway.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: But LC uses this signature and TypeScript doesn't understand that they are the same 😢

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: cc @hessamb

hessamb: I think that's a good idea. But falling back to truncate by char is also a good interim solution for hackathon (since we expect cohere API to be set anyway).

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
NickHeiner: Bummer. Ok, that's fine.

What if we tried this then?

```suggestion
export type LengthFunction = ((text: string) => number | Promise<number>);
```

NickHeiner:
No need to block this PR on my approval.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: Same issue.

I might end up ejecting (i.e. copy&pasting) the text splitter from LC. If I do that I'll make sure to use `Promisable` instead.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,222 @@
...
+  failedMaxLength: number;
+  numChunks: number;
+  lengthFunction?: LengthFunction;
+}
+
+async function LargeFunctionResponseHandler(
+  {
+    children,
+    maxLength = 4000,
+    failedMaxLength = 1000,
```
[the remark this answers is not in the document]
farzadab: No real reason. I just thought that if there is an error, we might not want to pollute the context window with it. This might not be an important concern since the effects won't be linear due to `Shrinkable`.

packages/ai-jsx/src/lib/cohere.tsx
```diff
@@ -0,0 +1,147 @@
...
+  if (!cohereConfig.api_key) {
+    logger.error(
+      { CohereContext: cohereConfig },
+      'Cohere API key is not set, but it is needed for doing reranking (e.g. loadBySimilarity).' +
+        ' Please set it in the context.'
+    );
+    // fall back to original order
+    // TODO: maybe try OpenAI too?
+    const countResultDocs = Math.min(documents.length, top_n);
+    return _.range(countResultDocs).map((index) => ({ index, relevance_score: 0 }));
```
petersalas, line 66: This seems like the wrong place for this fallback -- it seems like this function should probably just be the wrapper around the API?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -111,6 +111,13 @@ export function FunctionCall({
   return `Call function ${name} with ${partial ? '(incomplete) ' : ''}${JSON.stringify(args)}`;
 }
 
+export interface FunctionResponseProps {
```
petersalas, line 114: Assuming it's not a large number of places that needs this, did you consider `AI.PropsOfComponent<typeof FunctionResponse>`?

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,222 @@
...
+ *      <ExecuteFunction func={tool[name].func} name={name} args={args} />
+ *    </LargeFunctionResponseWrapper>
+ * ```
+ */
+export async function LargeFunctionResponseWrapper(
+  { children, ...props }: { children: AI.Node } & LargeFunctionResponseProps,
+  { render }: AI.ComponentContext
+) {
+  // We need to render the children to get the FunctionResponse elements
+  const elements = await render(children, { stop: (e) => e.tag == FunctionResponse });
```
petersalas, line 137: I think using `renderToConversation` and then replacing just the `<FunctionResponse>`s  is preferable. Would rather minimize the proliferation of partial rendering.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const splitter = new RecursiveCharacterTextSplitter({
+    chunkSize: maxLength / numChunks,
+    chunkOverlap: maxLength / numChunks / 10,
+    lengthFunction,
+  });
+  const chunks = await splitter.splitText(stringified);
+
+  return (
+    <FunctionResponse {...props} metadata={{ isRedacted: true, chunks }}>
+      ... The response is too big and hence redacted. The response can be queried using semantic similarity search by
```
petersalas, line 105: Will having multiple such `<FunctionResponse>s` in the conversation be a problem?

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
petersalas: I guess the dumbest/simplest option to consider would just be to use string length?

packages/fixie-sdk/src/request-wrapper.tsx
```diff
@@ -4,6 +4,7 @@ import { ShowConversation, ConversationHistoryContext } from 'ai-jsx/core/conver
 import { Json } from './json.js';
 import { FixieConversation } from './conversation.js';
 import { OpenAI, OpenAIClient, ValidChatModel } from 'ai-jsx/lib/openai';
```
petersalas, line 6: will need to rev fixie-sdk as well

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: I don't think so. We only present the model with the last one, and LLM gets at least one turn to react to it (also the `loadBySimilarity` itself will not trigger redaction so it can be called repeatedly). Is there any other scenario you were thinking about?

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: Hmm, I agree that that's more natural and less opinionated (which is preferred for a library). In practice though I kinda want to use OpenAI's tokenizer for sidekicks.

Specially because we probably want to maintain the parity between chunks from a corpus and chunks from an API (easier to reason about I guess?).


Idea: What if we push `tokenizer`, `lengthFunction`, and such to the context too?

packages/ai-jsx/src/lib/cohere.tsx
farzadab: You're right. Removed the fallback altogether.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
petersalas: To be clear I agree with you that just using OpenAI is the right choice vs. doing largely performative plumbing -- was just throwing out an "even simpler" option for you to (re)consider :)

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
petersalas: WDYM we only present the model with the last one -- won't this be persisted in the conversation? And re:side-by-side:

- user
- functionCall
- functionResponse
- functionCall
- functionResponse
- assistant

is certainly a valid scenario, but I'm likely misunderstanding something.

A reasonable answer might be just that the model is smart enough to infer that the last one takes precedence over previous ones, but I was curious if this is something that either can't happen (i.e. I'm misunderstanding things) or you've played around with and is fine.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: Done.

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
farzadab: Ooh sorry yeah you're right. I'm assuming that the model is smart enough. I have certainly used multiple of these in the same conversation and didn't have any issues.

But this is a valid point. I'll play with the descriptions a bit more to make it a bit more obvious.

farzadab: I improved the function description. For now I don't have better ideas to make sure this doesn't happen. I say we can assume that it works until we see any issues.
I put a TODO in the code to track this concern.

WDYT?

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -739,21 +739,23 @@ const defaultLangchainChunkLimit = 4;
...
+      return {
+        score: lcDocAndScore[1],
+        chunk: {
+          // TODO: Wrap chunker to track document name in ChunkMetadata in a way we can pull back out here.
+          content: lcDoc.pageContent,
+          metadata: lcDoc.metadata as ChunkMetadata,
+        },
+      } as ScoredChunk<ChunkMetadata>;
+    })
+    .filter((chunk) => chunk.score >= (params?.score_threshold ?? Number.MIN_VALUE));
```
farzadab, line 758: @mdepinet here is my workaround fix for the type issue with newer versions of LC. The issue is that `VectorStore` doesn't take a single `filter` type as input anymore. It's dependent on the type of store that you use, e.g. Pinecone still takes a dictionary but `MemoryStore` takes a filter function as input and it throws an error if we provide an `object`.

Unfortunately, because of the way they do the typing, we don't get any TypeScript errors either way (so it can't easily be handled at compile time) and at runtime the types don't exist so we cannot decide what to pass in.
I decided to kinda put the burden on the user by saying that `params.filter` is what they provide and if it doesn't work then it's their fault.
This makes the interface a bit weird, but since nobody is currently using it, this should probably be fine. We can revisit later. WDYT?

mdepinet: Seems like a reasonable approach to me. At the end of the day, we can only do so much to spare people from LC's garbage types if they want to use LC.

packages/fixie-sdk/src/request-wrapper.tsx
farzadab: Done. Updated package versions (`ai-jsx` and `fixie-sdk`) and Changelog.

farzadab:
PTAL, if there are no more comments, I'd like to merge this.

packages/ai-jsx/src/lib/cohere.tsx
```diff
@@ -0,0 +1,146 @@
...
+    body: JSON.stringify({
+      query,
+      documents,
+      model: cohereConfig.reranker_model_name ?? 'rerank-multilingual-v2.0',
+      top_n,
+    }),
+  });
+
+  if (logger) {
+    logger.info({ response }, 'Got response from reranker server');
```
petersalas, line 80: nit: should be `logger.debug` or `logger.trace`

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
petersalas: Works for me

petersalas (approved):

packages/ai-jsx/src/batteries/sidekick/platform/large-response-handler.tsx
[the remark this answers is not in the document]
farzadab: I'm resolving this to merge the PR, but if you have concerns let's address them in follow up PRs.

PR #269, opened 2023-09-08
Add support for token-less FixieCorpus tool

petersalas:
This change:
- Changes tools to be `Component<any>` (which is compatible with the existing signature) so they can access context and/or return other AI.JSX nodes
- Adds an `<ExecuteFunction>` component that encapsulates error handling so it can be shared between the two `UseTools` implementations
- Adds `FixieAPIContext` to AI.JSX where the Fixie API url/token can be configured
- Adds `FixieCorpus.createTool` which facilitates `FixieCorpus`-based `Tool` creation and reads the API configuration from context (rather than from `process.env`).

### DIFF
--- a/packages/ai-jsx/src/batteries/docs.tsx
+++ b/packages/ai-jsx/src/batteries/docs.tsx
@@ -18,6 +18,8 @@ import * as AI from '../index.js';
 import { Node } from '../index.js';
 import { getEnvVar } from '../lib/util.js';
 import { JsonChatCompletion } from './constrained-output.js';
+import { DEFAULT_API_CONFIGURATION, FixieAPIConfiguration, FixieAPIContext } from './fixie.js';
+import { Tool } from './use-tools.js';
 
 /**
  * A raw document loaded from an arbitrary source that has not yet been parsed.
@@ -616,31 +618,29 @@ export class LocalCorpus<
 
 /** A fully mananged {@link Corpus} served by Fixie. */
 export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implements Corpus<ChunkMetadata> {
-  private static readonly DEFAULT_FIXIE_API_URL = 'https://beta.fixie.ai';
+  private readonly fixieApiConfiguration: FixieAPIConfiguration;
 
-  private readonly fixieApiUrl: string;
-
-  constructor(private readonly corpusId: string, private readonly fixieApiKey?: string) {
-    if (!fixieApiKey) {
-      this.fixieApiKey = getEnvVar('FIXIE_API_KEY', false);
-      if (!this.fixieApiKey) {
-        throw new AIJSXError(
-          'You must provide a Fixie API key to access Fixie corpora. Find yours at https://beta.fixie.ai/profile.',
-          ErrorCode.MissingFixieAPIKey,
-          'user'
-        );
-      }
+  constructor(
+    private readonly corpusId: string,
+    fixieApiConfiguration: FixieAPIConfiguration = DEFAULT_API_CONFIGURATION
+  ) {
+    this.fixieApiConfiguration = fixieApiConfiguration;
+    if (!this.fixieApiConfiguration.authToken) {
+      throw new AIJSXError(
+        `You must provide a Fixie API key to access Fixie corpora. Find yours at ${this.fixieApiConfiguration.url}/profile.`,
+        ErrorCode.MissingFixieAPIKey,
+        'user'
+      );
     }
-    this.fixieApiUrl = getEnvVar('FIXIE_API_URL', false) ?? FixieCorpus.DEFAULT_FIXIE_API_URL;
   }
 
   async search(query: string, params?: { limit?: number }): Promise<ScoredChunk<ChunkMetadata>[]> {
-    const url = `${this.fixieApiUrl}/api/v1/corpora/${this.corpusId}:query`;
+    const url = new URL(`/api/v1/corpora/${this.corpusId}:query`, this.fixieApiConfiguration.url);
     const response = await fetch(url, {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
-        Authorization: `Bearer ${this.fixieApiKey}`,
+        Authorization: `Bearer ${this.fixieApiConfiguration.authToken}`,
       },
       body: JSON.stringify({
         corpus_id: this.corpusId,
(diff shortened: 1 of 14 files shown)


packages/ai-jsx/src/batteries/sidekick/platform/sidekick.tsx
```diff
@@ -52,31 +51,12 @@ export interface SidekickProps {
   role: string;
 }
 
-function makeObservedTools(tools: UseToolsProps['tools'], logger: AI.ComponentContext['logger']) {
```
farzadab, line 55: So the logging is not needed anymore?

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -84,15 +84,7 @@ export async function getNextConversationStep(
...
-      try {
-        return <FunctionResponse name={name}>{await tools[name].func(args)}</FunctionResponse>;
-      } catch (e: any) {
-        return (
-          <FunctionResponse failed name={name}>
-            {e.message}
-          </FunctionResponse>
-        );
-      }
+      return <ExecuteFunction func={tools[name].func} name={name} args={args} />;
```
farzadab, line 87: I like how you managed to get around the issue of passing `getContext` down to `func`. I'm not sure if I can do the same though. I'll take another stab at it.

farzadab (approved):
I like the way you managed this. I had gotten around this by literally passing the `getContext` down to where it was needed but your solution is more elegant.

I might need to take another pass at the PR, but overall it LGTM.

packages/ai-jsx/src/batteries/docs.tsx
```diff
@@ -663,6 +663,36 @@ export class FixieCorpus<ChunkMetadata extends Jsonifiable = Jsonifiable> implem
...
+         * Reverse the array so the closest chunk is listed last.
+         *
+         * N.B.: The results will not be sorted by `score` because the Fixie API reranks the chunks.
+         */
+        results.reverse();
+
+        return JSON.stringify({
+          kind: 'docs',
+          results: _.uniqBy(results, (result) => result.chunk.content),
+        });
```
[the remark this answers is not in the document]
petersalas: Yeah it's not great. :(

I think Eventually But Hopefully Soon (TM) we will replace partial rendering with the ability for components to produce (stringifiable) structure -- and when that happens we can allow components to return proper objects.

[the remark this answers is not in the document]
farzadab: I meant more like taking something similar to `ChunkFormatter` that our corpus uses (and then have a `default` for it).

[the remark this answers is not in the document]
petersalas: Oh I see -- yeah I have no particular opinions on this, was just copying it from here: https://github.com/fixie-ai/sidekick-helpscout/blob/main/src/tools.tsx#L132

packages/ai-jsx/src/batteries/fixie.tsx
```diff
@@ -0,0 +1,14 @@
+import { createContext } from '../core/render.js';
+import { getEnvVar } from '../lib/util.js';
+
+export interface FixieAPIConfiguration {
+  url: string;
+  authToken?: string;
+}
+
+export const DEFAULT_API_CONFIGURATION: FixieAPIConfiguration = {
+  url: getEnvVar('FIXIE_API_URL', false) ?? 'https://app.fixie.ai',
```
NickHeiner, line 10:
```suggestion
  url: getEnvVar('FIXIE_API_URL', false) ?? 'https://beta.fixie.ai',
```

NickHeiner: Or `console` or w/e

packages/ai-jsx/src/batteries/sidekick/platform/sidekick.tsx
NickHeiner: heheheh

packages/ai-jsx/src/batteries/use-tools.tsx
```diff
@@ -23,12 +23,10 @@ export interface Tool {
   parameters: FunctionParameters;
 
   /**
-   * A function to invoke the tool.
+   * A function/component that invokes the tool.
```
NickHeiner, line 26: Can we provide guidance on when one would want to provide a function vs. a component?

packages/examples/test/batteries/use-tools.tsx
```diff
@@ -0,0 +1,124 @@
...
+    return <AssistantMessage>DONE</AssistantMessage>;
+  }
+  return <FunctionCall name="myFunc" args={{ parameter: 'test parameter value' }} />;
+}
+
+describe('useTools', () => {
+  it('should give tools access to context', async () => {
+    const myContext = AI.createContext(0);
+
+    function MyTool({ parameter }: { parameter: string }, { getContext }: AI.ComponentContext) {
```
NickHeiner, line 24: I think it would be nice to have an example where the tool doesn't use JSX and is just a vanilla function.

packages/examples/test/batteries/use-tools.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+async function FakeChatCompletion({ children }: { children: AI.Node }, { render }: AI.ComponentContext) {
+  const nodes = await renderToConversation(children, render);
+  if (nodes.find((n) => n.type === 'functionCall')) {
+    return <AssistantMessage>DONE</AssistantMessage>;
+  }
+  return <FunctionCall name="myFunc" args={{ parameter: 'test parameter value' }} />;
+}
+
+describe('useTools', () => {
```
NickHeiner, line 20: uber-nit: I wouldn't bother with a top-level `describe` that wraps the entire file – it's not adding info beyond what's already in the file path (which Jest already outputs).

NickHeiner (approved):
I'm excited to see the result of merging this.

farzadab:
@petersalas not urgent, but just letting you know I'm waiting on this PR to merge before updating [my token limiting PR](https://github.com/fixie-ai/ai-jsx/pull/266) to remove my hacks around sending context to the function.

packages/ai-jsx/src/batteries/use-tools.tsx
petersalas: I mean...technically, they're always components, and components just happen to be functions. :)

I added a bit more exposition here but skirted your question because I wasn't sure how to answer it independently of "what are AI.JSX components", and that seemed quite a bit broader than what would be appropriate in this doc comment :P

PR #271, opened 2023-09-11
Add `fixie serve` to the Node.js CLI.

mdwelsh:
This seems to be working now.

### DIFF
--- a/packages/fixie-hello-world/src/index.tsx
+++ b/packages/fixie-hello-world/src/index.tsx
@@ -1,5 +1,7 @@
 import { ChatCompletion, SystemMessage, ConversationHistory } from 'ai-jsx/core/completion';
 
+console.log('Fixie Hello World agent starting.');
+
 export default function HelloWorld() {
   return (
     <ChatCompletion temperature={1}>
--- a/packages/fixie/package.json
+++ b/packages/fixie/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fixieai/fixie",
-  "version": "1.0.14",
+  "version": "1.0.15",
   "license": "MIT",
   "repository": "fixie-ai/ai-jsx",
   "bugs": "https://github.com/fixie-ai/ai-jsx/issues",
--- a/packages/fixie/package.json
+++ b/packages/fixie/package.json
@@ -25,6 +25,7 @@
     "@fixieai/sdk": "*",
     "apollo-upload-client": "^17.0.0",
     "commander": "^11.0.0",
+    "execa": "^8.0.1",
     "extract-files": "^13.0.0",
     "graphql": "^16.8.0",
     "js-yaml": "^4.1.0",
--- a/packages/fixie/package.json
+++ b/packages/fixie/package.json
@@ -22,6 +22,7 @@
   "bin": "./dist/main.js",
   "dependencies": {
     "@apollo/client": "^3.8.1",
+    "@fixieai/sdk": "*",
     "apollo-upload-client": "^17.0.0",
     "commander": "^11.0.0",
     "extract-files": "^13.0.0",
--- a/packages/fixie/package.json
+++ b/packages/fixie/package.json
@@ -22,7 +22,6 @@
   "bin": "./dist/main.js",
   "dependencies": {
     "@apollo/client": "^3.8.1",
-    "@fixieai/sdk": "*",
     "apollo-upload-client": "^17.0.0",
     "commander": "^11.0.0",
     "execa": "^8.0.1",
@@ -34,6 +33,7 @@
     "terminal-kit": "^3.0.0"
   },
   "devDependencies": {
+    "@fixieai/sdk": "*",
     "@tsconfig/node18": "^2.0.1",
     "@types/apollo-upload-client": "^17.0.2",
     "@types/extract-files": "^8.1.1",
--- a/packages/fixie/src/agent.ts
+++ b/packages/fixie/src/agent.ts
@@ -324,7 +324,7 @@ export class FixieAgent {
     let agent: FixieAgent;
     try {
       agent = await FixieAgent.GetAgent(client, agentId);
-      term('👽 Updating agent ').green(agentId)('...\n');
+      term('🦊 Updating agent ').green(agentId)('...\n');
       agent.update({
         name: config.name,
         description: config.description,
@@ -333,7 +333,7 @@ export class FixieAgent {
       });
     } catch (e) {
       // Try to create the agent instead.
-      term('🌲 Creating new agent ').green(agentId)('...\n');
+      term('🦊 Creating new agent ').green(agentId)('...\n');
       agent = await FixieAgent.CreateAgent(
         client,
         config.handle,
@@ -347,9 +347,9 @@ export class FixieAgent {
   }
 
   static spawnAgentProcess(agentPath: string, port: number): ChildProcess {
-    const spinner = ora(' 🌱 Starting local agent process...').start();
-    if (!fs.existsSync(`${agentPath}/dist/serve-bin.js`)) {
-      throw Error(`No dist/serve-bin.js found in ${agentPath}. Do you need to build your agent code first?`);
+    term(`🌱 Starting local agent process on port ${port}...\n`);
+    if (!fs.existsSync(`${agentPath}/dist/index.js`)) {
+      throw Error(`No dist/index.js found in ${agentPath}. Do you need to build your agent code first?`);
     }
 
     // Ensure we kill child process on exit.
@@ -362,7 +362,7 @@ export class FixieAgent {
 
     const cmdline = `npx --package=@fixieai/sdk fixie-serve-bin --packagePath ./dist/index.js --port ${port}`;
     // Split cmdline into the first value (argv0) and a list of arguments separated by spaces.
-    term('\n🌱 Running: ').green(cmdline)('\n');
+    term('🌱 Running: ').green(cmdline)('\n');
     const [argv0, ...args] = cmdline.split(' ');
     const subProcess = spawn(argv0, args, { cwd: agentPath, signal });
     subProcess.stdout.on('data', (sdata: string) => {
@@ -374,7 +374,6 @@ export class FixieAgent {
     subProcess.on('close', (returnCode: number) => {
       term(`🌱 Agent child process exited with code ${returnCode} - `).red('exiting.\n');
     });
-    spinner.succeed(`Local agent process started on port ${port}`);
     return subProcess;
   }
 
@@ -541,7 +540,7 @@ export class FixieAgent {
     const agent = await this.ensureAgent(client, agentId, config);
 
     for await (const currentUrl of deployment_urls_iter) {
-      term('👨‍🍳  Serving agent at ').green(currentUrl)('\n');
+      term('🥡 Serving agent at ').green(currentUrl)('\n');
       try {
         // Wait 3 seconds to ensure the tunnel is set up.
         await new Promise((resolve) => setTimeout(resolve, 3000));
--- a/packages/fixie/src/agent.ts
+++ b/packages/fixie/src/agent.ts
@@ -255,6 +255,8 @@ export class FixieAgent {
   }): Promise<string> {
     const uploadFile = tarball ? fs.readFileSync(fs.realpathSync(tarball)) : undefined;
 
+    console.log(`Creating agent revision at external URL ${externalUrl}`);
+
     const result = await this.client.gqlClient().mutate({
       mutation: gql`
         mutation CreateAgentRevision(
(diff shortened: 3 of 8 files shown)


packages/fixie/package.json
```diff
@@ -22,6 +22,7 @@
   "bin": "./dist/main.js",
   "dependencies": {
     "@apollo/client": "^3.8.1",
+    "@fixieai/sdk": "*",
```
petersalas, line 25: Can this dependency be delegated to the agent project? I'm starting to think it needs to be a dependency of the agent project anyway for the peerDependency resolution to work properly with ai-jsx.

mdwelsh: I don't know. I was unable to get this to work without putting this in here. It might have to do with the fact that I'm running it locally using yarn.

packages/fixie/src/agent.ts
```diff
@@ -317,30 +343,214 @@ export class FixieAgent {
...
       );
     }
+    return agent;
+  }
+
+  static spawnAgentProcess(agentPath: string, port: number): ChildProcess {
+    term(`🌱 Starting local agent process on port ${port}...\n`);
+    if (!fs.existsSync(`${agentPath}/dist/index.js`)) {
+      throw Error(`No dist/index.js found in ${agentPath}. Do you need to build your agent code first?`);
+    }
```
NickHeiner, line 353:
```suggestion
    const pathToCheck = path.resolve(path.join(agentPath, 'dist', 'index.js`));
    if (!fs.existsSync(pathToCheck)) {
      throw Error(`Your agent was not found at ${pathToCheck}. Do you need to build your agent code first?`);
    }
```

We need to use `path.join` instead of hardcoding `/` if we want Windows compat.

packages/fixie-hello-world/src/index.tsx
```diff
@@ -1,5 +1,7 @@
 import { ChatCompletion, SystemMessage, ConversationHistory } from 'ai-jsx/core/completion';
 
+console.log('Fixie Hello World agent starting.');
```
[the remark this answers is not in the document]
mdwelsh: Well, the idea was to show how one can do debugging in their own agent code.

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+      throw Error(`Your agent was not found at ${pathToCheck}. Do you need to build your agent code first?`);
+    }
+
+    // Ensure we kill child process on exit.
+    const controller = new AbortController();
+    const { signal } = controller;
+    process.on('exit', () => {
+      console.log('🌱 Terminating agent process');
+      controller.abort();
+    });
```
NickHeiner, line 364: If we use https://www.npmjs.com/package/execa, this is handled for us automatically.

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+    const [argv0, ...args] = cmdline.split(' ');
+    const subProcess = spawn(argv0, args, { cwd: agentPath, signal });
+    subProcess.stdout.on('data', (sdata: string) => {
+      console.log(`🌱 Agent stdout: ${sdata}`);
+    });
+    subProcess.stderr.on('data', (sdata: string) => {
+      console.error(`🌱 Agent stdout: ${sdata}`);
+    });
+    subProcess.on('close', (returnCode: number) => {
+      term(`🌱 Agent child process exited with code ${returnCode} - `).red('exiting.\n');
```
NickHeiner, line 378: It seems like this is always a problem, right? So maybe we should have more user-facing error messaging here?

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+    agentPath: string,
+    environmentVariables: Record<string, string> = {}
+  ): Promise<string> {
+    const config = await FixieAgent.LoadConfig(agentPath);
+    const agentId = `${(await client.userInfo()).username}/${config.handle}`;
+    term('🦊 Deploying agent ').green(agentId)('...\n');
+
+    // Check that the package.json path exists in this directory.
+    if (!fs.existsSync(path.resolve(path.join(agentPath, 'package.json')))) {
+      throw Error(`No package.json found in ${agentPath}. Only JS-based agents are supported.`);
```
NickHeiner, line 395: minor: I think it's always better to provide the full path we checked.

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+    port: number;
+    environmentVariables: Record<string, string>;
+    debug?: boolean;
+  }) {
+    const config = await FixieAgent.LoadConfig(agentPath);
+    const agentId = `${(await client.userInfo()).username}/${config.handle}`;
+    term('🦊 Serving agent ').green(agentId)('...\n');
+
+    // Check if the package.json path exists in this directory.
+    if (!fs.existsSync(`${agentPath}/package.json`)) {
```
NickHeiner, line 426: Use `path.join` for cross-platform compat

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+      throw Error(`No package.json found in ${agentPath}. Only JS-based agents are supported.`);
+    }
+
+    // Start the agent process locally.
+    const agentProcess = FixieAgent.spawnAgentProcess(agentPath, port);
+
+    // Wait for 5 seconds for it to start up.
+    await new Promise((resolve) => setTimeout(resolve, 5000));
```
NickHeiner, line 434: This can be a follow-up, but it's much better to do a spinning assert – sleeps are very brittle.

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+          '--output=json',
+        ],
+        { signal }
+      );
+      subProcess.stdout.setEncoding('utf8');
+
+      // Every time the subprocess emits a new line, we parse it as JSON ans
+      // extract the 'address' field.
+      let currentLine = '';
+      subProcess.stdout.on('data', (chunk: any) => {
```
NickHeiner, line 480:
```suggestion
      subProcess.stdout.on('data', (chunk: string) => {
```

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+        // We need to do buffering since the data we get from stdout
+        // will not necessarily be line-buffered. We can get 0, 1, or more complete
+        // lines in a single chunk.
+        currentLine += chunk;
+        let newlineIndex;
+        while ((newlineIndex = currentLine.indexOf('\n')) !== -1) {
+          const line = currentLine.slice(0, newlineIndex);
+          currentLine = currentLine.slice(newlineIndex + 1);
+          // Parse sdata as JSON.
```
NickHeiner, line 489:
```suggestion
          // Parse data as JSON.
```

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+          const pdata = JSON.parse(line);
+          // If pdata has the 'address' field, yield it.
+          if (pdata.address) {
+            // We can't yield until the parent has set up the yieldValue callback.
+            if (yieldValue) {
+              yieldValue(`https://${pdata.address}`);
```
NickHeiner, line 495:
```suggestion
              yieldValue?.(`https://${pdata.address}`);
```

packages/fixie/src/main.ts
```diff
@@ -48,6 +48,38 @@ function registerDeployCommand(command: Command) {
     });
 }
 
+/** Run an agent locally. */
+function registerServeCommand(command: Command) {
+  command
+    .command('serve [path]')
+    .description('Run an agent locally')
```
NickHeiner, line 55: Once the agent is running, where do we want to direct people to interact with it? (e.g. "am I supposed to CURL this thing?") I think we want to tell people to go to beta.fixie.ai?

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+          yield config.deploymentUrl;
+        })(),
+      ];
+    }
+
+    // TODO(mdw): We need to restore the original agent deployment (if one existed)
+    // after the process finishes.
+    const agent = await this.ensureAgent(client, agentId, config);
+
+    for await (const currentUrl of deployment_urls_iter) {
```
NickHeiner, line 543: What's the scenario under which there are multiple deployment URLs?

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+        whenYielded = new Promise<string>((resolve) => {
+          yieldValue = resolve;
+        });
+      }
+    }
 
-    return result.data.agentById.currentRevision as AgentRevision;
+    // This is an iterator which yields the public URL of the tunnel where the agent
+    // can be reached by the Fixie service. The tunnel address can change over time.
+    let deployment_urls_iter;
```
NickHeiner, line 525:
```suggestion
    let deploymentUrlsIter;
```

this ain't python y'all

packages/fixie/src/agent.ts
```diff
@@ -317,30 +345,213 @@ export class FixieAgent {
...
+      throw Error(`No package.json found in ${agentPath}. Only JS-based agents are supported.`);
+    }
+
+    // Start the agent process locally.
+    const agentProcess = FixieAgent.spawnAgentProcess(agentPath, port);
+
+    // Wait for 5 seconds for it to start up.
+    await new Promise((resolve) => setTimeout(resolve, 5000));
+
+    async function* spawnTunnel(port: number): AsyncGenerator<string> {
```
NickHeiner, line 436: The promise/iterator/event emitter combo is really awkward here, but I don't understand the overall intent well enough to suggest an alternative. I think https://github.com/fixie-ai/ai-jsx/pull/271/files#r1321994681 will help me get a better handle on it.

Or we could say "we're a startup and we write scrappy code" and just yolo this if it works.

packages/fixie/src/agent.ts
mdwelsh: As you can tell, I Have No Idea What I'm Doing[tm], so let me know what you mean by this :-)

packages/fixie/src/agent.ts
mdwelsh: Oh boy new and exciting syntax

packages/fixie/src/main.ts
mdwelsh: This happens already:

```
🥡 Revision 5467ec8e was deployed to https://beta.fixie.ai/agents/mdw/sidekick-github
```

packages/fixie/src/agent.ts
```diff
@@ -317,30 +344,196 @@ export class FixieAgent {
...
+    environmentVariables: Record<string, string> = {}
+  ): Promise<string> {
+    const config = await FixieAgent.LoadConfig(agentPath);
+    const agentId = `${(await client.userInfo()).username}/${config.handle}`;
+    term('🦊 Deploying agent ').green(agentId)('...\n');
+
+    // Check that the package.json path exists in this directory.
+    const packageJsonPath = path.resolve(path.join(agentPath, 'package.json'));
+    if (!fs.existsSync(packageJsonPath)) {
+      throw Error(`No package.json found in ${packageJsonPath}. Only JS-based agents are supported.`);
```
NickHeiner, line 390:
```suggestion
      throw Error(`No package.json found at ${packageJsonPath}. Only JS-based agents are supported.`);
```

packages/fixie/package.json
mdwelsh: Yeah, that works. Done.

packages/fixie/src/agent.ts
NickHeiner: I mean doing something like:

```ts
async function waitForServerStart() {
    return new Promise(resolve => {
        while (timeoutNotYetExceeded) {
            if (serverIsUp()) { resolve(); return; }
        }
        throw new Error(`Server didn't start within ${timeoutMs} milliseconds. You could try /* ... */`)
    });
}
```

mdwelsh: I agree we should do something smarter here, but do you mind if we do it as a follow up? I would need to write more code to figure out if the server is up, etc.

packages/fixie/src/agent.ts
NickHeiner: Ok. I think this should just be an event emitter, and dispense with the promise/iterator. But it's probably not the best use of your time to do that refactor. :smile:

NickHeiner (approved):

PR #273, opened 2023-09-12
Update CLI for source addition:

mdepinet:
Allow multiple urls
Allow setting max-documents and max-depth
Rely on server for default values

### DIFF
--- a/packages/fixie/src/main.ts
+++ b/packages/fixie/src/main.ts
@@ -145,18 +145,18 @@ corpus
 const sources = corpus.command('sources').description('Corpus source related commands');
 
 sources
-  .command('add <corpusId> <urls...>')
+  .command('add <corpusId> <startUrls...>')
   .description('Add a source to a corpus.')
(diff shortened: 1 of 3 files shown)


packages/fixie/src/main.ts
```diff
@@ -145,13 +145,21 @@ corpus
 const sources = corpus.command('sources').description('Corpus source related commands');
 
 sources
-  .command('add <corpusId> <urlPattern>')
+  .command('add <corpusId> <urls...>')
```
mdwelsh, line 148: Should this be 'urlPatterns' or 'startUrls' or something?

mdwelsh (approved):

packages/fixie/src/main.ts
mdepinet: Done.

mdepinet:
Will submit once https://github.com/fixie-ai/fixie/pull/1639 is in

PR #274, opened 2023-09-12
Some quality-of-life improvements for the CLI.

mdwelsh:
This PR has a few little things:
* Ensures that new Agents are always created in a "published" state, so users won't inadvertently try to deploy or serve an agent that the frontend can't talk to.
* Warns if the agent.yaml file has any unknown keys.
* Makes the message on successful serve and deploy more clear.


NickHeiner (approved):
LGTM on a quick scan

PR #278, opened 2023-09-12
Minor `fixie serve`/`fixie deploy` UX improments

petersalas:
- Pass stdio through from `npm pack` to surface build errors
- Run `npm pack` prior to `fixie serve` for consistency with `fixie deploy`
- Use environment variables with `fixie serve` for consistency with `fixie deploy`
- Add `yarn build` as a `prepack` script for the Fixie CLI itself


mdwelsh (approved):

packages/fixie/src/agent.ts
```diff
@@ -260,28 +260,31 @@ export class FixieAgent {
...
 
     // Create a temporary directory and run `npm pack` inside.
     const tempdir = fs.mkdtempSync(path.join(os.tmpdir(), `fixie-tmp-${packageJson.name}-${packageJson.version}-`));
-    const commandline = `npm pack ${path.resolve(agentPath)} --silent >/dev/null`;
-    execSync(commandline, { cwd: tempdir });
+    const commandline = `npm pack ${path.resolve(agentPath)}`;
+    try {
+      execSync(commandline, { cwd: tempdir, stdio: 'inherit' });
+    } catch (ex) {
+      throw new Error(`\`${commandline}\` failed. Check for build errors above and retry.`);
```
NickHeiner, line 272: Logging `ex` would be redundant?

packages/fixie/src/agent.ts
```diff
@@ -260,28 +260,31 @@ export class FixieAgent {
...
-    tarball?: string;
-    environmentVariables: Record<string, string>;
-  }): Promise<AgentRevision> {
-    const uploadFile = tarball ? fs.readFileSync(fs.realpathSync(tarball)) : undefined;
+  private async createRevision(
+    opts:
+      | { externalUrl: string; tarball?: undefined }
+      | {
+          externalUrl?: undefined;
+          tarball: string;
```
NickHeiner, line 283: It looks like you want `MergeExclusive` here from type-fest.

packages/fixie/src/agent.ts
```diff
@@ -475,8 +481,28 @@ export class FixieAgent {
...
+    const packageJson = fs.readFileSync(packageJsonPath, 'utf8');
+    let json;
+    try {
+      json = JSON.parse(packageJson);
+    } catch (ex) {
+      throw Error(`Error parsing package.json: ${ex}`);
+    }
+
+    if (json.scripts) {
+      const scriptName = json.scripts.prepack ? 'prepack' : json.scripts.build ? 'build' : undefined;
```
NickHeiner, line 493: To be fully pedantic, I think you may need to check for a few more keys: https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-pack.

Ultimately, the safest thing to do is just to run `npm publish` and let npm invoke the right thing. But I recognize that makes it harder to do what we're trying to do here of presenting a nicer DX, so maybe we can't do that.

NickHeiner (approved):
LGTM on very quick glance

packages/fixie/src/agent.ts
petersalas: Nice!

packages/fixie/src/agent.ts
petersalas: I buy it -- and it's also simpler to just run the same `npm pack` function we use for `fixie deploy`.

PR #280, opened 2023-09-13
Add `fixie auth` command to CLI.

mdwelsh:
This PR enables the OAuth-based browser login flow for users the first time they run the Fixie CLI. In fact, this will be the default for first-run, so there is no need to explicitly run `fixie auth` first. But, running `fixie auth` will show your current login credentials.

Like the Python CLI, it saves the credentials to `~/.config/fixie/config.yaml`.


packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+import { FixieClient } from './client.js';
+
+/** Represents contents of the Fixie CLI config file. */
+export interface FixieConfig {
+  apiUrl?: string;
+  apiKey?: string;
+}
+
+export const FIXIE_API_URL = 'https://api.fixie.ai';
+export const FIXIE_CONFIG_FILE = '~/.config/fixie/config.yaml';
```
NickHeiner, line 25: Plz use `os.homedir()` and `path.join` for cross-platform compat https://nodejs.org/api/os.html#oshomedir.

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/** Load the client configuration from the given file. */
+export function LoadConfig(configFile: string): FixieConfig {
```
NickHeiner, line 28:
```suggestion
export function loadConfig(configFile: string): FixieConfig {
```

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+  const fullPath = untildify(configFile);
+  if (!fs.existsSync(fullPath)) {
+    return {};
+  }
+  const config: object = yaml.load(fs.readFileSync(fullPath, 'utf8')) as object;
```
NickHeiner, line 33:
```suggestion
  const config = yaml.load(fs.readFileSync(fullPath, 'utf8')) as object;
```

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+  const validKeys = ['apiUrl', 'apiKey'];
+  const invalidKeys = Object.keys(config).filter((key) => !validKeys.includes(key));
+  for (const key of invalidKeys) {
+    term('❓ Ignoring invalid key ').yellow(key)(` in ${fullPath}\n`);
+  }
+  return config as FixieConfig;
+}
+
+/** Save the client configuration to the given file. */
+export function SaveConfig(config: FixieConfig, configFile: string) {
```
NickHeiner, line 44:
```suggestion
export function saveConfig(config: FixieConfig, configFile: string) {
```

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+  const fullPath = untildify(configFile);
+  const dirName = path.dirname(fullPath);
+  if (!fs.existsSync(dirName)) {
+    fs.mkdirSync(dirName, { recursive: true });
+  }
+  if (fs.existsSync(fullPath)) {
+    // Merge the new config with the existing config, so we don't
+    // overwrite any fields that are not specified.
+    const currentConfig = yaml.load(fs.readFileSync(fullPath, 'utf8')) as object;
```
NickHeiner, line 53: Wouldn't it be better to reuse `loadConfig` for this?

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  const apiKey = await oauthFlow(apiUrl ?? FIXIE_API_URL);
+  const config: FixieConfig = {
+    apiUrl: apiUrl ?? FIXIE_API_URL,
+    apiKey,
+  };
+  SaveConfig(config, configFile ?? FIXIE_CONFIG_FILE);
+  const client = await Authenticate({ apiUrl, configFile: configFile ?? FIXIE_CONFIG_FILE });
+  if (!client) {
+    throw new Error('Failed to authenticate');
```
NickHeiner, line 120: Can we provide a more actionable error message? Should the dev retry? Ask for help in Discord? Is it possibly their fault?

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+          const code = searchParams.get('code');
+          const receivedState = searchParams.get('state');
+          if (code && receivedState === state) {
+            try {
+              const bodyFormData = new FormData();
+              bodyFormData.append('code', code);
+              bodyFormData.append('redirect_uri', redirectUri);
+              bodyFormData.append('client_id', CLIENT_ID);
+              bodyFormData.append('grant_type', 'authorization_code');
+              const response = await axios.post(`${apiUrl}/access/token`, bodyFormData, {
```
NickHeiner, line 160: You can probably use native fetch and don't need axios but not worth changing now

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+      .listen(port);
+  });
+
+  await open(url);
+  term('🔑 Your browser has been opened to visit:\n\n   ').blue.underline(url)('\n\n');
+  return serverPromise as Promise<string>;
+}
+
+/** Return a free port on the local machine. */
+function findFreePort(): Promise<number> {
```
NickHeiner, line 195: We could use https://www.npmjs.com/package/get-port instead.

packages/fixie/src/auth.ts
```diff
@@ -0,0 +1,207 @@
...
+  const client = FixieClient.Create(useApiUrl, useApiKey);
+  const userInfo = await client.userInfo();
+  if (userInfo.is_anonymous) {
+    return null;
+  }
+  return client;
+}
+
+/** Returns an authenticated FixieClient, starting an OAuth flow to authenticate the user if necessary. */
+export async function AuthenticateOrLogin({
```
NickHeiner, line 90:
```suggestion
export async function AuthenticateOrLogIn({
```

VERY IMPORTANT TOTAL CR BLOCKER

"log in" is a verb; "login" is a noun.

NickHeiner (approved):
LGTM on very quick glance

packages/fixie/src/auth.ts
mdwelsh: untildify handles this for us!

packages/fixie/src/auth.ts
mdwelsh: Well, I want to retain all the keys in the original config file and loadConfig is intended to only return valid FixieConfig keys.

petersalas:
Given Nick's sign off, don't block on me

PR #289, opened 2023-09-13
Allow --include-patterns and --exclude-patterns to be specified on the command line.

mdwelsh:
With #1683 landing on fixie-ai/fixie, it is now required that the client specify include_glob_patterns and exclude_glob_patterns when creating a source. This PR adds this as a flag to the CLI.

@NickHeiner You will want to make a similar change to the UI I expect.

### DIFF
--- a/packages/fixie/src/isomorphic-client.ts
+++ b/packages/fixie/src/isomorphic-client.ts
@@ -125,6 +125,8 @@ export class IsomorphicFixieClient {
   addCorpusSource(
     corpusId: string,
     startUrls: string[],
+    includeGlobs?: string[],
+    excludeGlobs?: string[],
     maxDocuments?: number,
     maxDepth?: number
   ): Promise<Jsonifiable> {
@@ -137,11 +139,12 @@ export class IsomorphicFixieClient {
           web: {
             start_urls: startUrls,
             max_depth: maxDepth,
+            include_glob_patterns: includeGlobs,
+            exclude_glob_patterns: excludeGlobs,
           },
         },
       },
     };
-
     return this.request(`/api/v1/corpora/${corpusId}/sources`, body);
   }
 
--- a/packages/fixie/src/main.ts
+++ b/packages/fixie/src/main.ts
@@ -179,14 +179,36 @@ sources
   .description('Add a source to a corpus.')
   .option('--max-documents <number>', 'Maximum number of documents to crawl')
   .option('--max-depth <number>', 'Maximum depth to crawl')
+  .option('--include-patterns <pattern...>', 'URL patterns to include in the crawl')
+  .option('--exclude-patterns <pattern...>', 'URL patterns to exclude from the crawl')
   .action(
     async (
       corpusId: string,
       startUrls: string[],
-      { maxDocuments, maxDepth }: { maxDocuments?: number; maxDepth?: number }
+      {
+        maxDocuments,
+        maxDepth,
+        includePatterns,
+        excludePatterns,
+      }: { maxDocuments?: number; maxDepth?: number; includePatterns?: string[]; excludePatterns: string[] }
     ) => {
+      if (!includePatterns) {
+        term.yellow('Warning: ')(
+          'No --include-patterns specfied. This is equivalent to only crawling the URLs specified as startUrls.\n'
+        );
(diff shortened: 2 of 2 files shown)


packages/fixie/src/main.ts
```diff
@@ -179,14 +179,36 @@ sources
...
+        maxDepth,
+        includePatterns,
+        excludePatterns,
+      }: { maxDocuments?: number; maxDepth?: number; includePatterns?: string[]; excludePatterns: string[] }
     ) => {
+      if (!includePatterns) {
+        term.yellow('Warning: ')(
+          'No --include-patterns specfied. This is equivalent to only crawling the URLs specified as startUrls.\n'
+        );
+        term.yellow('Warning: ')('Use ').red("--include-patterns '*'")(
```
petersalas, line 199: Is there any scenario where this is actually a reasonable thing to do?

packages/fixie/src/main.ts
```diff
@@ -179,14 +179,36 @@ sources
   .description('Add a source to a corpus.')
   .option('--max-documents <number>', 'Maximum number of documents to crawl')
   .option('--max-depth <number>', 'Maximum depth to crawl')
+  .option('--include-patterns <pattern...>', 'URL patterns to include in the crawl')
+  .option('--exclude-patterns <pattern...>', 'URL patterns to exclude from the crawl')
```
mdepinet, line 183: OOC does this work if you have multiple include patterns and also multiple exclude patterns?

packages/fixie/src/main.ts
```diff
@@ -179,14 +179,36 @@ sources
...
     async (
       corpusId: string,
       startUrls: string[],
-      { maxDocuments, maxDepth }: { maxDocuments?: number; maxDepth?: number }
+      {
+        maxDocuments,
+        maxDepth,
+        includePatterns,
+        excludePatterns,
+      }: { maxDocuments?: number; maxDepth?: number; includePatterns?: string[]; excludePatterns: string[] }
```
mdepinet, line 193:
```suggestion
      }: { maxDocuments?: number; maxDepth?: number; includePatterns?: string[]; excludePatterns?: string[] }
```

packages/fixie/src/main.ts
mdepinet, line 199: This is incorrect.  It should be **.  Apify has a "test glob" button you can use to experiment.

mdepinet (approved):

packages/fixie/src/main.ts
mdwelsh: Sure, if you want to rely on max-depth and max-documents and crawl anything you find from a start URL. If you have a better proposed default, I'm all for it, but I don't know what it would be :-)

packages/fixie/src/main.ts
mdwelsh: Yes

packages/fixie/src/main.ts
mdwelsh: Ah, thanks.

petersalas: My concern is just if the suggestion is overwhelmingly likely to lead to a slow crawl that then fails out...it's worse than not suggesting anything, or simply letting the user consult docs to understand how to use `--include-patterns`.

But if hitting the max is "fine" (i.e. neither users nor Fixie is unhappy) then that's not a concern.

petersalas (approved):

NickHeiner (approved):

packages/fixie/src/isomorphic-client.ts
```diff
@@ -125,6 +125,8 @@ export class IsomorphicFixieClient {
   addCorpusSource(
     corpusId: string,
     startUrls: string[],
+    includeGlobs?: string[],
+    excludeGlobs?: string[],
```
NickHeiner, line 129: This was a breaking change, and should have been accompanied by a bump to `2.0.0`.

mdwelsh: Oops, sorry. What about all of the other unintentional breaking changes? :-)

NickHeiner: 😅

PR #290, opened 2023-09-14
Use the singular consistently

juberti:
If we have `fixie corpus list`, we should also support `fixie agent list` rather than `fixie agents list`


mdwelsh (approved):
I would prefer if you could use `.alias()` so one can use either. Why can't we have both?

mdwelsh:
You need to bump version in package.json also

juberti:
PTAL

PR #291, opened 2023-09-14
Try getting npm publish to work with GitHub Actions.

mdwelsh:


mdwelsh:
I'm going to YOLO this because, well, it's the only way to see if it works.

PR #294, opened 2023-09-14
Add auto-reload support to `fixie serve`.

mdwelsh:
Running `fixie serve` will now watch for changes in the agent directory and restart the agent process locally when this happens.


packages/fixie/src/agent.ts
```diff
@@ -480,7 +488,19 @@ export class FixieAgent {
...
+    // Watch files in the agent directory for changes.
+    const watchPath = path.resolve(agentPath);
+    console.log(`🌱 Watching ${watchPath} for changes...`);
+    const watcher = new Watcher(watchPath, { ignoreInitial: true, recursive: true });
+    watcher.on('all', async (event: any, targetPath: string, _targetPathNext: any) => {
+      console.log(`🌱 Restarting local agent process due to ${event}: ${targetPath}`);
+      agentProcess.kill();
+      // Let it shut down gracefully.
+      await new Promise((resolve) => setTimeout(resolve, 2000));
+      agentProcess = FixieAgent.spawnAgentProcess(agentPath, port, environmentVariables);
```
petersalas, line 502: Since this isn't triggering a new build, how does the code get updated?

packages/fixie/src/agent.ts
```diff
@@ -480,7 +488,19 @@ export class FixieAgent {
...
+
+    // Watch files in the agent directory for changes.
+    const watchPath = path.resolve(agentPath);
+    console.log(`🌱 Watching ${watchPath} for changes...`);
+    const watcher = new Watcher(watchPath, { ignoreInitial: true, recursive: true });
+    watcher.on('all', async (event: any, targetPath: string, _targetPathNext: any) => {
+      console.log(`🌱 Restarting local agent process due to ${event}: ${targetPath}`);
+      agentProcess.kill();
+      // Let it shut down gracefully.
+      await new Promise((resolve) => setTimeout(resolve, 2000));
```
petersalas, line 501: Does this work?

```ts
await new Promise((resolve) => (
  agentProcess.exitCode !== null || agentProcess.signalCode !== null
    ? resolve()
    : agentProcess.on('exit', resolve)
))
```

packages/fixie/src/agent.ts
petersalas: And the next question is -- if we did trigger a new build here, do we need to add something to suppress the event if the existing handler is still going so as not to trigger cascading restarts?

mdwelsh: Somehow I thought the `npx fixie-serve-bin` was triggering a rebuild, but I see that you are correct.
Let me fix this.

packages/fixie/src/agent.ts
NickHeiner, line 501: Hmm, so this means every agent reload is 2 seconds? Web devs complain if their hot reload is longer than 250ms.

When I did this for the v1 SDK, I had the server process do the reload itself, so there was no need to kill/respawn a subproc: https://github.com/fixie-ai/fixie-sdk/blob/feature-ts/fixieai/agents-ts/src/serve.ts#L169-L178. Could that work here as well?

NickHeiner:
If this even sorta works, we should merge it, because not having auto-reload is pretty annoying. :smile:

packages/fixie/src/agent.ts
mdwelsh: Yup!

mdwelsh:
I'm going to interpret Nick's "If this even sorta works, we should merge it, because not having auto-reload is pretty annoying. 😄" as an LGTM and merge this

PR #299, opened 2023-09-15
Add latency display v1 to voice agent page

juberti:
Screenshot 2023-09-15 at 2 00 07 PM


packages/nextjs-demo/src/app/agent/page.tsx
```diff
@@ -226,6 +258,13 @@ const ButtonComponent: React.FC<{ onClick: () => void; disabled: boolean; childr
   </button>
 );
 
+const LatencyComponent: React.FC<{ name: string; latency: number }> = ({ name, latency }) => (
```
NickHeiner, line 261:
```suggestion
const Latency: React.FC<{ name: string; latency: number }> = ({ name, latency }) => (
```

Just like we wouldn't call it `namesArray` or `indexNumber`.

NickHeiner (approved):

packages/nextjs-demo/src/app/agent/page.tsx
juberti: I knew you were going to say this

PR #304, opened 2023-09-16
Add support for Fixie Anthropic proxy

petersalas:
This adds support for Anthropic models within Fixie agents.

### DIFF
--- a/packages/fixie-sdk/src/request-wrapper.tsx
+++ b/packages/fixie-sdk/src/request-wrapper.tsx
@@ -3,7 +3,8 @@ import { InvokeAgentRequest } from './types.js';
 import { ShowConversation, ConversationHistoryContext } from 'ai-jsx/core/conversation';
 import { Json } from './json.js';
 import { FixieConversation } from './conversation.js';
-import { OpenAI, OpenAIClient, ValidChatModel } from 'ai-jsx/lib/openai';
+import { OpenAI, OpenAIClient, ValidChatModel as OpenAIChatModel } from 'ai-jsx/lib/openai';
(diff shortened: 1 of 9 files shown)


packages/fixie-sdk/src/request-wrapper.tsx
```diff
@@ -75,33 +76,56 @@ export function FixieRequestWrapper({ children }: { children: AI.Node }, { getCo
...
-    );
-  } else {
-    throw new Error(`The model provider ("${request.generationParams.modelProvider}") is not supported.`);
+  const modelProvider = request.generationParams?.modelProvider?.toLowerCase() ?? 'openai';
+  if (modelProvider !== 'openai' && modelProvider !== 'anthropic') {
+    throw new Error(`The model provider ("${request.generationParams?.modelProvider}") is not supported.`);
   }
 
-  return (
+  // Set both OpenAI and Anthropic Clients, but only configure the requested one as the default ChatCompletion model.
```
farzadab, line 84: So now if `children` doesn't specify which provider to use, it knows which one to take because one of them doesn't have `chatModel` set?
I'm just worried that the default would become the inner-most one (i.e. Anthropic).

farzadab (approved):
LGTM, just one question.

packages/fixie-sdk/src/request-wrapper.tsx
petersalas: Basically the props are additive:
- `client` -- sets the anthropic/openai client
- `chatModel` -- sets the default chat model
- `completionModel` -- sets the default completion model

The AI.JSX change is that we can now set `client` without also setting `chatModel`

PR #309, opened 2023-09-16
Add GCP TTS support

juberti:
GCP is annoying in that the audio is returned as a JSON object rather than a binary blob.


NickHeiner (approved):
Please do a version bump + changelog update for this since it's introducing new functionality.

PR #313, opened 2023-09-16
Voice agent: clean up menu display

juberti:
remove a few donuts that were making the page too tall
image


packages/nextjs-demo/src/app/agent/page.tsx
```diff
@@ -242,6 +242,13 @@ class ChatManager {
   }
 }
 
+const MenuItem: React.FC<{ name: string; price: number }> = ({ name, price }) => (
```
NickHeiner, line 245: If we cared about accessibility for this demo, you'd want to stick to the semantic markup (`ul` and `li`)s.

NickHeiner (approved):

PR #327, opened 2023-09-20
Add support for clearing and deleting sources to our CLI

mdepinet:
Depends on https://github.com/fixie-ai/fixie/pull/1779

### DIFF
--- a/packages/fixie/src/isomorphic-client.ts
+++ b/packages/fixie/src/isomorphic-client.ts
@@ -54,29 +54,25 @@ export class IsomorphicFixieClient {
   }
 
   /** Send a request to the Fixie API with the appropriate auth headers. */
-  async request(path: string, bodyData?: any): Promise<Jsonifiable> {
-    let res;
+  async request(path: string, bodyData?: any, method?: string): Promise<Jsonifiable> {
+    const fetch_method = method ?? (bodyData ? 'POST' : 'GET');
+    const headers = bodyData
+      ? {
+          Authorization: `Bearer ${this.apiKey}`,
+          'Content-Type': 'application/json',
+        }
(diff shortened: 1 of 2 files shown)


packages/fixie/src/isomorphic-client.ts
```diff
@@ -170,16 +166,35 @@ export class IsomorphicFixieClient {
...
    */
   refreshCorpusSource(corpusId: string, sourceId: string, force?: boolean): Promise<Jsonifiable> {
     return this.request(`/api/v1/corpora/${corpusId}/sources/${sourceId}:refresh`, { force });
   }
 
+  /**
+   * Clear the given Source, removing all its documents and their chunks.
+   *
+   * If a job is already running on this source, and force = false, this call will return an error.
+   * If a job is already running on this source, and force = true, that job will be killed and restarted.
```
NickHeiner, line 192: Thanks for the detailed notes!

packages/fixie/src/main.ts
```diff
@@ -251,19 +251,41 @@ source
...
     const result = await client.refreshCorpusSource(corpusId, sourceId, force);
     showResult(result, program.opts().raw);
   });
 
+source
+  .command('clear <corpusId> <sourceId>')
+  .description('Clear a corpus source.')
+  .option(
+    '--force',
+    "If a job is already running on this source, kill it and start a new one. If a job is already running and you don't pass --force, the command will fail."
```
NickHeiner, line 281:
```suggestion
    "By default, this command will fail if you try to clear a source that current has a job running. If you want to clear the source regardless, pass this flag."
```

NickHeiner (approved):

PR #334, opened 2023-09-21
`@fixieai/sdk`: Add support for request parameters

petersalas:
This adds support for request parameters to be passed to the handler as the first argument (i.e. as props).


NickHeiner (approved):
This seems good. I thought for a moment about whether we could make `Handler` generic and declare the request param shape it expected, but given that it's dynamically imported, I don't think there's any way to do that.

PR #337, opened 2023-09-23
Create more embedding options

NickHeiner:
* Fix a bug with `useSidekick` regarding how `conversationId` was managed
* Clean up the interface for `IsomorphicFixieClient`
* Add embedding react components
* Add a very sparse README
* Fix build setup

~TODO: rename `useSidekick` to `useFixie`~


packages/fixie/readme.md
```diff
@@ -2,6 +2,66 @@
...
+
+This hook provides a fully managed API for a conversation. It returns a number of fields you can use to drive a rich UI, including loading states and debug diagnostics.
+
+#### `IsomorphicFixieClient`
+
+```ts
+import { IsomorphicFixieClient } from 'fixie/web';
+```
+
+This low-level API provides direct access to the Fixie conversation API. You need to manage things like loading state and response parsing on your own, but it's the most flexible.
```
mdwelsh, line 63: "Fixie conversation API" -> "Fixie Conversation and Corpus APIs"

packages/fixie/src/fixie-embed.tsx
```diff
@@ -16,25 +17,138 @@ export interface FixieEmbedProps extends React.IframeHTMLAttributes<HTMLIFrameEl
...
  *
  * Any extra props to this component are passed through to the `iframe`.
  */
-export function FixieEmbed({ speak, debug, agentId, fixieHost, ...iframeProps }: FixieEmbedProps) {
-  const embedUrl = new URL(`/embed/${agentId}`, fixieHost);
+export function InlineFixieEmbed({ speak, debug, agentId, fixieHost, ...iframeProps }: FixieEmbedProps) {
+  return <iframe {...getBaseIframeProps({ speak, debug, agentId, fixieHost })} {...iframeProps}></iframe>;
+}
+
+export function ControlledFloatingFixieEmbed({
```
mdwelsh, line 45: I might recommend docstrings for the various parameters here. I can add them later if needed.

packages/fixie/src/isomorphic-client.ts
```diff
@@ -52,11 +52,13 @@ export class IsomorphicFixieClient {
...
   /**
    * Create a new FixieClient without an API key. This is only useful for accessing public APIs, such as the conversation APIs.
+   *
+   * You only need to pass url if you're pointing to a different Fixie backend than the default production. Unless you specificially know you need to do this, you don't.
    */
   // This is also useful for running in the console.fixie.ai webapp, because it's on the same host
   // as the backend and thus doesn't need the API key, assuming we set the auth cookies to be cross-domain.
-  static CreateWithoutApiKey(url: string) {
-    return new this(url);
+  static CreateWithoutApiKey(url?: string) {
```
mdwelsh, line 60: Why not just "Create" and have the API key parameter be optional?

mdwelsh (approved):
Nice!

packages/fixie/src/fixie-embed.tsx
NickHeiner: All the docstrings are already here. They come from `FixieEmbedProps`. If you hover over the param name in the editor, you can see the intellisense.

image

packages/fixie/src/isomorphic-client.ts
NickHeiner: That would be fine too. When I made this, we had this above:

```ts
  static Create(url: string, apiKey?: string) {
    const apiKeyToUse = apiKey ?? process.env.FIXIE_API_KEY;
    if (!apiKeyToUse) {
      throw new Error(
        'You must pass apiKey to the constructor, or set the FIXIE_API_KEY environment variable. The API key can be found at: https://console.fixie.ai/profile'
      );
    }
    return new this(url, apiKey);
  }
```

Clearly, someone felt strongly that an API key was a good thing. :smile: 

I think creating a client without an API key is a rare enough case that it's worth asking the dev to be more intentional when they do it.

PR #339, opened 2023-09-24
Remove unused field `environment` when doing `fixie deploy`

mdwelsh:
I'm going to YOLO this in unless y'all object


petersalas (approved):
Thank you!

PR #340, opened 2023-09-24
Incorporate `fixie` docs into generated API documentation.

mdwelsh:
This PR incorporates the autogenerated documentation for the `fixie` package in the API Reference docs for ai-jsx.com.

There are several ways of doing this, apparently, in a monorepo like ours. This approach merges all of the API docs together. There are many configuration options we could explore if this doesn't work well, but I think it's good to start with something rather than nothing!

### DIFF
--- a/packages/ai-jsx/tsconfig.json
+++ b/packages/ai-jsx/tsconfig.json
@@ -2,7 +2,8 @@
   "extends": "./tsconfig-base.json",
   "include": ["src/**/*", ".eslintrc.cjs"],
   "compilerOptions": {
-    "outDir": "dist/esm"
+    "outDir": "dist/esm",
+    "rootDir": "./src"
   },
   "typedocOptions": {
     "entryPoints": [
--- a/packages/ai-jsx/readme.md
+++ b/packages/ai-jsx/readme.md
@@ -57,7 +57,7 @@ For a full set of examples, see [the examples package](https://github.com/fixie-
 
 ## Contributing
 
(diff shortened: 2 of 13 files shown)


farzadab:
Nit: it seems like this is mashing up all packages together. This is probably fine. The only worry I have is what would happen if we have similar sounding module names in different packages. This is probably not important at this point and I agree this is better than what we had before.

image

petersalas (commented):
Like Farzad pointed out it seems sort of weird to have modules in different packages listed as peers. If the nav can distinguish between packages that seems better to me but I don't know if that's easy/possible/hard or has other undesirable side effects.

Seems like @benlower should probably take a look as well?

mdwelsh:
Right. I spent a lot of time trying to figure out good ways of doing this, and there are about a million options for configuring things. My proposal was to land this so at least *something* is online and we can iterate on it -- it will take me a while to find time to get back to this, I think.

packages/ai-jsx/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "./tsconfig-base.json",
   "include": ["src/**/*", ".eslintrc.cjs"],
   "compilerOptions": {
-    "outDir": "dist/esm"
+    "outDir": "dist/esm",
+    "rootDir": "./src"
```
NickHeiner, line 6: Just to be extra safe, after we land this PR, let's publish ai-jsx and make sure it's still usable when installed from a third-party library.

NickHeiner (approved):

PR #342, opened 2023-09-25
Use Azure OpenAI service if configured

juberti:


packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -45,17 +45,45 @@ export type ValidChatModel =
...
+  protected override authHeaders() {
+    return {
+      'Api-Key': this.apiKey,
+    };
+  }
+  override buildRequest(options: FinalRequestOptions) {
+    const result = super.buildRequest(options);
+    if (options.body && 'model' in options.body) {
+      const model = (options.body.model as string).replace('.', '');
+      result.url = result.url.replace('/chat/', `/deployments/${model}/chat/`);
```
NickHeiner, line 67: For extra safety, you could use the [url](https://nodejs.org/api/url.html) module to parse out the `pathname` and only run the `replace` on that.

But I guess the `/` characters can only appear in the pathname, so this probably will not have an impact in practice.

NickHeiner (approved):
Neat that the diff to do this was so small!

This will be a minor version bump for the release.

mdwelsh (approved):

packages/ai-jsx/src/lib/openai.tsx
juberti: yeah good call, I found a cleaner approach that isn't isolated to /chat/, so I went with that.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -45,17 +46,44 @@ export type ValidChatModel =
...
+  }
+}
+
 const openAiClientContext = AI.createContext<() => OpenAIClient>(
   _.once(() => {
     const baseURL = getEnvVar('OPENAI_API_BASE', false);
-    return new OpenAIClient({
-      apiKey: getEnvVar('OPENAI_API_KEY', false),
+    const useAzure = baseURL?.endsWith('azure.com');
+    const apiKey = getEnvVar(useAzure ? 'OPENAI_AZURE_API_KEY' : 'OPENAI_API_KEY', false);
```
petersalas, line 77: Should we fall back to `OPENAI_API_KEY` if `useAzure` and `OPENAI_AZURE_API_KEY` isn't set? Or maybe the OpenAI SDK will already do that?

petersalas (approved):

packages/ai-jsx/src/lib/openai.tsx
juberti: I think we should error in that case (which the OAI SDK will do, I think)

PR #347, opened 2023-09-26
Add real voice activity detection (VAD)

juberti:
We'll use this to improve speculative execution (understanding ASAP when voice has ended) and also do TTS interruption.


packages/ai-jsx/src/lib/asr/vad.tsx
```diff
@@ -0,0 +1,117 @@
+'use client';
+import VADBuilder, { VAD, VADMode, VADEvent } from '@ozymandiasthegreat/vad';
+
+/**
+ * Base class for voice activity detectors.
+ */
+export abstract class VoiceActivityDetectorBase {
```
NickHeiner, line 7: If every field is `abstract`, I think you just want an `interface` here instead.

NickHeiner (approved):
LGTM on a very quick glance

packages/ai-jsx/src/lib/asr/vad.tsx
juberti: good point. note that the callbacks are not abstract though.

PR #349, opened 2023-09-27
Don't use workspace dependency in fixie package.

mdwelsh:
This breaks the published `fixie` package which complains about the "workspace:" dependency.

### DIFF
--- a/packages/fixie/package.json
+++ b/packages/fixie/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fixie",
-  "version": "1.5.1",
+  "version": "1.5.2",
   "license": "MIT",
   "repository": "fixie-ai/ai-jsx",
   "bugs": "https://github.com/fixie-ai/ai-jsx/issues",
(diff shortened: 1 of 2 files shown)


packages/fixie/package.json
```diff
@@ -23,7 +23,7 @@
   "dependencies": {
     "@apollo/client": "^3.8.1",
     "@types/apollo-upload-client": "^17.0.2",
-    "ai-jsx": "workspace:*",
+    "ai-jsx": "^0.18.1",
```
petersalas, line 26: @NickHeiner why does this package need to depend on ai-jsx at all?

petersalas: Apparently the answer is for the "PropsOfComponent" type -- I suspect it's not worth bloating the CLI/frontend with the (quite large) dependency graph of ai-jsx, but that's not new with this change :)

petersalas (approved):

packages/fixie/package.json
NickHeiner: Yes, it's just for these two imports:

```ts
import { FunctionResponse as FunctionResponseNode, FunctionCall as FunctionCallNode } from 'ai-jsx/core/conversation';
import { PropsOfComponent } from 'ai-jsx';
```

I'm happy to say this isn't worth it and we should just manually copy. It shouldn't be bloating the frontend bundle because it's just types.

PR #350, opened 2023-09-27
Add support for listing corpora by owner type, and better error handling.

mdwelsh:
This allows us to select which type of corpora to list, which means I can fix the UI to do things like selectively show public corpora (since the app UI depends on this client).

Also, added better error handling to the CLI.

### DIFF
--- a/packages/fixie/package.json
+++ b/packages/fixie/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fixie",
-  "version": "1.5.1",
+  "version": "1.5.3",
   "license": "MIT",
   "repository": "fixie-ai/ai-jsx",
   "bugs": "https://github.com/fixie-ai/ai-jsx/issues",
@@ -25,7 +25,7 @@
     "@types/apollo-upload-client": "^17.0.2",
     "ai-jsx": "workspace:*",
     "apollo-upload-client": "^17.0.0",
-    "axios": "^1.5.0",
+    "axios": "^1.5.1",
     "commander": "^11.0.0",
     "execa": "^8.0.1",
     "extract-files": "^13.0.0",
--- a/packages/fixie/package.json
+++ b/packages/fixie/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fixie",
-  "version": "1.5.3",
+  "version": "1.5.4",
   "license": "MIT",
   "repository": "fixie-ai/ai-jsx",
   "bugs": "https://github.com/fixie-ai/ai-jsx/issues",
--- a/packages/fixie/src/isomorphic-client.ts
+++ b/packages/fixie/src/isomorphic-client.ts
@@ -100,6 +100,8 @@ export class IsomorphicFixieClient {
       ...options,
       method: fetchMethod,
       headers,
+      // This is needed so serverside NextJS doesn't cache POSTs.
+      cache: 'no-store',
       // eslint-disable-next-line
       body: bodyData ? JSON.stringify(bodyData) : undefined,
     }).catch((err) => {
--- a/packages/fixie/src/isomorphic-client.ts
+++ b/packages/fixie/src/isomorphic-client.ts
@@ -1,6 +1,5 @@
 import type { Jsonifiable } from 'type-fest';
 import { AgentId, ConversationId, MessageGenerationParams, MessageRequestParams } from './sidekick-types.js';
-import axios, { AxiosRequestConfig } from 'axios';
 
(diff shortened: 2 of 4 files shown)


packages/fixie/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fixie",
-  "version": "1.5.1",
+  "version": "1.5.3",
```
NickHeiner, line 3: What happened to `1.5.2.`? :smile:

packages/fixie/src/main.ts
```diff
@@ -89,6 +89,39 @@ const currentPath = path.dirname(fileURLToPath(import.meta.url));
...
+      if (process.env.FIXIE_API_URL) {
+        term('❌ FIXIE_API_URL is set to ').red(process.env.FIXIE_API_URL)('\n');
+        term('❌ Check to ensure that this is the correct API endpoint.\n');
+      }
+      if (process.env.FIXIE_API_KEY) {
+        term('❌ ').green('FIXIE_API_KEY')(' is set to ').red(process.env.FIXIE_API_KEY.slice(0, 12))('...\n');
+        term('❌ Check to ensure that this is the correct key.\n');
+      }
+    } else if (error.response.status == 400) {
+      term.red('❌ Client made bad request to ').green(`${url}.\n`);
```
NickHeiner, line 107: Maybe suggest that the user is on the latest version in this case?

packages/fixie/src/main.ts
```diff
@@ -102,76 +135,103 @@ registerServeCommand(program);
 program
   .command('user')
   .description('Get information on the current user')
-  .action(async () => {
-    const client = await AuthenticateOrLogIn({ apiUrl: program.opts().url });
-    const result = await client.userInfo();
-    showResult(result, program.opts().raw);
-  });
+  .action(
+    catchErrors(async () => {
```
NickHeiner, line 139: Hmm, commander doesn't have a way to set this at a command level, so we don't need to manually wrap anything? Bummer.

packages/fixie/src/isomorphic-client.ts
```diff
@@ -72,28 +73,24 @@ export class IsomorphicFixieClient {
     if (this.apiKey) {
       headers.Authorization = `Bearer ${this.apiKey}`;
     }
-
     if (debug) {
       console.log(`[Fixie request] ${this.url}${path}`, bodyData);
     }
-    const res = await fetch(`${this.url}${path}`, {
+    const res = await axios.request({
```
NickHeiner, line 79: This will introduce `axios` to the client bundle, which will massively increase the size of what devs need to accept in order to use `fixie/web`. This is not necessarily a dispositive consideration but I think it's worth being rigorous about what we're getting in exchange. 

Using https://minify-js.com/, I estimate the `IsomorphicClient` is ~3.5kb minified. By contrast, [axios is `~30kb`](https://bundlephobia.com/package/axios@1.5.1). So we're roughly 10x-ing the size of our bundle.

Axios was originally developed before native `fetch` was available. In the olden days, you'd definitely want to use it or `got`. Nowadays, I think it's less needed in general.

mdwelsh: Makes sense - will do this the old-fashioned way

packages/fixie/src/main.ts
mdwelsh: Good idea, done!

packages/fixie/src/main.ts
mdwelsh: Yup

mdwelsh (commented):
Thanks for the comments! PTAL.

packages/fixie/src/main.ts
```diff
@@ -89,6 +90,41 @@ const currentPath = path.dirname(fileURLToPath(import.meta.url));
...
+      term('❌ Error accessing Fixie API at ').green(url)(': ')(error.message)('\n');
+    }
+    term.green(JSON.stringify(error.detail, null, 2));
+  } else {
+    term('❌ Error: ')(error.message)('\n');
+  }
+}
+
+function catchErrors(fn: (...args: any[]) => Promise<any>) {
+  return (...args: any[]) => fn(...args).catch(errorHandler);
```
petersalas, line 125: nit:

```
try {
   await fn(...args);
catch (ex) {
   errorHandler(ex);
}
```

would also catch errors if `fn` returns a promise but is not `async`.

petersalas (approved):
LGTM

PR #351, opened 2023-09-27
Try to be more robust about subprocess death.

mdwelsh:
Trying to fix @benlower's issue where the agent restarts before the port is released. Apparently `close` happens after `exit` so maybe this will fix it?


NickHeiner (approved):
https://fixie-ai.slack.com/archives/C03H4FX54JZ/p1695858903071659?thread_ts=1695853003.312009&cid=C03H4FX54JZ

This is an improvement over the current behavior.

PR #352, opened 2023-09-27
Preload tokenizer to avoid long delay on first generation

petersalas:
Previously the OpenAI tokenizer was loaded lazily on first generation. Empirically, loading the tokenizer can take hundreds of ms, so this preloads it on module load rather than first use.

Also:
- Share the tokenizer instance between `LargeResponseHandler` and `ChatCompletion`
- Wrap `JSON.stringify` in `debug` with a try/catch and use a fallback value

### DIFF
--- a/packages/ai-jsx/src/core/debug.tsx
+++ b/packages/ai-jsx/src/core/debug.tsx
@@ -150,10 +150,16 @@ export function debug(value: unknown, expandJSXChildren: boolean = true): string
           return `[${values.join(', ')}]`;
       }
     } else if (typeof value === 'object') {
-      if (context === 'props' || context === 'children') {
-        return `{${JSON.stringify(value)}}`;
+      let stringified;
+      try {
+        stringified = JSON.stringify(value);
+      } catch {
+        stringified = '{/* ... */}';
       }
-      return JSON.stringify(value);
+      if (context === 'props' || context === 'children') {
+        return `{${stringified}}`;
+      }
+      return stringified;
     } else if (typeof value === 'function') {
       const toRender = value.name === '' ? value.toString() : value.name;
       if (context === 'props' || context === 'children') {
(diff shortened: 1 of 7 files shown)


packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -131,10 +132,14 @@ export function OpenAI({
   return result;
 }
 
-const getEncoder = _.once(() => getEncoding('cl100k_base'));
+// Preload the tokenizer to avoid a large delay on first use.
+const tiktoken = new Tiktoken(cl100k_base);
+export const tokenizer = {
+  encode: (text: string) => tiktoken.encode(text),
+  decode: (tokens: number[]) => tiktoken.decode(tokens),
+};
```
farzadab, line 140: I'm confused, how is the `tiktoken` object different from the `tokenizer` object?
Are you just trying to make the interface cleaner by removing access to other functions or is this needed to trick TypeScript somehow?

farzadab (approved):
LGTM, just added one question.

packages/ai-jsx/src/lib/openai.tsx
farzadab: Relatedly (maybe answered by the other question): why is the first object called `tiktoken` and not something like `openAITikToken` or `cl100kTikToken`?

petersalas: 1) Since we're now exporting this, I wanted to encapsulate (i.e. narrow) the Tiktoken interface.
2) Simply in the spirit of `const foo = new Foo(bar)`

farzadab: Okay works for me.

packages/ai-jsx/src/core/debug.tsx
```diff
@@ -150,10 +150,16 @@ export function debug(value: unknown, expandJSXChildren: boolean = true): string
           return `[${values.join(', ')}]`;
       }
     } else if (typeof value === 'object') {
+      let stringified;
+      try {
+        stringified = JSON.stringify(value);
+      } catch {
+        stringified = '{/* ... */}';
```
NickHeiner, line 157: What's the scenario where this comes up?

NickHeiner:
Big oof. Good find.

packages/ai-jsx/src/core/debug.tsx
petersalas: If the OpenAI client gets JSON-stringified (which happens if you set the env bar to otel-trace every render in a fixie deployment) it throws.

PR #353, opened 2023-09-28
ASR tool improvements

juberti:
- Compute transcript latency using a common clock, allowing comparisons between providers.
- Improved output for Speechmatics ASR.
- Use a react component for the ASR display rather than a textarea, allowing for graying out of partial transcripts and other visual niceties.
- Add links to the API docs for the various ASR providers.


packages/ai-jsx/src/lib/asr/asr.tsx
```diff
@@ -313,40 +357,73 @@ export abstract class SpeechRecognitionBase extends EventTarget {
       }
       this.streamSentMillis += (chunk.byteLength / (2 * this.manager.sampleRate)) * 1000;
     });
+    this.manager.addEventListener('vad', (evt: CustomEventInit<VoiceActivity>) => {
+      const update = evt.detail!;
+      if (!update.active) {
+        if (this.streamLastVoiceEndMillis == 0 && this.streamFirstTranscriptMillis != 0) {
```
NickHeiner, line 363: It probably won't matter in practice, but use `===` unless you also want to match `'0'`.

```
> '0' == 0
true
```

packages/nextjs-demo/src/app/asr/page.tsx
```diff
@@ -41,11 +63,12 @@ interface AsrProps {
 }
 
 const Asr: React.FC<AsrProps> = ({ name, link, id, costPerMinute, manager, transcript }) => {
-  const [output, setOutput] = useState('');
+  const [output, setOutput] = useState<Transcript[]>([]);
+  const outputRef = useRef(output);
```
NickHeiner, line 67: It feels weird to store the same value as a `ref` and as `state`, but not a huge deal.

NickHeiner (approved):

packages/nextjs-demo/src/app/asr/page.tsx
NickHeiner: `state` is for things that impact the render result; `ref` is for everything else. In this case, it looks like `output` does eventually flow into the render result, so we'd use that and drop the `ref`.

juberti: The real question here though is whether we should be rerendering (and reconnecting event listeners) on every new token. Plus I need the ref to ensure the div is scrolled to the bottom. Wdyt?

NickHeiner: Reconnecting event listeners on every new token doesn't seem right.

But overall, if things are working, I wouldn't worry more about this.

juberti: don't we need to reconnect event listeners on every rerender, since the actual JS callback is new each time?

PR #361, opened 2023-10-02
Simplify system prompts in base Sidekick

zkoch:
This PR is trying to simplify what a base "Sidekick" is and does. I'm trying to remove as many system messages so it's easier to customize and get the outcome you want. To that end, this PR:

* Kills the "role" idea
* Solves a duplicate system prompt bug
* Experiments with moving the user-provided system message to after the conversation history


packages/ai-jsx/src/batteries/sidekick/platform/sidekick.tsx
```diff
@@ -131,8 +126,8 @@ export function Sidekick(props: SidekickProps) {
             userProvidedGenUIUsageExamples={props.genUIExamples}
             userProvidedGenUIComponentNames={props.genUIComponentNames}
           />
-          {props.systemMessage}
           <ConversationHistory />
+          {props.systemMessage}
```
NickHeiner, line 130: I think rather than doing this and/or in addition to it, you want to remove the `finalSystemMessageBeforeResponse` prop.

@petersalas I notice that `finalSystemMessageBeforeResponse` is only actually used in the `functionResponse` case. In the original Sidekick implementation, that made sense, because we were trying to force the LLM to use a `Card`. 

But now that we're being more generic and/or not using cards, maybe we should kill `finalSystemMessageBeforeResponse` entirely and just do what @zkoch is doing here?

packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
```diff
@@ -29,21 +29,20 @@ export function SidekickSystemMessage({
...
-      Do not say "according to the documents I have been given"; just answer as if you know the answer. When you see
-      timestamps in your source data, format them in a nice, human-readable way. Speak politely but casually. Say `use`
+      Do not say "according to the documents I have been given", just answer as if you know the answer. When you see
+      timestamps in your source data, format them in a human-readable way. Speak politely but casually. Say `use`
       instead of `utilize`.
     </SystemMessage>
   );
 
-  const baseComponentNames = ['Card', 'Citation'];
+  const baseComponentNames = ['Citation'];
```
NickHeiner, line 45: If we're taking out Card, we should take it out everywhere, including from the MDX examples.

zkoch: I was hoping this was the simplest way to remove it, since I don't want it being rendered but might want to bring back a modified form later

NickHeiner: Given that everything in the system message competes for attention, I think we should go through the effort to remove it universally. We can always bring it back via the git history.

packages/ai-jsx/src/batteries/sidekick/platform/sidekick.tsx
petersalas: Seems reasonable to me.

PR #364, opened 2023-10-04
Use AudioWorklet to play out TTS audio instead of MSE

juberti:
- works on iOS
- fixes issue where TTS audio could get "stuck"
- allows use of PCM format from TTS providers, which will allow further latency optimizations
- also get Resemble AI TTS working


packages/ai-jsx/src/lib/tts/tts.tsx
```diff
@@ -1,6 +1,207 @@
...
+    this.bufPos = 0;
+    this.port.onmessage = (e) => {
+      this.msgs[e.data.seqNum] = e.data;
+      //console.debug('buf added, seq num = ' + e.data.seqNum + ' queue len=' + this.queue.length.toString());
+    } 
+  }
+  process(inputs, outputs) {
+    const output = outputs[0];
+    const outLeft = output[0];
+    const outRight = output[1];
```
NickHeiner, line 19:
```suggestion
    const [[outLeft, outRight]] = outputs;
```

packages/ai-jsx/src/lib/tts/tts.tsx
```diff
@@ -1,6 +1,207 @@
...
+    this.sampleRate = sampleRatePiece ? parseInt(sampleRatePiece.split('=')[1]) : 0;
+  }
+  get isPcm() {
+    return this.mimeType == AUDIO_PCM_MIME_TYPE || this.mimeType == AUDIO_L16_MIME_TYPE;
+  }
+}
+
+class AudioMessage {
+  constructor(public seqNum: number, public channelData: Float32Array[]) {}
+}
```
NickHeiner, line 87: This can also just be an interface.

juberti: OOC, why use an interface here? And how would I pass this to `postMessage` since I can no longer do `new AudioMessage()`?

NickHeiner:
```ts
// as a class
const audioMessage = new AudioMessage(seqNum, channelData);

// as an interface
const audioMessage = { seqNum, channelData }
```

NickHeiner: You can't just pass that object to `postMessage`?

juberti: Sure, but then it's no longer typed, right? I.e., what's the point of the interface definition if I'm just using {}?

NickHeiner: You generally type it at the place where you need it to have a type, such as a function param. This maximizes flexibility. 

But if you want to ensure it has a type at creation time, you can do:

```
const a: TypeName = { key: val }
```

NickHeiner: Types are, of course, erased at runtime.

PR #366, opened 2023-10-04
Add support for inferring custom runtime parameters

petersalas:
This adds CLI support for inferring/setting the schema for custom runtime parameters. We use `typescript-json-schema` to infer the schema and set it in GraphQL. (This change does not add CLI support for setting default runtime parameters, but adds it to the underlying GraphQL mutation.)

The bulk of the change consists of refactoring `fixie serve` so that _either_ a tunnel URL change or parameter schema change triggers a new revision.

### DIFF
--- a/packages/fixie/src/agent.ts
+++ b/packages/fixie/src/agent.ts
@@ -288,7 +288,7 @@ export class FixieAgent {
     );
     const program = TJS.programFromConfig(tsconfigPath, [tempPath]);
     const schema = TJS.generateSchema(program, 'RuntimeParameters', settings);
-    if (schema?.type !== 'object') {
+    if (schema && schema.type !== 'object') {
       throw new Error(`The first argument of your default export must be an object (not ${schema.type})`);
     }
 
--- a/packages/fixie/src/agent.ts
+++ b/packages/fixie/src/agent.ts
@@ -267,7 +267,7 @@ export class FixieAgent {
     // with the runtime parameters for the agent.
     const tsconfigPath = path.resolve(path.join(agentPath, 'tsconfig.json'));
     if (!fs.existsSync(tsconfigPath)) {
-      term.yellow('⚠️ tsconfig.json not found. Your agent will not support runtime parameters.\n');
+      term.yellow(`⚠️ tsconfig.json not found at ${tsconfigPath}. Your agent will not support runtime parameters.\n`);
       return null;
     }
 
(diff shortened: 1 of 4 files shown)


packages/fixie/src/agent.ts
```diff
@@ -260,6 +262,39 @@ export class FixieAgent {
     return config as AgentConfig;
   }
 
+  private static inferRuntimeParametersSchema(agentPath: string): TJS.Definition | null {
+    // If there's a tsconfig.json file, try to use Typescript to produce a JSON schema
+    // with the runtime parameters for the agent.
+    const tsconfigPath = path.resolve(path.join(agentPath, 'tsconfig.json'));
+    if (!fs.existsSync(tsconfigPath)) {
+      term.yellow('⚠️ tsconfig.json not found. Your agent will not support runtime parameters.\n');
```
NickHeiner, line 270: When giving a "not found" message, it's always nice to include the full path that was checked.

packages/fixie/src/agent.ts
```diff
@@ -260,6 +262,39 @@ export class FixieAgent {
...
+    fs.writeFileSync(
+      tempPath,
+      `
+      import Handler from '${handlerPath}';
+      export type RuntimeParameters = Parameters<typeof Handler>[0] extends infer T ? T : {};
+      `
+    );
+    const program = TJS.programFromConfig(tsconfigPath, [tempPath]);
+    const schema = TJS.generateSchema(program, 'RuntimeParameters', settings);
+    if (schema && schema.type !== 'object') {
```
NickHeiner, line 291:
```suggestion
    if (schema?.type !== 'object') {
```

packages/fixie/src/agent.ts
```diff
@@ -684,8 +655,11 @@ export class FixieAgent {
 
     // The tunnel may yield different URLs over time. We need to create a new
     // agent revision each time.
-    for await (const currentUrl of deploymentUrlsIter) {
-      await waitForAgentToBeReady();
+    for await (const [currentUrl, runtimeParametersSchema] of this.zipAsyncIterables(
+      deploymentUrlsIter,
+      schemaGenerator
+    )) {
```
NickHeiner, line 661: I feel like we should just be using event emitters but I don't think it's worth trying to change now.

NickHeiner (approved):
LGTM on an extremely quick review

packages/fixie/src/agent.ts
petersalas: Probably -- I briefly started down that path but (probably incorrectly) convinced myself it wasn't going to be significantly better.

packages/fixie/src/agent.ts
petersalas: Actually, this is intentional -- `null !== "object"`

PR #367, opened 2023-10-04
Run the Sidekick with different models

NickHeiner:
Related to https://github.com/fixie-ai/fixie/pull/1883.

### DIFF
--- a/packages/ai-jsx/src/batteries/sidekick/platform/sidekick.tsx
+++ b/packages/ai-jsx/src/batteries/sidekick/platform/sidekick.tsx
@@ -76,18 +76,18 @@ export type SidekickProps = UniversalSidekickProps & OutputFormatSidekickProps;
 
 export function Sidekick(props: SidekickProps) {
   return (
-      <ShowConversation present={present}>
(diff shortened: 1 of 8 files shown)


packages/ai-jsx/src/batteries/sidekick/platform/sidekick.tsx
```diff
@@ -108,10 +106,16 @@ type OutputFormatSidekickProps = MergeExclusive<
 export type SidekickProps = UniversalSidekickProps & OutputFormatSidekickProps;
 
 export function Sidekick(props: SidekickProps) {
+  const model = props.model ?? 'gpt-4-32k';
```
petersalas, line 112: Why does `Sidekick` do this at all instead of just using whatever is set on context?

NickHeiner: Good question. The main reason I wanted it was so we could do this error message, but I guess this error message is redundant with the one `UseTools` will throw. (`UseTools`' message is more generic and less actionable/useful, but we could fix that with an error boundary.)

petersalas (approved):

PR #368, opened 2023-10-04
Fixing type-fest dependency

benlower:
Moving type-fest to dependencies. Incrementing package to version 1.7.1.

### DIFF
--- a/packages/fixie/readme.md
+++ b/packages/fixie/readme.md
@@ -64,4 +64,8 @@ This low-level API provides direct access to the Fixie Conversation and Corpus A
 
 ## Dev Notes
 
-To publish this package, you first need to update the version number in `package.json`. Then run `yarn npm publish` in this directory.
+To publish this package:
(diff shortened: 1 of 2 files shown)


NickHeiner (approved):

packages/fixie/readme.md
```diff
@@ -64,4 +64,8 @@ This low-level API provides direct access to the Fixie Conversation and Corpus A
 
 ## Dev Notes
 
-To publish this package, run `yarn npm publish` in this directory.
+To publish this package:
+
+1. Update the version number in `package.json`.
+1. Run `yarn install` in the root to get the changes to `yarn.lock`.
```
NickHeiner, line 70: This is a bit misleading, because you don't need to run `yarn install` after updating `package.json`.

The time you need to run `yarn install` is after updating the `dependencies` or `devDependencies` fields in `package.json`.

PR #369, opened 2023-10-04
Pull conversation history from request

petersalas:
If the request includes the contents of the conversation, use that instead of `fetch`ing it from the API.


mdwelsh (commented):
See my comment on fixie#1888 since I'm confused about how this is supposed to work!

mdwelsh (approved):

PR #370, opened 2023-10-04
Remove dependency between fixie and ai-jsx, add exports for fixie/web

petersalas:
In particular:
- The types in `sidekick-types.ts` are Fixie contracts (that happen to mirror the ai-jsx types as they exist today)
- This avoids the need to `git clean -f` after every build


packages/fixie/package.json
```diff
@@ -75,5 +74,10 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "files": ["dist"],
+  "exports": {
+    ".": "./dist/main.js",
+    "./web": "./dist/web.js"
```
petersalas, line 81: @NickHeiner you expressed some FUD around this approach -- is there a particular scenario that won't work/I should test?

packages/fixie/src/sidekick-types.ts
```diff
@@ -81,15 +79,15 @@ export interface BaseMessage extends StateFields {
 
 export interface FunctionCall extends BaseMessage {
   kind: 'functionCall';
-  name?: PropsOfComponent<typeof FunctionCallNode>['name'];
-  args?: PropsOfComponent<typeof FunctionCallNode>['args'];
+  name?: string;
```
NickHeiner, line 82: lol, you don't want to pull in an entire dep just to get a type alias for `string`?

packages/fixie/package.json
NickHeiner: I would try this:

1. Publish a prerelease version (`1.8.0-0`)
2. Try to use this package from `fixie/src/frontend`
3. Try to use this package from `fixie/src/generic-sidekick-frontend`
4. Try to use this package from https://github.com/fixie-ai/redwood-fixie-sample

If it works in all those places, ✨🎵 tHiNgS sHoUlD bE fInE 🎵 ✨

PR #374, opened 2023-10-05
Remove extra initial whitespace

petersalas:
`LimitToValidMdx` yielded an initial single whitespace character, but there's no reason to do so. This removes it.


packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -235,7 +235,6 @@ function RepairMdx({ children }: { children: string }) {
 }
 
 async function* LimitToValidMdx({ children }: { children: AI.Node }, { render, logger }: AI.ComponentContext) {
-  yield ' ';
```
petersalas, line 238: > but there's no reason to do so

Right @NickHeiner?

NickHeiner: I think so! 🤪

NickHeiner (approved):

PR #413, opened 2023-10-18
Support using a Fixie agent as the backing LLM

petersalas:
This allows the voice demo to use a Fixie agent as the backing LLM by adding the `?llm=fixie/$UUID` search param.

This required a couple affordances in the code related to message persistence:
- Persisting the conversation ID (this is currently tacked onto/read from generated messages in a janky way)
- Disabling speculative execution for Fixie agents as there is no way to revise a user message after the fact

`?llm=fixie/2a33fba6-905e-4069-a5e2-b0440194fd30` has the current system prompt pasted in. (The greeting message is fixed, however.)

### DIFF
--- a/packages/voice/src/app/agent/page.tsx
+++ b/packages/voice/src/app/agent/page.tsx
@@ -10,10 +10,6 @@ import {
 import { createTextToSpeech, BuildUrlOptions, TextToSpeechBase, TextToSpeechProtocol } from 'ai-jsx/lib/tts/tts';
 import { useSearchParams } from 'next/navigation';
 import Image from 'next/image';
-// The import fails with https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/CJSResolvesToESM.md
-// but ostensibly the bundler is able to muddle through it, so we suppress the error for TS.
-// @ts-expect-error
-import { IsomorphicFixieClient } from 'fixie/web.js';
 import '../globals.css';
 
 // 1. VAD triggers silence. (Latency here is frame size + VAD delay)
@@ -94,6 +90,25 @@ class ChatMessage {
 }
 
 /**
+ * Transforms a text stream of JSON lines into a stream of JSON objects.
+ */
+function jsonLinesTransformer() {
+  let buffer = '';
+  return new TransformStream<string, any>({
+    async transform(chunk, controller) {
+      buffer += chunk;
+      const lines = buffer.split('\n');
+      buffer = lines.pop()!;
+      for (const line of lines) {
+        if (line.trim()) {
+          controller.enqueue(JSON.parse(line));
+        }
+      }
+    },
+  });
+}
+
+/**
  * A single request to the LLM, which may be speculative.
  */
 class ChatRequest {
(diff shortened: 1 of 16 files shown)


packages/voice/src/app/agent/page.tsx
```diff
@@ -105,32 +110,129 @@ class ChatRequest {
     private readonly model: string,
     private readonly docs: boolean,
     public active: boolean
-  ) {}
+  ) {
+    this.conversationId = inMessages.find((m) => m.conversationId !== null)?.conversationId;
+  }
+
+  get didStart() {
```
juberti, line 117: surrounding style here is `isStarted`

packages/voice/src/app/agent/page.tsx
```diff
@@ -105,32 +110,129 @@ class ChatRequest {
...
-    const reader = res.body!.getReader();
-    // eslint-disable-next-line no-constant-condition
-    while (true) {
-      const { done, value } = await reader.read();
-      if (done) {
-        this.done = true;
-        this.streamLatency = performance.now() - this.startMillis - this.requestLatency;
-        this.onComplete?.(this);
-        break;
+    if (this.model.startsWith('fixie/')) {
```
juberti, line 122: the fixie case is pretty complex, let's break that out into its own function

packages/voice/src/app/agent/page.tsx
```diff
@@ -105,32 +110,129 @@ class ChatRequest {
...
+        });
+      } else {
+        isStartConversationRequest = true;
+        const conversationResult = await fixieClient.startConversation(agentId, this.inMessages.at(-1)!.content);
+        this.conversationId = conversationResult.conversationId;
+        response = conversationResult.response;
+      }
+
+      const reader = response.body!.pipeThrough(new TextDecoderStream()).getReader();
+      this.requestLatency = performance.now() - this.startMillis;
```
juberti, line 149: we should probably make this first-token latency to make it apples-apples

packages/voice/src/app/agent/page.tsx
```diff
@@ -105,32 +110,129 @@ class ChatRequest {
...
+          break;
+        }
+
+        if (!this.done) {
+          bufferedText += value;
+
+          const lines = bufferedText.split('\n');
+          bufferedText = lines.pop()!;
+          for (const line of lines) {
+            const json = JSON.parse(line);
```
juberti, line 170: Might want to break this line-processing out into its own function

packages/voice/src/app/agent/page.tsx
```diff
@@ -248,11 +352,15 @@ class ChatManager {
     } else if (final) {
       const request = this.pendingRequests[normalized];
       request.active = true;
-      this.tts.play(request.outMessage);
-      if (!request.done) {
-        this.onOutputChange?.(request.outMessage, false, request.requestLatency!);
+      if (!request.didStart) {
```
juberti, line 355: I think I'd probably handle this by just not starting requests for partial transcrips if we're using Fixie, then we can control this with a single if statement

juberti (approved):

PR #442, opened 2023-10-26
ASR: Fix partial transcript rendering with useRef

juberti:


petersalas (approved):
Looking at the `useEffect` on line 155, I suspect there will be some weirdness if the component is ever unmounted but that's not new in this change

PR #444, opened 2023-10-27
Clean up exports from `fixie` package.

mdwelsh:
Export `FixieAgent` from the `fixie` package, move all source files under `src/`, and build to `dist/`.

### DIFF
--- a/packages/fixie/src/index.ts
+++ b/packages/fixie/src/index.ts
@@ -1,3 +1,4 @@
 export { FixieClient } from './client.js';
 export { IsomorphicFixieClient } from './isomorphic-client.js';
+export { FixieAgent } from './agent.js';
 export * from './sidekick-types.js';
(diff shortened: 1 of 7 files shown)


packages/fixie/tsconfig.json
```diff
@@ -9,10 +9,10 @@
     "moduleResolution": "node16",
     "jsx": "react",
     "esModuleInterop": true,
-    "outDir": ".",
+    "outDir": "dist",
```
petersalas, line 12: I tried very hard to do this a couple weeks ago and failed because of CJS/ESM/Typescript `moduleResolution` nightmares (the resulting package couldn't be consumed in a straightforward manner in NextJS projects).

We should definitely do this, but I think we need to do it in concert with other work.

packages/fixie/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export { FixieClient } from './client.js';
+export { IsomorphicFixieClient } from './isomorphic-client.js';
+export { FixieAgent } from './agent.js';
```
petersalas, line 3: What's the scenario where it makes sense to write code against `FixieAgent`? At the moment, it's a set of very opinionated CLI commands right now which I think we should keep separate from the rest of the client SDK.

mdwelsh:
Closing this in favor of https://github.com/fixie-ai/ai-jsx/pull/445 which does the surgical thing I need

PR #470, opened 2023-11-02
Fix broken `corpus list` command.

mdwelsh:
@petersalas Feel free to fold this into your 4.0.0 PR if you like

### DIFF
--- a/packages/fixie/src/main.ts
+++ b/packages/fixie/src/main.ts
@@ -215,7 +215,7 @@ corpus
       } else if (owner === 'public') {
         ownerType = 'OWNER_PUBLIC';
       }
-      const result = await client.listCorpora(ownerType, opts.offset, opts.limit);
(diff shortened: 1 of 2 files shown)


petersalas:
@mdwelsh I already did (with some differences from this one)

packages/fixie/src/main.ts
```diff
@@ -215,7 +215,7 @@ corpus
       } else if (owner === 'public') {
         ownerType = 'OWNER_PUBLIC';
       }
-      const result = await client.listCorpora(ownerType, opts.offset, opts.limit);
+      const result = await client.listCorpora(ownerType, opts?.offset, opts?.limit);
```
petersalas, line 218: This one isn't quite right, you also need to spread ...opts

(see https://github.com/fixie-ai/ai-jsx/pull/466)

PR #483, opened 2023-11-07
Add support for OpenAI parallel function calls

petersalas:
This adds support for OpenAI's new parallel function calls, where a single inference can request multiple function calls.

It consists of:
- Updating the OpenAI SDK version
- Using the new "tools" contracts
- Plumbing IDs to relate function call requests and corresponding results
- Supporting executing multiple function calls in parallel in `<UseTools>` and `<Sidekick>`

### DIFF
--- a/packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
+++ b/packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
@@ -22,64 +22,37 @@ import { SidekickOutputFormat } from './sidekick.js';
  * window.
  */
 export function getShrinkableConversation(messages: ConversationMessage[], fullConversation: ConversationMessage[]) {
-  // Group messages with tool IDs together because we can't leave unmatched function calls/responses.
-  const groupedMessages: ConversationMessage[][] = [[]];
-  fullConversation.forEach((message) => {
-    if (message.type === 'functionCall' || message.type === 'functionResponse') {
-      if (message.element.props.id) {
-        const lastMessage = groupedMessages.at(-1)?.at(-1);
-        if (lastMessage?.type === 'functionCall' || lastMessage?.type === 'functionResponse') {
-          if (lastMessage.element.props.id) {
-            groupedMessages.at(-1)?.push(message);
-            return;
-          }
-        }
-      }
-    }
-
-    groupedMessages.push([message]);
-  });
-
-  return groupedMessages.map((group, groupIndex) => {
+  return fullConversation.map((message, messageIndex) => {
     // Ensure that nothing in the most recent batch of messages gets dropped.
-    if (messages.length !== fullConversation.length && messages.find((m) => group.includes(m))) {
-      return group.map((message) => message.element);
+    if (messages.length !== fullConversation.length && messages.includes(message)) {
+      return message.element;
     }
 
-    if (group.find((message) => message.type === 'functionResponse')) {
-      // It's a group of function calls and responses. As a first pass, elide FunctionResponses.
-      const elements = group.map((message) =>
-        message.type === 'functionResponse' ? (
+    switch (message.type) {
+      case 'system':
+        // Never drop system messages.
+        return message.element;
+      case 'functionResponse':
+        // As a first pass, elide FunctionResponses.
+        return (
           <Shrinkable
             importance={0}
             replacement={
-              <FunctionResponse id={message.element.props.id} name={message.element.props.name}>
-                [snip...]
-              </FunctionResponse>
+              <Shrinkable importance={messageIndex + 1}>
+                <FunctionResponse id={message.element.props.id} name={message.element.props.name}>
+                  [snip...]
+                </FunctionResponse>
+              </Shrinkable>
             }
           >
             {message.element}
           </Shrinkable>
-        ) : (
-          message.element
-        )
-      );
-
-      // Then prune the whole group.
-      return <Shrinkable importance={groupIndex + 1}>{elements}</Shrinkable>;
-    }
-
-    const message = group[0];
-    switch (message?.type) {
-      case 'system':
-        // Never drop system messages.
-        return message.element;
+        );
       case 'user':
       case 'assistant':
       case 'functionCall':
-      case 'functionResponse':
         // Then prune oldest -> newest messages.
-        return <Shrinkable importance={groupIndex + 1}>{message.element}</Shrinkable>;
+        return <Shrinkable importance={messageIndex + 1}>{message.element}</Shrinkable>;
     }
   });
 }
--- a/packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
+++ b/packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
@@ -39,7 +39,9 @@ export function getShrinkableConversation(messages: ConversationMessage[], fullC
             importance={0}
             replacement={
               <Shrinkable importance={messageIndex + 1}>
-                <FunctionResponse name={message.element.props.name}>[snip...]</FunctionResponse>
+                <FunctionResponse id={message.element.props.id} name={message.element.props.name}>
+                  [snip...]
+                </FunctionResponse>
               </Shrinkable>
             }
           >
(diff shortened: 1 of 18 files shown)


packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -492,31 +539,44 @@ export async function* OpenAIChatModel(
 
     // TS doesn't realize that the Stream closure can make `delta` be `null`.
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    if (delta?.function_call) {
+    while (delta?.tool_calls) {
```
farzadab, line 542: Q: so this was a breaking change or does this name change only happen at their SDK layer (not the API layer)?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -387,23 +409,48 @@ export async function* OpenAIChatModel(
...
     max_tokens: props.maxTokens,
     temperature: props.temperature,
     top_p: props.topP,
-    messages,
-    functions: openaiFunctions.length > 0 ? openaiFunctions : undefined,
-    function_call: props.forcedFunction ? { name: props.forcedFunction } : undefined,
+    messages: mergedMessages,
+    tools: openaiTools.length > 0 ? openaiTools : undefined,
+    tool_choice: props.forcedFunction
+      ? { function: { name: props.forcedFunction }, type: 'function' as const }
```
farzadab, line 452: The idea is that these are `functions` and there can be other `tools` such as Code Interpreter that may be hard coded? Is that what `type` distinguishes?

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -93,25 +95,38 @@ export function getNextConversationStep(
...
-      if (tools && !(name in tools)) {
-        return executedFunction;
-      }
-      // Function responses can potentially be very large. In that case, we need
-      // some way of handling that so the context window doesn't blow up.
-      return (
-        <LargeFunctionResponseWrapper numChunks={10} maxLength={10500} failedMaxLength={4000}>
-          {executedFunction}
-        </LargeFunctionResponseWrapper>
+      // Collect all the adjacent function calls and execute them.
```
farzadab, line 98: Double check: so now the last message can either be a single `AssistantMessage` or a bunch of `FunctionCall`s?

farzadab (approved):
LGTM as long as it runs. My comments are mostly for my understanding.

packages/ai-jsx/src/lib/openai.tsx
petersalas: I think if you use `functions` on the request, they use `function_call` in the response; if you use `tools` on the request, they use `tool_ calls` in the response.

packages/ai-jsx/src/lib/openai.tsx
petersalas: Per their [docs](https://platform.openai.com/docs/api-reference/chat/create), currently only `"function"` is supported. But presumably they're leaving the door open for other types in the future.

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
petersalas: Yup, a single `<ChatCompletion>` can now produce multiple `<FunctionCall>`s.

packages/ai-jsx/src/lib/openai.tsx
farzadab: Seems to me like we can already pass in `{"type": "code_interpreter"}` and `{"type": "retrieval"}` to use OAI-native tools:
image
image

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -353,14 +411,28 @@ export async function* OpenAIChatModel(
...
+              tool_calls: [
+                {
+                  type: 'function',
+                  function: { name: message.element.props.name, arguments: JSON.stringify(message.element.props.args) },
+                  id: message.element.props.id,
+                },
+              ],
+            };
+          }
+
```
farzadab, line 435: GH doesn't allow me to link it, but on on line 367:439, we still emit an assistant message with a `function_call`. Can't we merge all of them to only have `tool_calls`? Or do we have to support both for legacy reasons?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -370,6 +442,14 @@ export async function* OpenAIChatModel(
...
         case 'functionResponse':
+          if (message.element.props.id && !unmatchedIds.has(message.element.props.id)) {
+            return {
+              role: 'tool',
+              tool_call_id: message.element.props.id,
+              content: await render(message.element.props.children),
+            };
+          }
+
           return {
```
farzadab, line 453: Same here, can't we only have `tool` not both `tool` and `function`?

farzadab: Or will OAI require IDs for all `tools` but not `function`s?

farzadab: Actually, you said that if we use `functions` they will respond with `delta.function_call`, so this is probably more serious since we don't handle that anymore.

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -341,8 +347,60 @@ export async function* OpenAIChatModel(
     promptTokenLimit
   );
 
-  const messages: OpenAIClient.Chat.CreateChatCompletionRequestMessage[] = await Promise.all(
-    conversationMessages.map(async (message) => {
+  // OpenAI requires that tool calls with IDs be immediately followed by tool messages with the corresponding IDs.
```
farzadab, line 350: Nit: a general comment that this function is now 350 lines long :)

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -370,6 +442,14 @@ export async function* OpenAIChatModel(
             },
           };
         case 'functionResponse':
+          if (message.element.props.id && !unmatchedIds.has(message.element.props.id)) {
```
farzadab, line 445: Does OAI throw an error for unmatched/incorrectly matched function responses? If not, do we really need to be so meticulous in removing them?

packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -341,8 +347,60 @@ export async function* OpenAIChatModel(
...
+      !activeFunctionResponses.every((id) => dedupedFunctionCalls.has(id));
+
+    if (isInvalid) {
+      for (const id of activeFunctionCalls) {
+        unmatchedIds.add(id);
+      }
+      for (const id of activeFunctionResponses) {
+        unmatchedIds.add(id);
+      }
+    }
```
farzadab, line 375: Isn't this a bit harsh? Shouldn't we just set the ones that didn't match to `unmatchedIds` instead of invalidating the whole run?

petersalas: The challenge is that the API asserts the invariant very strictly -- so the entire batch needs to fully matched/valid or else OAI will reject the request.

But it does allow a single request to have mixed old- and new-style messages -- how the response is emitted is determined by whether the functions were _defined_ as tools vs. functions.

packages/ai-jsx/src/lib/openai.tsx
petersalas: They do :)

packages/ai-jsx/src/lib/openai.tsx
petersalas: > Or will OAI require IDs for all tools but not functions?

Yup exactly. OAI is strict w.r.t. to tools, but not w.r.t. functions.

packages/ai-jsx/src/lib/openai.tsx
juberti: yeah, hoisting this ID management to a helper would be a nice simplification.

juberti (approved):

PR #487, opened 2023-11-08
Update to new user API.

mdwelsh:
This PR updates the `fixie` package to use the new `/api/v1/user/me` API for getting and updating details on the current user. This is needed so we can support this in the web UI.

PR #490, opened 2023-11-09
Add support for Teams API.

mdwelsh:
This PR is stacked on #487.

This PR adds support for the Teams API to both client.ts and the CLI. It allows one to create a team, invite members, cancel invitations, and update and remove members from a team. I'll use this in an upcoming PR to add Teams support to the web UI.


packages/fixie/src/client.ts
```diff
@@ -746,4 +756,159 @@ export class FixieClient {
...
+
+  /** Delete the given team. */
+  deleteTeam({ teamId }: { teamId: string }): Promise<Jsonifiable> {
+    return this.requestJson(`/api/v1/teams/${teamId}`, undefined, 'DELETE');
+  }
+
+  /**
+   * List the teams visible to the current user.
+   *
+   * @param options.offset The offset into the list of Sources to return.
```
petersalas, line 800: jsdoc comments are wrong

packages/fixie/src/client.ts
```diff
@@ -746,4 +756,159 @@ export class FixieClient {
...
+   * @param options.limit The maximum number of Sources to return.
+   */
+  listTeams({ offset = 0, limit = 100 }: { offset?: number; limit?: number }): Promise<Jsonifiable> {
```
petersalas, line 803: Can you annotate the type to be narrower than `Jsonifiable`?

packages/fixie/src/client.ts
```diff
@@ -746,4 +756,159 @@ export class FixieClient {
...
+   */
+  inviteTeamMember({
+    teamId,
+    email,
+    isAdmin,
+  }: {
+    teamId: string;
+    email: string;
+    isAdmin?: boolean;
+  }): Promise<Jsonifiable> {
```
petersalas, line 858: ditto on the rest of these `Jsonifiable`s

petersalas (approved):
nice!

packages/fixie/src/client.ts
mdwelsh: Not easily, at least not until we have a better way of generating TS from the proto types. Rather than do a clumsy job at this I prefer to leave the types underspecified since otherwise it's a headache to keep things consistent between the proto and TS code.

PR #491, opened 2023-11-10
Give <Sidekick> the ability to interject messages prior to the first function call

petersalas:
This adds the ability for the `<Sidekick>` component to interject assistant text (e.g. "Let me check on that.") if the first message in a generation is a function call and the most recent `<AssistantMessage>` was not an interjection.

It's currently empty by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "eslint-import-resolver-typescript": "^3.5.5",
     "eslint-plugin-import": "^2.27.5",
     "prettier": "^2.8.8",
-    "turbo": "^1.10.4",
+    "turbo": "^1.10.16",
     "typescript": "^5.1.3"
   },
   "type": "module",
--- a/packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
+++ b/packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
@@ -106,7 +106,7 @@ export function getNextConversationStep(
   fullConversation: ConversationMessage[],
   outputFormat: SidekickOutputFormat,
   tools: UseToolsProps['tools'] | undefined,
-  functionCallInterjection: AI.Node
+  beforeFunctionCallInterjection: AI.Node
 ) {
   const shrinkableConversation = getShrinkableConversation(messages, fullConversation);
   const lastMessage = messages[messages.length - 1];
@@ -170,7 +170,7 @@ export function getNextConversationStep(
       const lastAssistantMessage = messages.findLast((m) => m.type === 'assistant');
       if (
         hasTools &&
-        functionCallInterjection &&
+        beforeFunctionCallInterjection &&
         // Ensure we don't interject if the last assistant message was already an interjection. (If back to back generations
         // request function calls, we don't want to interject twice.)
         !lastAssistantMessage?.element.props.metadata?.isFunctionCallInterjection
@@ -179,7 +179,7 @@ export function getNextConversationStep(
           <InterjectBeforeFunctionCall
             interjection={
               <AssistantMessage metadata={{ isFunctionCallInterjection: true }}>
-                {functionCallInterjection}
+                {beforeFunctionCallInterjection}
               </AssistantMessage>
             }
           >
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "eslint-import-resolver-typescript": "^3.5.5",
     "eslint-plugin-import": "^2.27.5",
     "prettier": "^2.8.8",
-    "turbo": "^1.10.4",
+    "turbo": "^1.10.16",
```
petersalas, line 11: I upgraded turbo in the hopes that it would stop breaking my terminal...but it didn't. But I went ahead and left it upgraded.

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -136,9 +161,33 @@ export function getNextConversationStep(
...
-        <ChatCompletion functionDefinitions={tools ? updatedTools : undefined}>{shrinkableConversation}</ChatCompletion>
+      let generation = (
+        <ChatCompletion functionDefinitions={hasTools ? updatedTools : undefined}>
+          {shrinkableConversation}
+        </ChatCompletion>
       );
+
+      const lastAssistantMessage = messages.findLast((m) => m.type === 'assistant');
+      if (
+        hasTools &&
```
farzadab, line 172: Nit: Why would we get a functionResponse if there are no tools?
And if we somehow get it, it's probably safe to interject still :?

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -136,9 +161,33 @@ export function getNextConversationStep(
     case 'system':
     case 'user':
     case 'functionResponse': {
-      const generation = (
-        <ChatCompletion functionDefinitions={tools ? updatedTools : undefined}>{shrinkableConversation}</ChatCompletion>
+      let generation = (
```
farzadab, line 164: Q: shouldn't we do this instead as soon as RAG is requested not when we are trying to respond to RAG, i.e. when trying to execute the function?
Or am I missing something?

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
```diff
@@ -68,6 +69,28 @@ export function present(conversationElement: ConversationMessage, outputFormat:
...
+ */
+function InterjectBeforeFunctionCall(
+  { interjection, children }: { interjection: AI.Node; children: AI.Node },
+  { memo }: AI.ComponentContext
+) {
+  const memoizedInterjection = memo(interjection);
+  return (
+    <ShowConversation
+      present={(message, index) => {
+        if (index === 0 && message.type === 'functionCall') {
```
farzadab, line 83: So we only interject if we see a `functionCall` in response to a `functionResponse`?
Or am I misunderstanding something?

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
petersalas: Yes, exactly -- note that this is wrapping the `<ChatCompletion>` (which produces `<FunctionCall>`s), not `<ExecuteFunction>` (which produces `<FunctionResponse>`s).

So as soon as we see that the `<ChatCompletion>` is _requesting_ a function call, the content will be interjected.

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
petersalas: We interject before the `functionCall` (which in turn precedes the `functionResponse`).

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
petersalas: You're right, I don't think there's any need to make this conditional on tools.

packages/ai-jsx/src/batteries/sidekick/platform/conversation.tsx
farzadab: Okay I understand the issue now. I was reading the code as `if case 'functionResponse' then do X`, but it was actually `if case system/user/response then do X`.

farzadab (approved):

juberti (approved):

PR #492, opened 2023-11-10
Slimming down system message

farzadab:
This PR tries to slim down the default System message for Sidekicks.

General changes:
- Trying to reduce unneeded words (I didn't go so far, but [LLMLingua](https://llmlingua.com/) shows that LLMs can understand a highly shortened format of language)
- The "document system message" is used only if there Sidekick has some `tools`. Corpus vs other tools are not disambiguated
- Reducing number of System messages: not a large factor, but I verified that each additional system message uses about 4 extra tokens
- Date format change
    - before: `11/10/2023, 1:46:34 PM .... Friday`
    - after: `Friday, Nov 10, 2023, 1:46:34 PM`
    - Interestingly, this uses slightly less tokens (not including actual prompt changes) while disambiguating 11/10 (US "standard" ...) to Nov 10

Token count for plain text mode (text only):
- before: 136 (probably slightly higher due to having multiple system messages)
- after (with tools): 77
- after (w/o tools): 56

### DIFF
--- a/packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
+++ b/packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
@@ -23,7 +23,7 @@ export function SidekickSystemMessage({
   useCitationCard,
   outputFormat,
 }: SidekickSystemMessageProps) {
-  /* Format: 'Friday, Nov 10, 2023, 1:46:34 PM' */
+  /* Format: 'Friday, Nov 10, 2023, 1:46:34 PM PST' */
   const dateStringOptions = {
     weekday: 'long',
     year: 'numeric',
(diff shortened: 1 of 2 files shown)


packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
```diff
@@ -69,10 +80,11 @@ export function SidekickSystemMessage({
   }
 
   return (
-    <>
+    <SystemMessage>
       {dateTimeSystemMessage}
-      {responseFramingSystemMessage}
-      {outputFormatToUse === 'text/markdown' && <SystemMessage>Respond with Markdown.</SystemMessage>}
+      {hasKnowledgeBase && responseFormatSystemMessage}
+      {styleSystemMessage}
```
petersalas, line 86: These will be joined without whitespace, e.g. "...just answer as if you know the answer.Speak politely but casually.Respond with Markdown."

It's perhaps unlikely to matter, but I guess I could imagine the model responding like that if enough of the prompt is like that?

packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
```diff
@@ -18,28 +19,38 @@ export function SidekickSystemMessage({
...
+  const dateStringOptions = {
+    weekday: 'long',
+    year: 'numeric',
+    month: 'short',
+    day: 'numeric',
+    hour: 'numeric',
+    minute: 'numeric',
+    second: 'numeric',
+    hour12: true,
+    timeZone,
```
petersalas, line 36: If the user time zone were specified, will it be obvious to the model that this time is in the user's time zone? Wondering if this should be explicitly UTC or explicitly in the user's time zone.

petersalas (approved):

packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
farzadab: Aah, I was wondering what would happen and instead of trying it I asked Perplexity and seems like the answer was wrong. Why is the formatting so weird? Just apply the conditionals first, then do HTML formatting :(

But yeah, this should not matter really. The result looks weird which I'll probably fix, but it seems like LLMs are comfortable even if a lot of the small details are omitted, not just spaces but stopwords and all.

packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
farzadab: Hmm, I agree that the confusion and possibly different behaviour based on where the server is located is not good and I should fix that.

Still I'd rather the LLM doesn't do math here. Maybe I'll set it as PDT time if `timeZone` is not provided (to bias against our likely audience?).

farzadab: Done.

packages/ai-jsx/src/batteries/sidekick/platform/system-message.tsx
farzadab: Done.

PR #493, opened 2023-11-13
Agent: Interact over WebRTC if URL param is set

juberti:
e.g. https://voice.fixie.ai/?webrtc=true

- audio visualizers aren't hooked up yet
- nor are state changes
- but audio flows both ways

### DIFF
--- a/packages/voice/src/app/agent/chat.tsx
+++ b/packages/voice/src/app/agent/chat.tsx
@@ -6,6 +6,17 @@ import {
   Transcript,
 } from 'ai-jsx/lib/asr/asr';
 import { createTextToSpeech, BuildUrlOptions, TextToSpeechBase, TextToSpeechProtocol } from 'ai-jsx/lib/tts/tts';
+import {
+  createLocalTracks,
+  DataPacket_Kind,
+  LocalAudioTrack,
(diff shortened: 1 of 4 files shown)


mdepinet (approved):

packages/voice/src/app/agent/chat.tsx
```diff
@@ -477,3 +510,175 @@ export class ChatManager {
...
+    return this._state;
+  }
+  get inputAnalyzer() {
+    return this.inAnalyzer?.analyzer;
+  }
+  get outputAnalyzer() {
+    return this.outAnalyzer?.analyzer;
+  }
+  warmup() {
+    const url = 'ws://localhost:8080/livekit'; //`${window.location.protocol === 'https:' ? 'wss:' : 'ws:'}//${window.location.host}/livekit')`;
```
zkoch, line 567: is this what you want?

juberti: Until we have a prod deploy, yes, I think

zkoch: This is what I see when trying to run the preview:

Screen Shot 2023-11-13 at 5 37 08 PM

PR #496, opened 2023-11-15
[Voice-QuickFix] Start speaking filler immediately

farzadab:
I noticed that the filler sentences (or interjections as Peter calls them, which I'm not a total fan of), don't start until after a long pause and seemed to only start after the corpus response was ready, which defeats the purpose. The issue seemed to have been due to the filler not being recognized as a separate sentence which meant TTS would not be called.

This PR adds a newline between separate `AssistantMessage` responses from Fixie to solve this issue for Fixie calls.

Note that I did not investigate what happens in the other cases (LLM not from Fixie).

Logs before: notice no chunk is being requested from TTS for a long time.
Screenshot 2023-11-15 at 9 47 08 AM

Logs after: TTS being called before actual content is ready, though it's still not easy to understand what's happening underneath when simply listening to it (there's no gap between filler TTS end and start of actual content).
Screenshot 2023-11-15 at 9 43 31 AM

### DIFF
--- a/packages/voice/src/app/agent/chat.tsx
+++ b/packages/voice/src/app/agent/chat.tsx
@@ -186,7 +186,7 @@ export class ChatRequest {
         const currentMessage = currentTurn.messages
           .filter((m: any) => m.kind === 'text')
           .map((m: any) => m.content)
-          .join('');
+          .join('\n');
 
         if (currentMessage === this.outMessage) {
           continue;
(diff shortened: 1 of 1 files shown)


juberti (commented):
New line seems kind of artificial here, perhaps this should just be a space?

farzadab:
Sure, either works. I kinda see different assistant messages as being more separate (i.e. totally separate paragraphs even) but I don't have a strong opinion here. I'll change it to a space.

petersalas (approved):
FWIW I don't really love "interjection" either as it's more of a description of the implementation (the code is interjecting w.r.t. the LLM) rather than the result. But that was the best I came up with 🤷‍♂️

PR #499, opened 2023-11-16
Introduce transcript newlines for function calls/responses

petersalas:
The TTS buffer only gets flushed when it sees content after a sentence. To ensure that it gets flushed after any pre-function-call filler text, this change:
 - Adds the `batchFrames` render option to ai-jsx to coalesce updates from different parts of the render tree. (This delays surfacing frames until further rendering is blocked on I/O.)
 - Adds message-level state tracking (`"in-progress" | "done"`) in `fixie-sdk` to text messages
 - Changes the fixie backend in the voice demo to emit newlines after `"done"` text messages
 - Stops appending text after `"in-progress"` text messages

### DIFF
--- a/packages/fixie-sdk/src/fixie-serve-bin.tsx
+++ b/packages/fixie-sdk/src/fixie-serve-bin.tsx
@@ -71,7 +71,10 @@ async function serve({
           <Handler {...(invokeAgentRequest.parameters ?? {})} />
         </FixieRequestWrapper>
       );
-      const generator = createRenderContext({ enableOpenTelemetry: true }).render(renderable)[Symbol.asyncIterator]();
+
+      // Enable frame batching to run ahead as aggressively as possible and ensure we don't have frame tearing. (See MessageState.)
+      const renderResult = createRenderContext({ enableOpenTelemetry: true }).render(renderable, { batchFrames: true });
+      const generator = renderResult[Symbol.asyncIterator]();
       return res
         .status(200)
         .type('application/jsonl')
@@ -79,15 +82,10 @@ async function serve({
           Readable.from(
             (async function* () {
               let lastMessages = [];
-              let currentValue = undefined as string | undefined;
               while (true) {
+                let currentValue = undefined as string | undefined;
                 try {
                   const next = await generator.next();
-                  if (!next.done && currentValue == next.value) {
-                    // No change, don't send anything.
-                    continue;
-                  }
-
                   currentValue = next.value;
                   lastMessages = currentValue
                     .split('\n')
--- a/packages/fixie-sdk/src/fixie-serve-bin.tsx
+++ b/packages/fixie-sdk/src/fixie-serve-bin.tsx
@@ -79,10 +79,15 @@ async function serve({
           Readable.from(
             (async function* () {
               let lastMessages = [];
+              let currentValue = undefined as string | undefined;
               while (true) {
-                let currentValue = undefined as string | undefined;
                 try {
                   const next = await generator.next();
+                  if (!next.done && currentValue == next.value) {
+                    // No change, don't send anything.
+                    continue;
+                  }
+
                   currentValue = next.value;
                   lastMessages = currentValue
                     .split('\n')
--- a/packages/fixie-sdk/src/request-wrapper.tsx
+++ b/packages/fixie-sdk/src/request-wrapper.tsx
@@ -14,7 +14,13 @@ export const RequestContext = AI.createContext<{
 } | null>(null);
 
 /**
- * Renders to "in-progress" while the children are still being rendered, and "done" when they're done.
+ * Renders to "in-progress" while `children` is still being rendered, and "done" when it's done.
+ *
+ * `children` should already be memoized to ensure that it's only rendered once.
+ *
+ * To ensure that this component renders consistently with `children`, a render containing both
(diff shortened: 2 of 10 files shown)


packages/voice/src/app/agent/chat.tsx
```diff
@@ -183,11 +183,10 @@ export class ChatRequest {
...
       if (!this.done) {
         const currentTurn = isStartConversationRequest ? value.turns.at(-1) : value;
-        const currentMessage = currentTurn.messages
-          .filter((m: any) => m.kind === 'text')
-          .map((m: any) => m.content)
-          .join(' ');
 
+        // Map non-text fragments to empty strings so that they don't appear in the transcript,
+        // but still introduce newlines to make clear that any earlier content is complete.
+        const currentMessage = currentTurn.messages.map((m: any) => (m.kind === 'text' ? m.content : '')).join('\n');
```
farzadab, line 189: Nit: I think Justin didn't like newline and preferred spaces (I do prefer newlines though). Not sure how strongly he felt about it. Also, this might create multiple newlines. Maybe try replacing `\n+` with `\n`? (Not quite sure how regex/replace works in JS).

farzadab (approved):

packages/voice/src/app/agent/chat.tsx
petersalas, line 189: One catch with this is that if an assistant message _streams_ in before the function call, it will break the append-only delta calculation below. Need to think through this more.

juberti: Newlines seem kind of artificial to me, you don't see that in typical English text (unless you are suggesting a paragraph break).

petersalas: Since the scenario is:

```
<AssistantMessage>Let me check on that for you.</AssistantMessage>
[...function calls...]
<AssistantMessage>Here's what I found...</AssistantMessage>
```

I think a paragraph break is reasonably appropriate at the end of each `AssistantMessage`?

packages/fixie-sdk/src/request-wrapper.tsx
```diff
@@ -49,6 +65,7 @@ export function FixieRequestWrapper({
               <Json>
                 {{
                   kind: 'text',
+                  state: <MessageState>{message.element}</MessageState>,
```
farzadab, line 68: Q: is there no double rendering because the element is memoized by default? Is that only true of OpenAI or all models/mechanisms?

packages/fixie-sdk/src/request-wrapper.tsx
```diff
@@ -13,6 +13,22 @@ export const RequestContext = AI.createContext<{
...
+  const renderResult = render(children);
+  let didYield = false;
+  for await (const _ of renderResult) {
+    if (!didYield) {
+      didYield = true;
+      yield 'in-progress';
+    }
+  }
+
+  return 'done';
```
farzadab, line 29: I can't 100% verify that this works correctly due my limited knowledge of `for await` details. So the idea is that up until `renderResult` is yielding we stay in the `for` loop and once it actually `returns` the loop is done?
If you've verified this I'd take your word for it.

packages/fixie-sdk/src/fixie-serve-bin.tsx
```diff
@@ -79,10 +79,15 @@ async function serve({
...
               let lastMessages = [];
+              let currentValue = undefined as string | undefined;
               while (true) {
-                let currentValue = undefined as string | undefined;
                 try {
                   const next = await generator.next();
+                  if (!next.done && currentValue == next.value) {
+                    // No change, don't send anything.
+                    continue;
+                  }
```
farzadab, line 89: Q: Is this just for performance optimization or is it needed for correctness too?

packages/fixie-sdk/src/request-wrapper.tsx
petersalas: Excellent question :)

The (not at all obvious) answer is that elements surfaced by partial rendering are automatically memoized (this was not always the case), so any time we see a "conversational" element, it's already been memoized.

packages/fixie-sdk/src/fixie-serve-bin.tsx
petersalas: Good question, it's just an optimization.

packages/fixie-sdk/src/request-wrapper.tsx
petersalas: Another good question. The short answer is that yes, that's how the `render` API works, but your question prompted me to wonder whether this might effectively introduce [screen tearing](https://en.wikipedia.org/wiki/Screen_tearing) because there's no guarantee that this will resolve _after_ the other render of the element resolves.

Need to think about that more :)

packages/voice/src/app/agent/chat.tsx
juberti: I can get behind that.

packages/fixie-sdk/src/request-wrapper.tsx
petersalas: Ok, I added the (slightly heavy-handed, but I think net "good") `batchFrames` option.

The main catch is that it can(/will) delay frames in favor of CPU-bound rendering work, but I suspect on the balance it will perform better in the cases we care about.

juberti (approved):

PR #500, opened 2023-11-17
Upgrade js-tiktoken, use standard interface

petersalas:
js-tiktoken 1.0.8 broke the import we were using to pull in just the cl100k_base encoding. Switch to the standard `getEncoding` import instead.


farzadab (approved):
Woah, that's a quick fix, thanks!

PR #503, opened 2023-11-20
Fix memo bug, allow conversation rounds to run in parallel

petersalas:
This change:
- Fixes a bug where memoized streams sometimes duplicated their content
- Refactors `<Converse>` to allow the rounds to run in parallel when the content allows

### DIFF
--- a/packages/ai-jsx/package.json
+++ b/packages/ai-jsx/package.json
@@ -4,7 +4,7 @@
   "repository": "fixie-ai/ai-jsx",
   "bugs": "https://github.com/fixie-ai/ai-jsx/issues",
   "homepage": "https://ai-jsx.com",
-  "version": "0.27.0",
+  "version": "0.27.1",
   "volta": {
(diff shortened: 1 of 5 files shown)


farzadab (approved):
I'm not sure I understand the difference 100% but I read it multiple times and I don't see an issue :)

petersalas:
Yeah it's sort of mind-bending and I have to constantly remind myself why it matters. The relevant difference between:

```typescript
yield AI.AppendOnlyStream;
yield x;
yield y;
```

and

```typescript
return [x, y]
```

is that in the top one, `x` will (might, depending on what the parent does) be rendered to completion before `y` starts being rendered, whereas in the second one they will render in parallel. What makes it especially mind-bending is that in this case `y` actually _has an internal partial dependency on `x`_. So the trick is to construct a render tree with inter-branch dependencies rather than inter-_frame_ dependencies where possible.

PR #504, opened 2023-11-20
[fixie] Add ability to set default runtime parameters in the CLI

petersalas:
This adds the ability to set default runtime parameters in the `fixie` CLI, which allows experimentation against the default agent runtime using the fixie CLI itself.

Additionally, tighten lint settings and fix lint.


mdwelsh (approved):
Thanks for doing this!

PR #506, opened 2023-11-20
Improve prompt/completion logging

petersalas:
Currently the prompt/completion log events include debug-serialized JSX. But this is a) difficult to extract data from and b) unnecessarily generic because we're already working with a very finite set of conversation components.

This changes the structure to include the final text and props (sans-`children`) instead. Additionally, we include a `getElement()` function so that consumers have the option to access the element itself, without it being JSON-serialized by default.

### DIFF
--- a/packages/ai-jsx/src/core/conversation.tsx
+++ b/packages/ai-jsx/src/core/conversation.tsx
@@ -241,7 +241,7 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
 async function loggableMessage(
   message: ConversationMessage,
   render: AI.RenderContext['render'],
-  measureCost?: (message: ConversationMessage, render: AI.ComponentContext['render']) => Promise<number>
+  cost?: (message: ConversationMessage, render: AI.ComponentContext['render']) => Promise<number>
 ) {
   let textPromise: PromiseLike<string> | undefined = undefined;
   switch (message.type) {
@@ -261,11 +261,13 @@ async function loggableMessage(
     }
   }
 
-  const costPromise = measureCost?.(message, render);
-  const loggableProps = { ...message.element.props };
-  if ('children' in loggableProps) {
-    delete loggableProps.children;
-  }
+  const costPromise = cost?.(message, render);
+
+  const { children, ...propsWithoutChildren } = {
+    children: undefined,
+    ...message.element.props,
+  };
+  const loggableProps: Record<string, Jsonifiable> = propsWithoutChildren;
 
   return {
     // Use a function so that it doesn't serialize to JSON, but can be accessed if needed.
--- a/packages/ai-jsx/src/core/conversation.tsx
+++ b/packages/ai-jsx/src/core/conversation.tsx
@@ -238,6 +238,45 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
   });
 }
 
+async function loggableMessage(
+  message: ConversationMessage,
+  render: AI.RenderContext['render'],
+  measureCost?: (message: ConversationMessage, render: AI.ComponentContext['render']) => Promise<number>
+) {
+  let textPromise: PromiseLike<string> | undefined = undefined;
+  switch (message.type) {
+    case 'user':
+    case 'assistant':
+    case 'system':
+      textPromise = render(message.element);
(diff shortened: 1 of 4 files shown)


packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -238,6 +238,45 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
...
+
+  const costPromise = measureCost?.(message, render);
+  const loggableProps = { ...message.element.props };
+  if ('children' in loggableProps) {
+    delete loggableProps.children;
+  }
+
+  return {
+    // Use a function so that it doesn't serialize to JSON, but can be accessed if needed.
+    getElement: () => message.element,
```
farzadab, line 272: This won't cause issues for logging? Does it get ignored?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -238,6 +238,45 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
...
+  message: ConversationMessage,
+  render: AI.RenderContext['render'],
+  measureCost?: (message: ConversationMessage, render: AI.ComponentContext['render']) => Promise<number>
+) {
+  let textPromise: PromiseLike<string> | undefined = undefined;
+  switch (message.type) {
+    case 'user':
+    case 'assistant':
+    case 'system':
+      textPromise = render(message.element);
```
jordangarcia, line 251: what's the cost of rendering in this function?  will it start a chain of other renders if part of the tree has not rendered? Or is the nature of when this is called gaurantee that the element already has been rendered and is now memoized?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -238,6 +238,45 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
...
+      const neverMessage: never = message;
+      throw new Error(`Unexpected message type ${(neverMessage as any).type}`);
+    }
+  }
+
+  const costPromise = measureCost?.(message, render);
+  const loggableProps = { ...message.element.props };
+  if ('children' in loggableProps) {
+    delete loggableProps.children;
+  }
```
farzadab, line 268: Are all the other `props` Jsonifiable and `children` is the only one that gets rendered?

packages/ai-jsx/src/core/conversation.tsx
```diff
@@ -238,6 +238,45 @@ function toConversationMessages(partialRendering: AI.PartiallyRendered[]): Conve
   });
 }
 
+async function loggableMessage(
+  message: ConversationMessage,
+  render: AI.RenderContext['render'],
+  measureCost?: (message: ConversationMessage, render: AI.ComponentContext['render']) => Promise<number>
```
farzadab, line 244: Nit: I'd rather have this name and the one in `renderToConversation` match. I don't have a preference as to which one to keep.

farzadab (approved):
LGTM with some Qs/comments.

packages/ai-jsx/src/core/conversation.tsx
petersalas: Good question! It's certainly possible for this to block on rendering that hasn't yet happened. But in practice, we already do that when logging conversations to measure token cost, and this will only happen if the caller requests the conversation to be logged (i.e. the pre-model/post-model invocations).

So in practice, it won't cause any rendering to happen meaningfully earlier than it already would.

packages/ai-jsx/src/core/conversation.tsx
petersalas: Yup:

```
> JSON.stringify({ x: () => 1, y: 2 })
'{"y":2}'
```

packages/ai-jsx/src/core/conversation.tsx
petersalas: Yup exactly. Will improve this to make it explicitly type-checked :)

packages/ai-jsx/src/core/conversation.tsx
farzadab: Woah that's so nice. Python just throws an exception!

PR #508, opened 2023-11-21
Allow override of websocket URL for staging

juberti:


packages/voice/src/app/agent/page.tsx
```diff
@@ -221,7 +221,7 @@ const AgentPageComponent: React.FC = () => {
   const ttsVoice = searchParams.get('ttsVoice') || agentVoice;
   const model = getAgent(agentId) === undefined ? 'fixie' : searchParams.get('llm') || DEFAULT_LLM;
   const docs = searchParams.get('docs') !== null;
-  const webrtc = searchParams.get('webrtc') !== null;
+  const webrtcUrl = searchParams.get('webrtc') ?? undefined;
```
mdepinet, line 224: Could you clarify this for me? My expectation is that with this change we'll have:

"/" => client-hosted
"/?webrtc" => server-hosted with default urls
"/?webrtc=wss://some-gcr-url" => server-hosted with an arbitrary deployment

Is that right?

mdepinet (approved):

packages/voice/src/app/agent/page.tsx
juberti: Yeah, I just wanted to be able to test a local FE with a staging deployment.

PR #513, opened 2023-11-27
Fix WebRTC visualizer bug

farzadab:
Bugfix:
- state changes to speaking before `outAnalyzer` gets set (React doesn't know about underlying object change)
- setting `audioContext` on the `mediaStream` of LiveKit remote track

### DIFF
--- a/packages/voice/src/app/agent/chat.tsx
+++ b/packages/voice/src/app/agent/chat.tsx
@@ -544,6 +544,8 @@ export class WebRtcChatManager implements ChatManager {
   private socket?: WebSocket;
   private room?: Room;
   private localAudioTrack?: LocalAudioTrack;
+  /** True when we should have entered speaking state but didn't due to analyzer not being ready. */
+  private delayedSpeakingState = false;
   private inAnalyzer?: StreamAnalyzer;
   private outAnalyzer?: StreamAnalyzer;
   private pinger?: NodeJS.Timer;
@@ -571,7 +573,7 @@ export class WebRtcChatManager implements ChatManager {
   }
   warmup() {
     const isLocalHost = window.location.hostname === 'localhost';
-    const url = this.params.webrtcUrl || (!isLocalHost ? 'wss://wsapi.fixie.ai' : 'ws://localhost:8100');
+    const url = this.params.webrtcUrl || (isLocalHost ? 'wss://wsapi.fixie.ai' : 'ws://localhost:8100');
     this.socket = new WebSocket(url);
     this.socket.onopen = () => this.handleSocketOpen();
     this.socket.onmessage = (event) => this.handleSocketMessage(event);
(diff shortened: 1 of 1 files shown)


packages/voice/src/app/agent/chat.tsx
```diff
@@ -675,8 +675,10 @@ export class WebRtcChatManager implements ChatManager {
     const audioTrack = track as RemoteAudioTrack;
     audioTrack.on(TrackEvent.AudioPlaybackStarted, () => console.log(`[chat] audio playback started`));
     audioTrack.on(TrackEvent.AudioPlaybackFailed, (err) => console.error(`[chat] audio playback failed`, err));
+    audioTrack.setAudioContext(this.audioContext);
```
juberti, line 678: OOC, why is this needed?

packages/voice/src/app/agent/chat.tsx
```diff
@@ -685,6 +687,11 @@ export class WebRtcChatManager implements ChatManager {
       console.debug(`[chat] worker RTT: ${elapsed_ms.toFixed(0)} ms`);
     } else if (data.type === 'state') {
       const newState = data.state;
+      if (newState === 'speaking' && this.outAnalyzer === undefined) {
```
juberti, line 690: should this be `ChatManagerState.SPEAKING`?

packages/voice/src/app/agent/chat.tsx
```diff
@@ -675,8 +675,10 @@ export class WebRtcChatManager implements ChatManager {
     const audioTrack = track as RemoteAudioTrack;
     audioTrack.on(TrackEvent.AudioPlaybackStarted, () => console.log(`[chat] audio playback started`));
     audioTrack.on(TrackEvent.AudioPlaybackFailed, (err) => console.error(`[chat] audio playback failed`, err));
+    audioTrack.setAudioContext(this.audioContext);
     audioTrack.attach(this.audioElement);
     this.outAnalyzer = new StreamAnalyzer(this.audioContext, track.mediaStream!);
+    this.changeState(ChatManagerState.SPEAKING);
```
juberti, line 681: Probably only want to do this in the case you mention below, right?

juberti (commented):
Nice finds, this mostly makes sense to me. Noted a couple nits.

packages/voice/src/app/agent/chat.tsx
farzadab: The real answer is I was just playing with it and saw the function exists, thought maybe I should connect the two objects because they seemed to be coming from different places ... so basically just trial and error 😅

packages/voice/src/app/agent/chat.tsx
farzadab: Updated.

packages/voice/src/app/agent/chat.tsx
juberti: OK... would be good to try to understand what is happening internally and add a comment indicating that as it's not clear at all to me how setting this context affects our own visualizer objects

packages/voice/src/app/agent/chat.tsx
```diff
@@ -571,7 +573,7 @@ export class WebRtcChatManager implements ChatManager {
   }
   warmup() {
     const isLocalHost = window.location.hostname === 'localhost';
-    const url = this.params.webrtcUrl || (!isLocalHost ? 'wss://wsapi.fixie.ai' : 'ws://localhost:8100');
+    const url = this.params.webrtcUrl || (isLocalHost ? 'wss://wsapi.fixie.ai' : 'ws://localhost:8100');
```
juberti, line 576: this doesn't seem right (localhost is set up to point at a local voice server)

juberti (approved):
LG but see comment about `setAudioContext`

packages/voice/src/app/agent/chat.tsx
farzadab: Sorry I didn't get to figure it out. I left a TODO for when I come back to it.

PR #536, opened 2024-01-22
Change function parameters to be JSON Schema

petersalas:
Previously the function definitions were typed as either Zod or JSON Schema-ish (with some differences around `required`). Rather than have a layer of conversion logic (`getParametersSchema`) this makes the input types explicitly JSON Schema. This facilitates integration with other types, e.g. OpenAPI specs. (Callers using Zod must now do the conversion themselves, e.g. with `zod-to-json-schema`.)


farzadab (approved):
LGTM

PR #541, opened 2024-04-09
Update AnthropicChatModel, relax OpenAIChatModel

petersalas:
- Update Anthropic SDK, switch to messages API, add support for beta tool use API
- Relax OpenAI function calling restrictions (all models support function calling now)
- Add `maxInputTokens` prop to `OpenAIChatModel` to support explicitly limiting prompt size
- Add new GPT-4 models to enum

### DIFF
--- a/packages/ai-jsx/src/lib/openai.tsx
+++ b/packages/ai-jsx/src/lib/openai.tsx
@@ -3,7 +3,7 @@
  * @packageDocumentation
  */
 
-import { Jsonifiable, MergeExclusive } from 'type-fest';
+import { Jsonifiable } from 'type-fest';
 import {
   ChatProvider,
   CompletionProvider,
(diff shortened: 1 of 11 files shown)


packages/ai-jsx/src/lib/openai.tsx
```diff
@@ -156,15 +158,6 @@ function logitBiasOfTokens(tokens: Record<string, number>) {
   );
 }
 
-/**
- * Returns true if the given model supports function calling.
- * @param model The model to check.
- * @returns True if the model supports function calling, false otherwise.
- */
-function chatModelSupportsFunctions(model: ValidChatModel) {
```
mdepinet, line 164: Does it make sense for ChatModel to define whether functions are supported (defaulting to true at this point)?  I'm thinking about our llama-2 proxy use case in particular.

mdepinet (approved):

packages/ai-jsx/src/lib/openai.tsx
petersalas: I think once we point OpenAIChatModel at a non-OAI endpoint, we should take care of that on our side. But since all of OpenAI's current chat models support function calling -- IMO not at this layer.

mdepinet: Does it make sense to formalize the idea of a non-OAI model that uses the same API surface then?

(I happy with whatever you decide. Just want to make sure we have a plan for this!)

PR #546, opened 2024-05-10
Add support for images in UserMessages

petersalas:
This adds support for passing images as inputs to OpenAI models. They can be embedded in `<UserMessage>`s using the existing `<Image>` tag.

By default `OpenAIChatModel` passes images for models that supports it, but this can be forced (or denied) by setting the `includeImages` props explicitly.


mdepinet (commented):
LGTM!

